PR #15, opened 2022-10-24
Feat/tokens/create automated flows

InteractiveTimmy:
# Description

The scope of the changes in this Pull request are to modify the existing build flows to support tiered token generation [core => light/dark/etc]. This can be easily configured via the `./packages/@momentum-design/tokens/src/brands/{brand}/` scripts to set up new themes/styles for a specified brand utilizing the scoped `models` within this sub-package, then by injecting the new brand into the `./packages/@momentum-design/tokens/src/commands/build/build.ts::build()` method.

These changes also create a clear definition on how tokens should be utilized in order to create a scalable solution across multiple design systems while offering the flexibility for designers and engineers to designate which tokens/files/platforms should be generated.

Screen Shot 2022-10-24 at 12 55 11 PM
Screen Shot 2022-10-24 at 12 56 34 PM


packages/@momentum-design/tokens/reference/brand/webex/theme-dark.json
```diff
@@ -0,0 +1,13 @@
+{
+  "color": {
+    "theme": {
+      "button": {
```
chrisadubois, line 4: right so the wiring is thjere for all the other types that designers will provide

packages/@momentum-design/tokens/src/brands/webex/webex.constants.ts
```diff
@@ -0,0 +1,61 @@
+const CONSTANTS = {
+  FILES: [
```
chrisadubois, line 2: the constants inside of this constants file should probably be constantized as well, as there will be others (not webex) that use the same values and if we change them, would want to propagate

packages/@momentum-design/tokens/src/commands/build/build.constants.ts
```diff
@@ -0,0 +1,7 @@
+const CONSTANTS = {
+  SCOPES: [
+    'webex',
```
chrisadubois, line 3: this one is fine though.

I would also say to change name of file to just constants.ts

packages/@momentum-design/tokens/src/brands/webex/webex.constants.ts
```diff
@@ -0,0 +1,61 @@
...
+    {
+      CATEGORIES: ['color'],
+      FILTERS: ['**/core.json'],
+      ITEMS: [],
+      NAME: 'core',
+      TYPES: ['core', 'decorative', 'gradient', 'mobile'],
```
chrisadubois, line 8: prefer ENUMs to array of strings

packages/@momentum-design/tokens/src/brands/webex/webex.constants.ts
chrisadubois: ideally these values would be plucked from an enum

packages/@momentum-design/tokens/src/index.ts
```diff
@@ -1,42 +1,20 @@
...
+  if (args.build) {
+    commands.build(...args.build as Array<string>);
+
+    return;
+  }
+
+  throw new Error('provided arguments are not valid');
+};
+
+const args = yargs(process.argv.slice(2)).options({
```
chrisadubois, line 16: hmmmm

packages/@momentum-design/tokens/src/models/file/file.fixture.ts
```diff
@@ -0,0 +1,23 @@
+import File from './file';
+import type { Config } from './file.types';
+
+const FILE_FIXTURE_CONFIG = {
+  categories: ['category'],
+  extension: 'ext',
```
chrisadubois, line 6: samesies on enum and constants here

packages/@momentum-design/tokens/src/models/file/file.types.ts
```diff
@@ -0,0 +1,27 @@
+import { TransformedToken } from 'style-dictionary';
```
chrisadubois, line 1: same on name of file here

packages/@momentum-design/tokens/src/models/scope/scope.ts
```diff
@@ -0,0 +1,99 @@
...
+    return this.filters.length > 0
+      ? this.filters.map((filter) => path.join(this.referencePath, filter))
+      : [path.join(this.referencePath, Scope.CONSTANTS.REFERENCE_WILDCARD)];
+  }
+
+  public get referencePath(): string {
+    return path.join(COMMON_CONSTANTS.PATHS.REFERENCE, this.type, this.name);
+  }
+
+  public get serial(): Serial {
```
chrisadubois, line 67: prefer serialized() to serial

chrisadubois (dismissed):
See comments

PR #34, opened 2022-10-31
feat(pipeline): fixes #18

### LINKED ISSUE #18: Github Actions PR Config
# Description

This project will require testing for pull requests on the various packages that reside within it. Implement a GitHub actions configuration that allows for pull request validation.

# Requirements

* [ ] Automation configuration should trigger on each new pull request after it is validated
* [ ] Automation configuration should auto-validate any pull requests generated by the [momentum-design-engineers team](https://github.com/orgs/momentum-design/teams/momentum-design-engineers)
* [ ] Automation configuration should **not** execute any package scripts, as they are not within the scope of this task.
* [ ] Automation configuration should emit a message indicating that it is operational. *Note: this is to be removed after the automation CI needs are established*

chrisadubois:
**Description of changes made (Required)**
Adding PR github config and stub for CD config

**Associated issue number/tracking link (Required)**
#18


chrisadubois:
here's a sample of the job run: https://github.com/chrisadubois/momentum-design/actions/runs/3358603474/jobs/5565793673

I tried using nektos/act locally but i couldnt get it to work. I think it's due to Apple M1

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
+name: Continuous Integration
+run-name: ${{ github.actor }} is running continuous integration
+
+on:
+  pull_request:
+  push:
+  # The "workflow_dispatch" event gives us a button in GitHub's "Action" UI
+  workflow_dispatch:
```
InteractiveTimmy, line 8: I don't believe this will work, you'll need to use [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows). This will need to be manually validated via a label or something similar. See [this example](https://github.com/momentum-design/momentum-react-v2/blob/master/.github/workflows/pull-request.yml).

I may be wrong, but, this should only work on branches to this repo using `pull_request` instead of `pull_request_target`, which works on forks.

InteractiveTimmy (changes requested):
One comment around CI/CD triggers, I'm pretty sure this will fail until we introduce `pull_request_target` triggers.

InteractiveTimmy (approved):
LGTM!

Issue #24, opened 2022-10-25
Tokens Documentation


InteractiveTimmy:
# Description

Add documentation to the `@momentum-design/tokens` package, inclusive of core documentation along with TSDoc documentation.

# Requirements

* [ ] Should generate JSON output within the `./docs/metadata` folder of the `@momentum-design/tokens` package
* [ ] Should include the following core markdown files
  * [ ] `README.md`
  * [ ] `CONTRIBUTING.md`
  * [ ] `SCRIPTS.md`

PR #52, opened 2022-11-01
docs(token-builder): generate package markdown files

InteractiveTimmy:
# Description

The scope of the changes in this Pull Request are to generate the core project documentation within the `@momentum-design/token-builder` package. This is inclusive of:

* `./README.md`
* `./CONTRIBUTING.md`
* `./SCRIPTS.md`
* `./LICENSE.md`

While cleaning up the `./PROPOSAL.md` file.

## Links

https://github.com/momentum-design/momentum-design/issues/24


packages/@momentum-design/token-builder/package.json
```diff
@@ -2,6 +2,10 @@
   "name": "@momentum-design/token-builder",
   "packageManager": "yarn@3.2.4",
   "version": "0.0.0",
+  "engines": {
+    "npm": ">=8.0.0",
```
chrisadubois, line 6: should we have a yarn version specification here?

Should we have these engines upleveled to package root as well?

chrisadubois (dismissed):
see question

packages/@momentum-design/token-builder/package.json
InteractiveTimmy: >should we have a yarn version specification here?

Yes, good idea. This is necessary since all of the scripts use `yarn`

>Should we have these engines upleveled to package root as well?

Yes, I'll add this to our root repo as well in this PR.

chrisadubois (approved):

PR #61, opened 2022-11-03
feat(momentum sync): get momentum data

ArthusLiang:
# Description

***required***

*Description of the changes made in this Pull Request*

## Links

***required***

*Links to resources related to this Pull Request*


packages/@momentum-design/momentum-abstract-builder/src/common/constants.ts
```diff
@@ -0,0 +1,18 @@
+const path = require('path');
+
+const packagePath = path.resolve(__dirname, '../../../');
```
cscheffauer, line 3: where is this path pointing to? is this pointing to `packages/@momentum-design`?

packages/@momentum-design/momentum-abstract-builder/src/models/momentum-abstract-sync.ts
```diff
@@ -0,0 +1,54 @@
...
+    async sync() {
+        const myPackage = require(CONSTANTS.packageJsonPath);
+        const online = JSON.parse(await this.getOnlineVersion('momentum-abstract'));
+        if(online && online.latest && myPackage.dependencies && myPackage.dependencies['momentum-abstract']) {
+            if(this.ifNeedUpdate(myPackage.dependencies['momentum-abstract'].replace(regVersion,''), online.latest.replace(regVersion,''))) { // >=
+                execSync('yarn workspace @momentum-design/momentum-abstract-builder up momentum-abstract');
+                momentumAbtractBuilder.buildMomentumDesign();
+            }
+        } else {
+            //console.log(`${ packageJson.name }'s laster verison is not found.`);
```
cscheffauer, line 32: pls remove commented code

packages/@momentum-design/momentum-abstract-builder/src/models/momentum-abstract-sync.ts
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    ifNeedUpdate(packageVerison:string, onlineNewVersion:string):boolean {
+        const packageVerisonArr = packageVerison.split('.');
```
cscheffauer, line 37: typo `Verison`

packages/@momentum-design/momentum-abstract-builder/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  // "extends": "../../../config/typescript/tsconfig.base.json",
```
cscheffauer, line 2: is it intentional to not extend the base config here? I think thats a common pattern we followed in the other packages. If so, pls remove the commented line (no need to keep commented lines in prod code)

packages/@momentum-design/momentum-abstract-builder/src/sync.ts
```diff
@@ -0,0 +1,7 @@
+import { momentumAbstractSync } from './models/index';
+
+const sync =  async ()=> {
+    await momentumAbstractSync.sync();
+}
+
+sync();
```
cscheffauer, line 7: sync will run `synchronously`, but it is a async function. Should it be `await sync()` here instead?

packages/@momentum-design/momentum-abstract-builder/src/models/momentum-abstract-sync.ts
```diff
@@ -0,0 +1,54 @@
...
+        const onlineNewVersionArr = onlineNewVersion.split('.');
+        if(packageVerisonArr.length!==onlineNewVersionArr.length) {
```
cscheffauer, line 39: this looks like prettier or eslint haven't run over it - can u pls run them / format these lines correctly?
for example I think we put spaces before & after a `!==` etc

packages/@momentum-design/momentum-abstract-builder/src/models/momentum-abstract-sync.ts
```diff
@@ -0,0 +1,54 @@
+import { momentumAbtractBuilder } from './momentum-abstract-builder';
+import { CONSTANTS } from '../common/index';
+const { execSync } = require('child_process');
+
+const https = require('https');
+const regVersion = /^[\~|\^]/;
+
+class MomentumAbstractSync {
```
cscheffauer, line 8: can u pls add tests for these class methods? writing tests for those will make sure the methods execute as expected and do not break over time

packages/@momentum-design/momentum-abstract-builder/src/models/momentum-abstract-builder.ts
```diff
@@ -0,0 +1,23 @@
...
+        mfs.clean(config.output);
+        mfs.save(config.output, config.type, config.filter);
+    }
+
+    buildMomentumDesign() {
+        this.build({ output: CONSTANTS.source.icon, type: MomentumAbstractType.icon });
+        this.build({ output: CONSTANTS.source.iconColored, type: MomentumAbstractType['icon-colored'] });
+        this.build({ output: CONSTANTS.source.iconBrand, type: MomentumAbstractType['icon-brand'] });
+        this.build({ output: CONSTANTS.source.illustration, type: MomentumAbstractType.illustration });
+        //this.build({ output: CONSTANTS.source.sonic, type: MomentumAbstractType.sonic });
```
cscheffauer, line 16: are those commented out, cause we don't have the folders yet?
i would recommend to remove commented code out of prod code (if we forget about it, it will become stale code, which just pollutes the repo over time)

packages/@momentum-design/momentum-abstract-builder/src/common/constants.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const CONSTANTS = {
+    source: {
+        icon: path.join(packagePath, 'icons/src/core'),
```
cscheffauer, line 7: all these constants rely on the fact that the paths are not changing / are named correctly. Shall we add unit tests for those folders? i could see value in tests which do verify if the provides paths exist.
That would make sure in the future that whenever devs are going to change the folder structure, the pipeline will tell them that they shouldn't change the folder structure ;)

packages/@momentum-design/momentum-abstract-builder/README.md
```diff
@@ -0,0 +1,24 @@
+---
+package: @momentum-design/momentum-abstract-builder
+---
+
+# MomentumAbstractBuilder
+
+## Overview
+
+This component package was generated with a script, please report any issues.
```
cscheffauer, line 9: might be worth adding more information here. Like what is the purpose of MomentumAbstractBuilder.
like if it runs, x y z is happening :)

cscheffauer (changes requested):
some comments :) great work so far 💯

packages/@momentum-design/momentum-abstract-builder/src/common/constants.ts
ArthusLiang: Yes

packages/@momentum-design/momentum-abstract-builder/tsconfig.json
ArthusLiang: Got it

packages/@momentum-design/momentum-abstract-builder/src/sync.ts
ArthusLiang: await is not allowed in according the es version. 
Sync() works here when await is inside the function.

packages/@momentum-design/momentum-abstract-builder/src/models/momentum-abstract-sync.ts
ArthusLiang: Hm, let me check it why eslint passed here in my local environment,

packages/@momentum-design/momentum-abstract-builder/src/models/momentum-abstract-sync.ts
ArthusLiang: Sure

packages/@momentum-design/momentum-abstract-builder/src/common/constants.ts
ArthusLiang: Sure

packages/@momentum-design/momentum-abstract-builder/src/common/constants.ts
cscheffauer: agree - pls try to import the packages accordingly, without having relative paths :)

.github/workflows/sync.yml
```diff
@@ -0,0 +1,24 @@
+on:
+  schedule:
+    # runs on 1:01  17:01 UTC
+    - cron:  '01 2,8,14,20 * * *'
```
InteractiveTimmy, line 4: Can we do a `workflow_dispatch` here? We'll re-enable a cron job as soon as we're ready to officially.

.github/workflows/sync.yml
```diff
@@ -0,0 +1,24 @@
...
+name: Publish To NPM
+jobs:
+  jobAutomation:
+    name: Sync momentum-abstract data
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'momentum-design' }}
+    steps:
+    - name: Check out
+      uses: actions/checkout@master
```
InteractiveTimmy, line 13: This should be `actions/checkout@v3`

InteractiveTimmy (changes requested):
Some small changes needed in the CI/CD flow, but otherwise, looks good!

packages/@momentum-design/momentum-abstract-builder/src/common/constants.ts
ArthusLiang: Hm, I do not import packages.  The code here will copy data to the packagePath.

.github/workflows/sync.yml
```diff
@@ -0,0 +1,23 @@
...
+      run: yarn workspace @momentum-design/momentum-abstract-builder init-build
+    - name: Publish
+      run: yarn workspace @momentum-design/momentum-abstract-builder sync
+    - name: config 
+      run: |
+        git config --global user.email "automation@cisco.com"
+        git config --global user.name "Automation"
+    - name: commit new release
+      run: yarn workspace @momentum-design/momentum-abstract-builder commit token=${{ secrets.ACTION_TOKEN }} repository=${{github.repository}}
+
```
InteractiveTimmy, line 23: Needs a new line at EOF.

.gitignore
```diff
@@ -111,3 +111,6 @@ dist
 
 # TernJS port file
 .tern-port
+
+# local files
+**/.DS_Store
```
InteractiveTimmy, line 116: good add! helps remove bloat in system files from certain OSes

.github/workflows/sync.yml
```diff
@@ -0,0 +1,23 @@
+on:
```
InteractiveTimmy, line 1: If you could apply a name to the top of this file much like that which is seen in the other workflows. This will help us identify it if it runs.

InteractiveTimmy (changes requested):

.github/workflows/sync.yml
ArthusLiang: Sure

packages/@momentum-design/momentum-abstract-builder/src/common/constants.ts
cscheffauer: I understand that. The problem with the absolute/relative paths (`'../../../'`) u have there is that whenever the folder structure of the mono repository changes, this script here will fail. 
i'm not 100% into the monorepo tech, but I feel like Typescript aliases would be a good solution to solve this problem.
https://rossbulat.medium.com/typescript-working-with-paths-packages-and-yarn-workspaces-6fbc7087b325

Found this one about it - as mentioned, i'm not 100% into how to set it up properly - @InteractiveTimmy should be able to help out on this specific topic.

packages/@momentum-design/momentum-abstract-builder/src/models/momentum-abstract-sync.test.ts
```diff
@@ -0,0 +1,30 @@
+import { momentumAbstractSync } from './momentum-abstract-sync';
+import { CONSTANTS } from '../common/index';
+
+describe('momentum-abstract-sync', () => {
+  test('test get momentum-absract version', async () => {
+    const online = JSON.parse(await momentumAbstractSync.getOnlineVersion('momentum-abstract'));
```
cscheffauer, line 6: this sort of test should be avoided if possible. In this test u are going to always fire a https.get call to the npm registry service - which means that whenever that URL is not reachable / or there are network issues or similar on the CI setup, this will fail. Also its costly in terms of runtime to fire https calls as part of unit tests.

In this scenario I would recommend to mock the https.get if possible:
https://stackoverflow.com/questions/68745947/how-to-mock-https-get-with-jest
and u test that the mock is fired. That will not only make your test quicker, but more stable against any sort of network issues.

packages/@momentum-design/momentum-abstract-builder/src/models/momentum-abstract-sync.ts
```diff
@@ -0,0 +1,58 @@
...
+      const pVersion = myPackage.dependencies['momentum-abstract'].replace(regVersion, '');
+      const oVersion = online.latest.replace(regVersion, '');
+      if (this.ifNeedUpdate(pVersion, oVersion)) { // >=
+        execSync('yarn workspace @momentum-design/momentum-abstract-builder up momentum-abstract');
+        momentumAbtractBuilder.buildMomentumDesign();
+      }
+    }
+  }
+
+  ifNeedUpdate(packageVerison:string, onlineNewVersion:string):boolean {
```
cscheffauer, line 41: typo `packageVerison`

packages/@momentum-design/momentum-abstract-builder/README.md
cscheffauer: @ArthusLiang have u already changed this README ? i am missing the requested information - like what is the purpose of MomentumAbstractBuilder, how does it work, what happens if script x y z is called

cscheffauer (changes requested):
pipeline is also failing - pls double check why :)

packages/@momentum-design/momentum-abstract-builder/src/models/momentum-abstract-sync.test.ts
ArthusLiang: Hm, I think I can remove this test.

ArthusLiang: If the folder structure changes, it always breaks code. @InteractiveTimmy Maybe we need a tsconfig at the root?

packages/@momentum-design/momentum-abstract-builder/README.md
ArthusLiang: I thought I changed. But it is not here. Let me check it.

chrisadubois:
@ArthusLiang can you just point the icons illustrations figma tokens plugin to post to this repository

PR #71, opened 2022-11-03
fix(cd): publish flow

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to fix the publish flow, as it was missing a checkout and install.


chrisadubois (approved):
i think we should try to use CI script in this one but this is good for now

chrisadubois (approved):
should ideally reuse ci-script

PR #87, opened 2022-11-09
feat: add semantic release

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to add `semantic-release` and `semantic-release-monorepo` to the list of dependencies available, and amend the associated flows to our existing `tokens` package.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,8 +25,8 @@ jobs:
       - name: Build Packages
         run: yarn build
 
-      - name: Set NPM Authorization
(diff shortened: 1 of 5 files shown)


chrisadubois (changes requested):
Please add these changes that are root level to the plop generator,

Why isn't it part of token builder as well for semantic release?

InteractiveTimmy:
>Please add these changes that are root level to the plop generator,
Why isn't it part of token builder as well for semantic release?

These changes are not root level. We do not want new packages being created that immediately begin their publish flow, as this will pollute npm at this stage in development.

We want to implicitly add publish flows to each package as they are needed, until we reach our first release.

PR #139, opened 2022-11-18
docs: added harbor investigation

cscheffauer:
# Description

- Added the harbor investigation docs & decision


cscheffauer:
moved to wiki, as suggested by @torgeadelin

PR #150, opened 2022-11-22
Feat/docs/add initial project

torgeadelin:
# Description

Added docs project structure + tokens docs.
Most of the code is boilerplate.

Most code I added  is under `pages/tokens`
## Links

Demo: http://10.209.86.40:3000/


torgeadelin:
Things left to do:
- update the plop generated files to reflect the correct scripts 
- maybe figure out a way to automatically generate the token pages
- add build in the pipeline process and publish to github pages
- grab all the API references generated by api-extractor and add them to the docs.

chrisadubois (approved):
saw your comments on the PR, that would be good to do

requirement is for sure not to store the built code on /docs but to build it via github actions and deploy to the docs branch every time a merge occurs

requirement is also to add automation for the token pages (should not be manual as the designers are automated)

else than that -- I think the cat gif is funny, but we should probably not have that screen there

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,79 @@
...
+  # BUILD_PATH: subfolder
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Detect package manager
```
InteractiveTimmy, line 37: Use `actions/node` instead, as it does a check similar to this:

```
    steps:
      - name: Checkout Project
        uses: actions/checkout@v3

      - name: Setup Node
        uses: actions/setup-node@v3
        with:
          node-version-file: '.nvmrc'
          registry-url: 'https://registry.npmjs.org'
          cache: 'yarn'
```

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,79 @@
...
+  BUILD_PATH: "." # default value when not using subfolders
+  # BUILD_PATH: subfolder
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
InteractiveTimmy, line 36: new line after step completion:

```
steps:
  - name: Checkout
  - uses: actions/checkout@v3
  
  - name: ...
```

style suggestion: needed for all steps in this file

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,79 @@
...
+            echo "Unable to determine packager manager"
+            exit 1
+          fi
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: "18"
+          cache: ${{ steps.detect-package-manager.outputs.manager }}
+          cache-dependency-path: ${{ env.BUILD_PATH }}/package-lock.json
+      - name: Install dependencies
```
InteractiveTimmy, line 58: Suggestion: This could be an initialization job:

```
      - name: Validate Dependencies
        id: validate-dependencies
        uses: actions/cache@v3
        with:
          path: '**/node_modules'
          key: node-modules-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}

      - name: Install Dependencies
        if: steps.validate-dependencies.outputs.cache-hit != 'true'
        run: yarn

      - name: Cache Dependencies
        uses: actions/cache@v3
        if: steps.validate-dependencies.outputs.cache-hit != 'true'
        with:
          path: '**/node_modules'
          key: node-modules-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
```

See: .github/workflows/deploy-package.yml

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,79 @@
...
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow one concurrent deployment
+concurrency:
```
InteractiveTimmy, line 22: not sure if concurrency is needed, is there a reason for this?

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,79 @@
+# Sample workflow for building and deploying an Astro site to GitHub Pages
+#
+# To get started with Astro see: https://docs.astro.build/en/getting-started/
+#
+name: Deploy Astro site to Pages
```
InteractiveTimmy, line 5: This *could* exist in the `deploy-package.yml` file, unless the intent is to have this run via manual trigger, in this case, add `workflow_dispatch:` to your `on` clause

```
on:
  push:
    branches:
      - main
  workflow_dispatch:
```

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["main"]
```
InteractiveTimmy, line 10: recommendation:

```
on:
  push:
    branches:
      - main
  workflow_dispatch:
```

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,79 @@
...
+  pages: write
+  id-token: write
+
+# Allow one concurrent deployment
+concurrency:
+  group: "pages"
+  cancel-in-progress: true
+
+env:
+  BUILD_PATH: "." # default value when not using subfolders
```
InteractiveTimmy, line 27: Question: Assuming this allows for astro to build to the project root dir?

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,79 @@
...
+          elif [ -f "${{ github.workspace }}/package.json" ]; then
+            echo "manager=npm" >> $GITHUB_OUTPUT
+            echo "command=ci" >> $GITHUB_OUTPUT
+            exit 0
+          else
+            echo "Unable to determine packager manager"
+            exit 1
+          fi
+      - name: Setup Node
+        uses: actions/setup-node@v3
```
InteractiveTimmy, line 53: This one makes the above check redundant, doesn't it?

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,79 @@
...
+          cache: ${{ steps.detect-package-manager.outputs.manager }}
+          cache-dependency-path: ${{ env.BUILD_PATH }}/package-lock.json
+      - name: Install dependencies
+        run: ${{ steps.detect-package-manager.outputs.manager }} ${{ steps.detect-package-manager.outputs.command }}
+        working-directory: ${{ env.BUILD_PATH }}
+      - name: Build with Astro
+        run: ${{ steps.detect-package-manager.outputs.manager }} run build
+        working-directory: ${{ env.BUILD_PATH }}
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v1
```
InteractiveTimmy, line 65: Is this needed with the below step? This is likely correct, but I'm curious as to if `@actions/deploy-pages` does this already.

InteractiveTimmy (changes requested):
review of just `docs.yaml`. Rename this file extension to `yml` if you can.

.github/workflows/docs.yaml
chrisadubois: i actually dont think we need this file at all?

.github/workflows/docs.yaml
torgeadelin: refactored

.github/workflows/docs.yaml
torgeadelin: refactored

.github/workflows/docs.yaml
torgeadelin: hmm, I'm not sure what the best thing to do is. maybe we leave it as separate flow to make everything more decoupled?

.github/workflows/docs.yaml
torgeadelin: I think that's what this is for, yes. 
I used a template to create this workflow.

InteractiveTimmy (approved):
LGTM, @chrisadubois , could you take a quick look at this PR?

PR #161, opened 2022-11-23
feat(builder): added icons builder and icons package

cscheffauer:
# Description

- Extended the builder package with icons
  - Modified initialise, more to come on the process method. 
- Added config & build script to `icons` package, which will transform & optimise the `src` folder and outputs it to `dist` folder (this will be the dist we are trying to have the same as momentum-ui does for the icons)

### DIFF
--- a/packages/@momentum-design/icons/config/webex.json
+++ b/packages/@momentum-design/icons/config/webex.json
@@ -1,0 +1,7 @@
+{
+    "definitions": [{
+        "srcDir": "./src",
+        "distDir": "./dist",
+        "type": "icons"
+    }]
+}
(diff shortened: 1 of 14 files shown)


packages/@momentum-design/icons/config/webex.json
```diff
@@ -0,0 +1,7 @@
+{
+    "definitions": [{
+        "srcDir": "./src",
+        "distDir": "./dist",
+        "type": "icons"
+    }]
+}
```
torgeadelin, line 7: new line

torgeadelin (approved):
LGTM!

packages/@momentum-design/icons/config/webex.json
cscheffauer: will do in next PR

PR #166, opened 2022-11-25
Format-config: add format options to token builder

nataliadelmar:
# Description

Currently, our configuration object does not include the ability to add additional options to formats. These formats are currently accepted as an array of strings, specifying a specific format name. The file object within StyleDictionary, configured in ./src/models/file, allow for `options` key:value object to be passed in to provide details, such as `selector` in the case of the `CSS_VARIABLES` format. Investigate this and determine a way to apply formatting to specific files based on the `options` value needed in StyleDictionary.

## Links

[***required*** - *Links to resources related to this Pull Request*](https://app.asana.com/0/1203271134507278/1203323596328764/f)


packages/@momentum-design/token-builder/src/models/file/file.ts
```diff
@@ -71,13 +71,45 @@ class File {
...
     };
   }
 
   public get references(): boolean {
-    return this.config.references || false;
+    return this.config.outputReferences || false;
+  }
+
+  public get selector(): string | undefined {
+    return this.config.cssSelector || undefined;
```
InteractiveTimmy, line 89: Not sure if we need `|| undefined` here.

packages/@momentum-design/token-builder/src/models/file/file.ts
```diff
@@ -71,13 +71,45 @@ class File {
...
+    }
+    return this.config.showFileHeader;
+  }
+
+  public get scssThemeable(): boolean {
+    return this.config.scssThemeable || false;
+  }
+
+  public get iosAccessControl(): string | undefined {
+    return this.config.iosAccessControl || undefined;
```
InteractiveTimmy, line 104: Same with this one, not sure if we need `|| undefined` here.

packages/@momentum-design/token-builder/src/models/file/file.ts
```diff
@@ -71,13 +71,45 @@ class File {
...
+  }
+
+  public get iosImport(): string | Array<String> | undefined {
+    return this.config.iosImport || undefined;
```
InteractiveTimmy, line 108: Same with this one, not sure if we need `|| undefined` here.

InteractiveTimmy (approved):
LGTM! non-blocking comments.

InteractiveTimmy (approved):
LGTM!

PR #160, opened 2022-11-23
Design token updates

seheekim1009:
A lot of incorrect HEX, core value, description, and inconsistent structures are fixed for dark, light, and common. 
Review is already done by Natalia, you can go ahead and approve it ;)

PR #169, opened 2022-11-29
feat(remove-defualt): defualt typo in schemas

nataliadelmar:
# Description

Schemas had a typo on default (using 'defualt'). Correcting it in this PR so that https://github.com/momentum-design/momentum-design/pull/160 can be testes against correct schemas.

# Notes
Im only changing the theme/webex files so that the pipeline in this PR passes. But I do not intend this changes to be permanent, but overwritten by design-token-updates branch PRs


InteractiveTimmy (approved):
LGTM!

PR #170, opened 2022-11-29
fix(lint): eslint in monorepo

cscheffauer:
# Description

- Currently in the existing packages VS Code can't detect the ESLint config. Therefore the existing packages are going to extend the root eslint config, which will allow packages to override & add new rules if necessary.
- Fixed the plop script to also copy DOT files correctly over, to ensure when a new package is added the eslint config is set correctly.
- Cleaned up the `package.json` (we don't have example package)

## Links

https://plopjs.com/documentation/#addmany
https://github.com/mrmlnc/fast-glob#dot


InteractiveTimmy (approved):
LGTM!

PR #173, opened 2022-11-30
feat: added flows to assets builder

cscheffauer:
# Description

- Changed assets builder to allow flows
- Flows are running in series, such that they can be chained in the future in the config to convert svg -> css -> scss -> font etc.
- Added async utils for running promises in series
- Added Flow class abstraction for easier use

**Note:**
To test locally, build the builder and then build the icons package. 

**Upcoming work:**
- Add renaming step
- Extend Transformer class to allow different transforms, like SVG to CSS, etc.

### DIFF
--- a/packages/@momentum-design/builder/FORMATS.md
+++ b/packages/@momentum-design/builder/FORMATS.md
@@ -1,0 +1,5 @@
+# Supported Output Formats
+
+Optimised SVG
+TTF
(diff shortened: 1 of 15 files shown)


InteractiveTimmy (dismissed):
LGTM!

torgeadelin (approved):

PR #174, opened 2022-12-01
feat: added transformer abstract to assets builder

cscheffauer:
# Description

- Abstracted Transformer to Factory Pattern
- Improved config & logging
- Added replace pattern to allow for file name changes (like `_` to `-`)

### DIFF
--- a/packages/@momentum-design/builder/src/assets/builder/transformer/svg-transformer.ts
+++ b/packages/@momentum-design/builder/src/assets/builder/transformer/svg-transformer.ts
@@ -33,9 +33,7 @@ class SVGTransformer extends Transformer {
    * Transform the passed in files by optimising each of them
    */
   public override transformFiles() {
-    this.logStartTransform();
     this.outputFiles = this.inputFiles?.map((file) => this.optimize(file));
-    this.logEndTransform();
   }
 }
 
--- a/packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
+++ b/packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
@@ -45,23 +45,10 @@ class Transformer {
   }
 
   /**
-   * Log method to log the start of the transform
-   */
-  protected logStartTransform() {
-    this.logger.debug(`Started transform of format ${this.format.type}.`);
-  }
-
-  /**
-   * Log method to log the end of the transform
-   */
-  protected logEndTransform() {
-    this.logger.debug(`Finished transform of format ${this.format.type}.`);
-  }
-
-  /**
    * No-op method, which has to be overridden by instances
    */
   public transformFiles(): void {
+    this.logger.warn(`No transform rule found for the provided format ${this.format.type}. Skipping...`);
   }
 
   /**
@@ -76,11 +63,10 @@ class Transformer {
     this.inputFiles = files;
     this.outputFiles = files;
 
+    this.logger.debug(`Started transform of format ${this.format.type}.`);
     this.transformFiles();
+    this.logger.debug(`Finished transform of format ${this.format.type}.`);
 
-    if (this.inputFiles === this.outputFiles) {
-      this.logger.warn(`No transformation applied for the provided format ${this.format.type}`);
-    }
     return this.outputFiles;
   }
 }


packages/@momentum-design/builder/src/assets/builder/builder.ts
```diff
@@ -20,6 +20,8 @@ const logger = Logger.child(generateMetadata(PACKAGE, CONSTANTS.TYPE));
  * @beta
  */
 class Builder extends CoreBuilder {
+  buildName: string;
+
```
torgeadelin, line 24: can we add docs for the class properties?

packages/@momentum-design/builder/src/assets/builder/constants.ts
```diff
@@ -7,6 +7,7 @@ const TYPE: string = 'assets';
 
 const FORMATS = {
   OPTIMISED_SVG: 'OPTIMIZED_SVG' as const,
+  CSS: 'CSS' as const,
```
torgeadelin, line 10: is the `builder` package targeted for `assets` only? If so, all these FORMATS are just describing the formats it outputs/can process/build for assets?

packages/@momentum-design/builder/src/assets/builder/file-handler.ts
```diff
@@ -24,6 +29,12 @@ const logger = Logger.child(generateMetadata(PACKAGE, `${CONSTANTS.TYPE}-file-ha
  * @beta
  */
 class FileHandler {
+  replacePatterns?: Array<ReplacePattern>;
```
torgeadelin, line 32: docs

packages/@momentum-design/builder/src/assets/builder/file-handler.ts
```diff
@@ -35,16 +46,39 @@ class FileHandler {
     return path.join(dirPath, path.basename(filePath));
   }
 
+  public sanitiseFileName(filePath: string): string {
```
torgeadelin, line 49: use american english (`sanitize`)

packages/@momentum-design/builder/src/assets/builder/transformer/svg-transformer.ts
```diff
@@ -0,0 +1,41 @@
...
+import CONSTANTS from '../constants';
+import type { FileType } from '../file-handler';
+import Transformer from './transformer';
+
+const PACKAGE = 'builder';
+const logger = Logger.child(generateMetadata(PACKAGE, `${CONSTANTS.TYPE}-svg-transformer`));
+
+/**
+ * The SVGTransformer class.
+ *
```
torgeadelin, line 12: please add more description

packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
```diff
@@ -0,0 +1,56 @@
...
+ * The Transformer class.
+ *
+ * @beta
+ */
+class Transformer {
+  format: Formats;
+
+  inputFiles?: Array<FileType>;
+
+  outputFiles?: Array<FileType>;
```
torgeadelin, line 22: docs missing

packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  /**
+   * No-op method, which has to be overridden by instances
+   */
+  public transformFiles() {
+
+  }
+
+  public run(files: Array<FileType>): Array<FileType> {
```
torgeadelin, line 43: docs missing

packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.inputFiles = files;
+    this.outputFiles = files;
+
+    this.transformFiles();
+
+    if (this.inputFiles === this.outputFiles) {
+      logger.warn(`No transformation applied for the provided format ${this.format.type}`);
```
torgeadelin, line 50: `Transformation had no change effects for the provided input and format ${type}`.

packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  protected logEndTransform() {
+    logger.debug(`Finished transform of format ${this.format.type}.`);
+  }
+
+  /**
+   * No-op method, which has to be overridden by instances
+   */
+  public transformFiles() {
```
torgeadelin, line 39: not sure how i feel about this.

torgeadelin: add the debugging directly before and after running `this.transformFiles();` in the `run` method.

torgeadelin (changes requested):
Looking good, just some small things.

packages/@momentum-design/builder/src/assets/builder/flow.ts
```diff
@@ -52,7 +49,7 @@ class Flow {
      */
     public read(): Promise<this> {
       return new Promise((resolve, reject) => {
-        logger.info(`Started read step of flow '${this.id}'.`);
+        logger.debug(`Started read step of flow '${this.id}'.`);
```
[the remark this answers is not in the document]
cscheffauer: i thought we can then decide on CI to exclude or include debug messages if we want.
this will allow filtering them out if needed, such that we have a bit more levels of logs based on how much info they provide.

packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
cscheffauer: that sounds more complicated to me - like what are `change effects`? :D

packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
cscheffauer: If its in the run method, the logging of `Started transform` & `Finished transform` will always happen, even if there is no specific transformer applied. Thats why I kept it optional, such that each Transformer class can define when to log the start and end of transforming.

packages/@momentum-design/builder/src/assets/builder/constants.ts
cscheffauer: yeah, if u take a look at the path, the builder class is part of the assets folder. 
For consumers, they have to set the "type: assets" in their config (=json file) to make use of of the assets builder.
If they want a different builder, they would set a different type.

I can explain that in detail in a call if u need more info.

packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
torgeadelin: Maybe just `had no effect` would be more clear?

packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
torgeadelin: > If its in the run method, the logging of `Started transform` & `Finished transform` will always happen, even if there is no specific transformer applied. Thats why I kept it optional, such that each Transformer class can define when to log the start and end of transforming.

I see.. To be honest, we can log something like "No transform rule found. Skipping..." in the default implementation of `transformFiles` so that when we don't override it it will output something like:
```
$ Started transform...
$ No transform rule found. Skipping...
$ Finished transform...
```

packages/@momentum-design/builder/src/assets/builder/transformer/factory.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { Formats } from '../types';
+import SVGTransformer from './svg-transformer';
+import Transformer from './transformer';
+
+type Transformers = Transformer | SVGTransformer;
+
+const transformerFactory = {
+  createTransformer(format: Formats): Transformers {
+    switch (format.type) {
+      case CONSTANTS.FORMATS.OPTIMISED_SVG:
```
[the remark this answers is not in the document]
cscheffauer: sticked to the same format as in react-v2

packages/@momentum-design/builder/src/assets/builder/transformer/transformer.ts
cscheffauer: done

InteractiveTimmy (approved):
LGTM! Pending @torgeadelin's comments.

torgeadelin (approved):
looking great!

PR #184, opened 2022-12-06
Feat/assets/woff transformer

torgeadelin:
# Description
Added
`TTF` -> `WOFF` Font Transform
Changed FileHandler to support multiple encoding configurations.

### DIFF
--- a/packages/@momentum-design/builder/src/assets/builder/file-handler.ts
+++ b/packages/@momentum-design/builder/src/assets/builder/file-handler.ts
@@ -6,7 +6,7 @@ import {
 } from '@momentum-design/telemetry';
 import path from 'path';
 import CONSTANTS from './constants';
-import type { FileType, ReplacePattern } from './types';
+import type { Encoding, FileType, ReplacePattern } from './types';
 
 const logger = Logger.child(generateMetadata(CONSTANTS.PACKAGE, `${CONSTANTS.TYPE}-file-handler`));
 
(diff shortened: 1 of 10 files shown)


packages/@momentum-design/icons/config/momentum.json
```diff
@@ -81,11 +81,30 @@
           "destination": "./dist/fonts",
           "format": {
             "type": "TTF",
+            "encoding": {
+              "read": "utf-8",
+              "write": null
```
cscheffauer, line 86: what is null encoding? doesn't make really sense to me

cscheffauer (approved):
question on encoding

chrisadubois (approved):
Small comment. Also produces woff2?

packages/@momentum-design/icons/config/momentum.json
torgeadelin: ttf holds binary information, and that's what we need for woff and woff2. If i don't add the null, it encodes it as utf-8 and when i read it again in the woff transform it's invalid.

packages/@momentum-design/builder/src/assets/builder/file-handler.ts
```diff
@@ -112,9 +118,9 @@ class FileHandler {
    * @param file - File to read data from
    * @returns Promise, which resolves to the file object with data if successful
    */
-  public readFile(file: FileType): Promise<FileType> {
+  public readFile(file: FileType, encoding: string): Promise<FileType> {
```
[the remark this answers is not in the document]
torgeadelin: woff2 is separate

PR #185, opened 2022-12-06
Feat/assets/woff2 transformer

torgeadelin:
# Description

`TTF` -> `WOFF2` transformer


chrisadubois (approved):
IMO it’d be good to add logs to all of this

and if you’re feeling fancy … maybe a metric (similar to how we did it before) so we know how often this is occurring and a general loose idea of what it saves us by doing it

PR #221, opened 2022-12-08
Asset Automation 2022-12-08T20:39:45.162Z

chrisadubois:
feat(assets): Asset Automation 2022-12-08T20:39:45.162Z


ArthusLiang (commented):
Does figma optimized the path now? Why the attribute d in path is different?

cscheffauer:
Pls DO NOT MERGE yet. there need to be some checks before

cscheffauer:
> Does figma optimized the path now? Why the attribute d in path is different?

@ArthusLiang maybe - or designers have updated the icons and have forgotten to export them. since the new plugin does export all icons, it might catch some unpublished changes as well

PR #238, opened 2022-12-09
Assets final READY

chrisadubois:
1. Sets up legacy icons
2. Sets up legacy illustrations
3. Adds in new icons
4. Adds in new illustrations
5. Does the optimization configurations


packages/@momentum-design/figma-plugin-assets-export/dist/index.js
```diff
@@ -1,376 +1 @@
-"use strict";
```
cscheffauer, line 1: i don't see any changes to the figma-plugin source code, so where is this change coming from? 
can u pls check that? if there is no change in the plugin source code, the dist should not be updated either.

cscheffauer (changes requested):

cscheffauer (approved):

PR #240, opened 2022-12-12
Feat/docs/add icon+illustration assets

torgeadelin:
# Description

- refactor theme tokens tables
- add theme tokens (stable + next)
- added icons/illustration section
- search functionality
- filter by weight


packages/@momentum-design/builder/src/assets/builder/constants.ts
```diff
@@ -20,6 +20,7 @@ const FORMATS = {
   TTF: 'TTF' as const,
   WOFF: 'WOFF' as const,
   WOFF2: 'WOFF2' as const,
+  MANIFEST: 'MANIFEST' as const,
```
InteractiveTimmy, line 23: why the `as const`?

InteractiveTimmy (approved):
LGTM! one question, but not blocking.

packages/@momentum-design/builder/src/assets/builder/constants.ts
torgeadelin: just followed the pattern.. haha

PR #244, opened 2022-12-12
fix(automation): deploy docs diff

InteractiveTimmy:
# Description

The scope of the changes in this pr are to add a unique hash to each docs deploy so that `git add` does not `exit 1`.


torgeadelin (approved):
LGTM

PR #246, opened 2022-12-12
Feat/autogenerated api docs/in stro

torgeadelin:
# Description

- add README docs
- add API reference docs for the packages in monorepo


chrisadubois (approved):
re your last bullet, this is called frontmatter and should not interfere with astro. We will have to add this information back in at some point

torgeadelin:
> re your last bullet, this is called frontmatter and should not interfere with astro. We will have to add this information back in at some point

It does tho, when importing those readmes that have 
```
---
package: ...
---
```

it thinks it's Astro syntax when it's not, and it fails to build.

chrisadubois:
It's not astro syntax. It's markdown syntax

https://daily-dev-tips.com/posts/what-exactly-is-frontmatter/

PR #259, opened 2022-12-13
fix: small issues in figma plugin

cscheffauer:
# Description

- UI: Polished the look and feel of the export status
- UI: The export status will clear after the assets type switched (like after exporting icons, switching to illustrations will clear the export status paragraphs)
- Fix: Fixed the PR Link once export is successful (was not pointing to the right url)
- Quality of life: The asset type selection in the export section will now persist, if changing tabs
- Chore: moved `<a ... />` tag out to separate Link component


chrisadubois (approved):
We like *need* to publish this so that people get updates without having to download dist

PR #412, opened 2023-08-14
fix(actions): attempt at fixing cache-dependency-path

chrisadubois:
# Description

Adding cache-dependency-path to setup node.

I think this will fix https://github.com/momentum-design/momentum-design/actions/runs/5857942350/job/15888490025

## Links

https://github.com/momentum-design/momentum-design/actions/runs/5857942350/job/15888490025

### DIFF
--- a/.github/workflows/deploy-package.yml
+++ b/.github/workflows/deploy-package.yml
@@ -25,7 +25,7 @@ jobs:
           node-version-file: '.nvmrc'
           registry-url: 'https://registry.npmjs.org'
           cache: 'yarn'
-          cache-dependency-path: node-modules-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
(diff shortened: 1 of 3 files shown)


InteractiveTimmy (changes requested):
the `cache-dependency-path` should be the path to the localized `package-lock.json` or `yarn.lock` file, the file at the value of this parameter is read by `actions/setup-node@3`, then the contents are processed similar to the native GitHub Actions `hashFiles()` call within `actions/setup-node@3`.

See [actions/setup-node@3 documentation](https://github.com/actions/setup-node#usage). Let me know if you disagree and believe these initial changes are correct.

See this [related issue](https://github.com/actions/setup-node/issues/801), it seems this was an API-breaking change they didn't account for.

InteractiveTimmy (approved):
LGTM per documentation of the consumed workflow.

PR #434, opened 2023-09-29
feat: add swift icons format

cscheffauer:
# Description
- Added new svg glyphs transformer to allow sharing glyphs data between ttf and swift
- Added new swift transformer to export swift file based on handlebar hbs file
- Added hbs template file to icons lib 
- Modified icons build config to export new swift format as part of dist
- Added handlebars & lodash to builder package.json


chrisadubois:
Overall really agree where this is going

Couple of questions before a deeper review

1. Why handlebars?
2. Why are we adding this new template to generate swift files? I believe that we have been using some different patterns to generate our outputs but this one goes so far as to create a new template - can you explain that a bit further?
3. Shouldn't there be a specific command in the package json to reference the swift file generator?

cscheffauer:
> Overall really agree where this is going
> 
> Couple of questions before a deeper review
> 
> 1. Why handlebars?

It is used in old Momentum-ui so it was easy to make sure our new library matches the old one. Also it allows quite easy flexibility in terms of modifying the content - you would just need to adjust the template and done.

> 2. Why are we adding this new template to generate swift files? I believe that we have been using some different patterns to generate our outputs but this one goes so far as to create a new template - can you explain that a bit further?

That template was copied over from momentum-ui to follow that. With the existing builder for tokens for example, where we also generate swift files, i don't see how that can be modified. For icons we need a specific structure of the swift file, so something like handlebars would have been needed anyways. We could look into merging the tokens and icons generation of swift later on if necessary.

> 3. Shouldn't there be a specific command in the package json to reference the swift file generator?

nope, that is handled over the config in icons itself (check the momentum.json file changes in that package)

chrisadubois (approved):

PR #527, opened 2024-02-20
feat: svg path merge transformer

bhanusengar:
# Description

SVG path merge transformer


cscheffauer:
Closing this, since done in https://github.com/momentum-design/momentum-design/pull/532 more efficiently.
Keeping it in closed state for reference, in case we need to go back to this solution.

PR #551, opened 2024-03-19
fix: illustrations icon fix for docs module

gauravchauhan3030:
illustrations icon fix for docs module


packages/@momentum-design/docs/src/components/IllustrationTable/IllustrationTable.tsx
```diff
@@ -79,7 +80,8 @@ export const Pagination = () => {
 
   const paginatedItems = useMemo(
     () => Object.entries(illustrationsManifest)
-      .filter(([key]) => (query ? key.includes(query) && key.includes(size) : key.includes(size)))
+      // eslint-disable-next-line max-len
+      .filter(([key]) => (query ? key.includes(query) && key.includes(IllustrationSize[size as unknown as keyof typeof IllustrationSize]) : key.includes(IllustrationSize[size as unknown as keyof typeof IllustrationSize])))
```
bhanusengar, line 84: In line 84 i think you can try instead of rewriting  `keyof typeof IllustrationSize` twice
`type IllustrationSIZE = keyof typeof IllustrationSize;` on top  and use type variable

bhanusengar (approved):

packages/@momentum-design/docs/src/components/IllustrationTable/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const IllustrationSize = {
+  320: 'threetwozero',
+  192: 'oneninetwo',
+  120: 'onetwozero',
+  60: 'sixzero',
+  '': '',
+};
+
+export type illustrationSizeType = keyof typeof IllustrationSize;
```
cscheffauer, line 9: Types and constants shouldn't be mixed in a module.
Either split the files into 2, or you can define the type in the IllustrationTable at the top directly.
Also additionally: Types should always be Capitalized, like `IllustrationSizeType`

cscheffauer (changes requested):

cscheffauer (approved):

PR #553, opened 2024-03-20
test: unit test case for swift transformer

bhanusengar:
# Description

Unit test case for swift transformer.


gnanammcp (approved):

packages/@momentum-design/builder/src/assets/builder/transformer/swift-transformer.unit.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      const transformHbsSpy = jest.spyOn(Utils, 'transformHbs').mockReturnValue(
+        new Promise((resolve) => { resolve(templateSpy); }),
+      );
+      await transformer.transformFilesAsync();
+      expect(transformFilesAsyncSpy).toBeCalledTimes(1);
+      expect(transformHbsSpy).toBeCalledTimes(1);
+      expect(transformer.outputFiles).toStrictEqual([
+        {
+          data: [inputFilesData[0]],
+          distPath: path.join('/dist', 'MyFont'),
```
cscheffauer, line 52: u can use `FORMAT.config.fileName` here instead of `MyFont`.
Also the `/dist` can be extracted to a constant and being used in line 18 and 30 and here

packages/@momentum-design/builder/src/assets/builder/transformer/swift-transformer.unit.test.ts
```diff
@@ -0,0 +1,59 @@
+import path from 'path';
+import type { Formats, SwiftFormat } from '../types';
+import Transformer from './transformer';
+import SwiftTransformer from './swift-transformer';
+import * as Utils from '../utils';
+
+const inputFilesData = { 0: {
```
cscheffauer, line 7: not quite sure why `{ 0: ...` is necessary here - can u try to create the inputFilesData without it, just being a non-nested object?

cscheffauer (changes requested):

packages/@momentum-design/builder/src/assets/builder/transformer/swift-transformer.unit.test.ts
bhanusengar: Changes added.

packages/@momentum-design/builder/src/assets/builder/transformer/swift-transformer.unit.test.ts
bhanusengar: Changes addressed.

cscheffauer (approved):

PR #555, opened 2024-03-22
feat: test cases for svg glyph transformer file

gauravchauhan3030:
test cases for svg glyph transformer file


gnanammcp (approved):

packages/@momentum-design/builder/src/assets/builder/transformer/svg-glyphs-transformer.unit.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      expect(transformer.outputFiles).toEqual([
+        {
+          data: expect.any(String),
+          distPath: path.join('dist', 'MyFont'),
+          srcPath: '',
+        },
+      ]);
+    });
+  });
+  describe('getNextCodepoint', () => {
```
cscheffauer, line 41: Can u split this test out to a separate test file, along with the utils? like the getNextCodepoint is part of a different file, so we would want to have a separate test file for that

packages/@momentum-design/builder/src/assets/builder/transformer/svg-glyphs-transformer.unit.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    it('should extend Builder', () => {
+      expect(transformer instanceof Transformer).toBe(true);
+    });
+    it('should mount the format provided to the class object', () => {
+      expect(transformer.format).toBe(FORMAT);
+      expect(transformer.destination).toBe('dist');
+    });
+  });
+  describe('transformFilesSync', () => {
+    it('transform the passed in files by optimizing each of them', () => {
```
cscheffauer, line 27: the `generateGlyphsData` does have more functionality in it, we would want to test (in specific line 25, where we filter out input files if they include numbers). We would want to have a separate test case which tests that inputFiles with a srcPath with a number in it will be filtered out

cscheffauer (changes requested):

cscheffauer (approved):

PR #614, opened 2024-04-26
Brand visual export png

bhanusengar:
# Description

The scope of this to automatically detect if there is a image inside of the component in Figma and then export that component as a PNG (instead of a SVG)

### DIFF
--- a/packages/@momentum-design/figma-plugin-assets-export/src/plugin/models/component.ts
+++ b/packages/@momentum-design/figma-plugin-assets-export/src/plugin/models/component.ts
@@ -1,5 +1,6 @@
 /* eslint-disable no-restricted-syntax */
 /* eslint-disable no-undef */
+import { Buffer } from 'buffer';
 import { CONSTANTS } from '../constants';
 import { normaliseObject } from '../utils/object';
 import type { Asset, AssetSetting } from '../../shared/types';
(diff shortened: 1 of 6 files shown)


packages/@momentum-design/figma-plugin-assets-export/src/plugin/models/component.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable no-restricted-syntax */
 /* eslint-disable no-undef */
+import { Buffer } from 'buffer';
```
arjunkesava, line 3: Please import this npm package.
If it is already imported, then please point out on where it's located.

packages/@momentum-design/figma-plugin-assets-export/src/plugin/models/component.ts
```diff
@@ -91,12 +92,18 @@ class Component {
...
+      const exportSettings = this.isNodeContainingImage(this.node)
+        ? this.assetSetting.input.asset.exportSettingsImage
+        : this.assetSetting.input.asset.exportSettings;
       this.node
-        .exportAsync(this.assetSetting.input.asset.exportSettings)
+        .exportAsync(exportSettings)
         .then((uint8Array: Uint8Array) => {
+          const imageSvgData = this.isNodeContainingImage(this.node)
+            ? Buffer.from(uint8Array).toString('base64')
+            : String.fromCharCode.apply(null, uint8Array as any);
```
arjunkesava, line 103: Instead of calling the same method two times, we can use a variable and call it once.

```typescript
      const isImageExists = this.isNodeContainingImage(this.node);
      const exportSettings = isImageExists
        ? this.assetSetting.input.asset.exportSettingsImage
        : this.assetSetting.input.asset.exportSettings;
      this.node
        .exportAsync(exportSettings)
        .then((uint8Array: Uint8Array) => {
          const imageSvgData = isImageExists
            ? Buffer.from(uint8Array).toString('base64')
            : String.fromCharCode.apply(null, uint8Array as any);
```

packages/@momentum-design/figma-plugin-assets-export/src/plugin/models/component.ts
```diff
@@ -106,6 +113,20 @@ class Component {
     });
   }
 
+  /**
+   * This function checks if node contains any IMAGE fill
+   *
+   * @returns true, if this.node contains any IMAGE fill, otherwise false
+   */
+  isNodeContainingImage(node: ComponentNode) {
```
arjunkesava, line 121: Please add new unit specs for this method and update unit specs for the `asset` method in `component.unit.spec.ts`.

arjunkesava:
The work reg this PR is carried on #662. Hence closing this one.

PR #625, opened 2024-05-24
fix: hex values for black alpha 4 and 7 generating invalid css

NotNestor:
# Description

Update the values for black alpha 4 and 7 to be a valid colour. Incorrectly had 7 digits instead of 8. This means the generated scss files cannot be used


cscheffauer:
closing, since this has been fixed by the automatic push from Figma

PR #662, opened 2024-07-04
fix: Brand Visual export png and svg

arjunkesava:
# Description

- This PR is a fix on the existing PR: https://github.com/momentum-design/momentum-design/pull/614
- 13937cd33d8d091d7179561e5565f21f641d0757 - this commit fixes the file extension so that png will also be added.

## Links

Here is the example PR which is the result of this PR: https://github.com/momentum-design/momentum-design/pull/661
We can now upload both png and svg files into the brand visuals export library.


packages/@momentum-design/figma-plugin-assets-export/src/plugin/models/component.ts
```diff
@@ -106,6 +112,20 @@ class Component {
     });
   }
 
+  /**
+   * This function checks if node contains any IMAGE fill
+   *
+   * @returns true, if this.node contains any IMAGE fill, otherwise false
+   */
+  isNodeContainingImage(node: ComponentNode) {
```
cscheffauer, line 120: issue: any source code change should come with a unit test / test change.

packages/@momentum-design/figma-plugin-assets-export/src/shared/types.ts
```diff
@@ -45,6 +45,18 @@ interface ExportSettingsSVG {
   readonly svgSimplifyStroke?: boolean; // defaults to true
 }
 
+interface ExportSettingsConstraints {
+  type: 'SCALE' | 'WIDTH' | 'HEIGHT'
+  value: number
+}
+interface ExportSettingsImage {
+  readonly format: 'JPG' | 'PNG';
```
cscheffauer, line 53: issue: my recommendation would be to only set format 'PNG' here, since we do not need to support JPG for now, this PR was only tested with PNG - because of that we should not introduce a type which gives future developers the feeling that we also support JPG (which we don't)

packages/@momentum-design/figma-plugin-assets-export/src/ui/models/github.ts
```diff
@@ -157,8 +157,15 @@ class Github {
    */
   private calculateChangesChunks(data: AssetChunks): ChangesChunks {
     return data?.map((assets, index) => ({
-      files: assets.reduce((accum: { [key: string]: string }, cur: Asset) => {
-        accum[[this.config.git.gitRepoFilePath, cur.path].join('/')] = cur.data;
+      files: assets.reduce((accum: { [key: string]: any }, cur: Asset) => {
+        if (cur.path.match(/\.(jpg|jpeg|png|gif)$/i)) {
```
cscheffauer, line 161: issue: we only tested png, thats why my suggestion would be to put png only here for now to not cause confusion.

cscheffauer (changes requested):
looking good - some issues / gaps to work on before merging

packages/@momentum-design/figma-plugin-assets-export/src/plugin/models/component.ts
arjunkesava: Yes, I will add unit test in this ticket: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-483011

packages/@momentum-design/figma-plugin-assets-export/src/plugin/models/component.unit.test.ts
```diff
@@ -50,15 +50,13 @@ describe('@momentum-design/figma-plugin-assets-export - models.Component', () =>
...
+      expect(component.assetName).toEqual('display-sf-rtl-bold-state-twofour');
     });
   });
   describe('asset', () => {
     it('check the appropriate result', async () => {
-      componentNodeMock.exportAsync.mockImplementation(() => Promise.resolve({
-        path: `${DESTINATION}/display-sf-rtl-bold-state-twofour.svg`,
-        data: String.fromCharCode.apply(null, new Uint8Array(8) as any),
-      }));
+      componentNodeMock.exportAsync.mockImplementation(() => Promise.resolve(''));
```
cscheffauer, line 58: mocking the implementation to resolve with a empty string doesn't match the expected return value of the exportAsync. please mock this properly.

cscheffauer (dismissed):

cscheffauer (approved):

PR #666, opened 2024-07-09
chore: add CSS format support to builder

rajrdas:
# Description

- *Added support to builder to generate CSS file for the icons*
This PR adds the CSS-transformer, which generates the MomentumFontIcon.css file from all the momentum-design icons. 

## Links

***required*** - *Links to resources related to this Pull Request*


packages/@momentum-design/builder/src/assets/builder/transformer/css-transformer.ts
```diff
@@ -0,0 +1,88 @@
+import path from 'path';
+import { promisify } from 'util';
```
cscheffauer, line 2: we aren't using promisify anywhere else and there shouldn't be a reason to do so. Just use the normal async/await or Promise.then syntax for fs.readfile.

packages/@momentum-design/builder/src/assets/builder/transformer/css-transformer.ts
```diff
@@ -0,0 +1,88 @@
...
+   * @param glyphData - glyph data to be used
+   * @returns object with css data
+   */
+  private async generateCssData(): Promise<{ data: string }> {
+    const glyphData = this.getGlyphData();
+    const template = await transformHbs(path.resolve(this.format.config.hbsPath));
+    return {
+      data: template({
+        glyphsData: Object.values(glyphData),
+        woffUrl: await this.getFileNameWithHash(this.format.config.woffPath),
```
cscheffauer, line 59: is there a reason why we hash the filename? i think there is no need for it imho.

packages/@momentum-design/icons/config/templates/momentum.css.hbs
```diff
@@ -0,0 +1,261 @@
+@font-face {
```
cscheffauer, line 1: has this been tested locally? i tried your PR locally and ran yarn icons build and it did create a empty css file with no content. can u check again please?

packages/@momentum-design/builder/src/assets/builder/types.ts
```diff
@@ -22,7 +22,7 @@ export interface OptimizedSVGFormat {
  *
  * @beta
  */
-interface CSSFormat {
+export interface CSSFormat {
   type: typeof CONSTANTS.FORMATS.CSS;
   // TODO: add CSS config type:
```
cscheffauer, line 27: this type can be fixed now

cscheffauer (changes requested):

packages/@momentum-design/icons/config/templates/momentum.css.hbs
rajrdas: This has worked for me.

cscheffauer (approved):

PR #673, opened 2024-07-16
fix(asset): check image export settings before accepting

arjunkesava:
# Description

- The `exportSettingsImage` is an optional variable which is mainly used to export images in Brand Visuals Library.
- When we try to export Icons and Illustrations, there is a chance of this variable being undefined which leads to plugin failure.
- This PR will fix the issue by double checking the `exportSettingsImage` variable before we consume it.
- Plus, I've added the `dist/index.js` file which is generated via `yarn build:plugin`.


2020Deeya (approved):
LGTM

2020Deeya (approved):
LGTM

PR #694, opened 2024-07-25
chore: update folder structure for tokens and token-builder

NotNestor:
# Description

Update folder structure for tokens and token builder. 
@momentum-design/tokens -> assets/tokens
@momentum-design/token-builder -> tools/token-builder

## Links

N/A


cscheffauer (changes requested):
The folder changes itself are fine. There needs to be an update to the "workspace" section in the `package.json` scripts. Line 36, 37 - the paths need to be updated, otherwise scripts can not be run correctly.

cscheffauer (approved):

PR #707, opened 2024-07-30
feat: upgrade all the dependencies in the package.json of icons

2020Deeya:
# Description

Upgrade all the dependencies in the package.json of icons


packages/@momentum-design/icons/package.json
```diff
@@ -23,6 +23,6 @@
   },
   "devDependencies": {
     "@momentum-design/builder": "workspace:^",
-    "rimraf": "^3.0.2"
+    "rimraf": "^6.0.1"
```
cscheffauer, line 26: This change also will cause yarn.lock file changes, which will be needed as part of the commit

cscheffauer (changes requested):

packages/@momentum-design/icons/package.json
2020Deeya: Resolved

2020Deeya:
Dependencies are already merged into main using syncpack. Hence, closing the PR.

PR #708, opened 2024-07-30
feat: upgrade all the dependencies in the package.json of illustrations

2020Deeya:
# Description

Upgrade all the dependencies in the package.json of illustrations


packages/@momentum-design/illustrations/package.json
```diff
@@ -22,6 +22,6 @@
   },
   "devDependencies": {
     "@momentum-design/builder": "workspace:^",
-    "rimraf": "^3.0.2"
+    "rimraf": "^6.0.1"
```
cscheffauer, line 25: For this change there should also be a yarn.lock file change, which needs to be committed as well.

cscheffauer (changes requested):

packages/@momentum-design/illustrations/package.json
2020Deeya: Resolved

2020Deeya:
Dependencies are already merged into main using syncpack. Hence, closing the PR.

PR #715, opened 2024-08-01
feat(components): added component lib foundation

cscheffauer:
# Description

- This adds the foundation for the Component Lib work
- Adding sample components / base components, including provider concept
- Added folder structure, including e2e tests, stories, etc
- Components are not published

There are still outstanding todo's and things to be cleaned up. This is only the foundation to keep going.


arjunkesava (dismissed):

packages/@momentum-design/fonts/src/css/fonts.css
```diff
@@ -0,0 +1,4 @@
+@font-face {
+    font-family: "Momentum";
```
arjunkesava, line 2: Q: Doesn't the font name should be "Inter" here?

cscheffauer: good question: if we keep the Font Family as Momentum, for the consumer it doesn't really matter than what the font "really" is. That means whenever we change the font itself (lets say from "Inter" to "XYZ"), we can just change the font-face in here and there will be no changes for consumers, as long as they import the font-family "Momentum".

PR #721, opened 2024-08-02
chore(figma): reduce asset chunk limit from 250 to 50

arjunkesava:
# Description

- This PR can be considered as an update on the existing https://github.com/momentum-design/momentum-design/pull/688 (see desc).
- We reduce the limit from 250 to 50 as `Icons Library` design file in figma has a huge number of small files and all of them are in SVG format (which is literally a gigantic string).
- The github rest API limitations are very tight and strict, due to which we have to reduce the chunk count.
- Reduction from 250 to 50 will further increase the number of commits for a auto generated PR and it will also take a bit more time than usual to process all of these commits, but still it's a good way rather than a failed git rest api.
- Here is the PR: https://github.com/momentum-design/momentum-design/pull/720 which was generated by reducing the file size.
- Delete `dist/` directory in the figma-plugin-assets-export/* package.


package.json
```diff
@@ -41,7 +41,8 @@
     "common": "yarn workspace @momentum-design/common",
     "figma-plugin-assets-export": "yarn workspace @momentum-design/figma-plugin-assets-export",
     "animations": "yarn workspace @momentum-design/animations",
-    "brand-visuals": "yarn workspace @momentum-design/brand-visuals"
+    "brand-visuals": "yarn workspace @momentum-design/brand-visuals",
+    "figma": "yarn figma-plugin-assets-export"
```
arjunkesava, line 45: This addition is only meant for developers to reduce the command line, and it's optional.

cscheffauer: i think if we want to reduce the typing, `assets-export` might be better, cause we might have more `figma` plugins later on. Please remove the existing `figma-plugin-assets-export` script

packages/@momentum-design/figma-plugin-assets-export/dist/index.html
```diff
@@ -65,7 +65,7 @@
...
...calculateChangesChunks(t),i=o==null?void 0:o.length,u=this.generateChunksOfFilesWhichShouldBeDeleted(l,r,i),s=[...o,...u];if(!s.length)return Promise.reject(new Error("No changes found!"));switch(this.config.mode){case"ADD":return o;case"DELETE_ONLY":return u;case"FULL_SYNC":return s;default:return o}}async pullRequest(t){return this.octokit.createPullRequest({owner:this.config.git.githubOwner,repo:this.config.git.gitRepo,title:this.config.git.prTitle,body:this.config.git.prMessage,forceFork:!0,createWhenEmpty:!1,head:this.config.git.gitBranch,changes:t||[]})}}const bh=({assetSetting:e,setExportStatus:t,exportStatus:n,gitConfig:r,mode:l,inputConfig:o,assetChunks:i})=>{const[u,s]=I.exports.useState(),[a,m]=I.exports.useState(),h=y.useRef();I.exports.useEffect(()=>{r&&(h.current=new Jh(r,l,o))},[r,l,o]);const p=S=>{t("Complete"),Wp(parent,S),m(S)},g=S=>{t("Failure"),s(S)},w=()=>{e&&(m(null),s(void 0),t("Clicked"),Vp(parent,e))};return I.exports.useEffect(()=>{n==="Calculating Changes In Progress"&&h.current&&h.current.generateChanges(i).then(S=>{var O;t("Exporting to Github In Progress"),(O=h.current)==null||O.pullRequest(S).then(f=>{if(!f){g(new Error(`No response after creating PR.
```
cscheffauer, line 68: i think we don't need to commit the dist folder anymore here, since there is no use of it.
Can u remove those 2 dist files, commit the deletion and we will create a follow up PR with the gitignore changes for this? thanks

cscheffauer (approved):

PR #725, opened 2024-08-05
fix(ci): remove validate dependencies to avoid cache issues

cscheffauer:
# Description

- The validate dependencies action does cause a runtime issue, where the wrong cache is used during CI.
- This will make sure the latest dependency update is always used for the pipeline by always using the correct cache.


2020Deeya (approved):
LGTM

PR #732, opened 2024-08-06
feat(specs): run unit tests for automation commands

arjunkesava:
# Description

- Write unit tests for the commands `get-packages` and `increment-packages`.
- Implement `dryRun` for increment packages command as we don't need to update package files for a unit spec.
- Set up the package.json scripts to run tests in the automation package.

Screenshot 2024-08-06 at 10 54 17 PM


packages/@momentum-design/automation/package.json
```diff
@@ -32,9 +32,10 @@
     "publish": "yarn publish:npmjs",
     "publish:npmjs": "echo \"script 'publish:npmjs' has not been implemented\"",
     "test": "yarn test:prebuild && yarn test:postbuild",
-    "test:postbuild": "echo \"script 'test:prebuild' has not been implemented\"",
-    "test:prebuild": "yarn test:unit",
-    "test:unit": "jest ./src/**/*.test.ts"
+    "test:prebuild": "yarn build && yarn test:unit:prebuild",
+    "test:unit:prebuild": "jest ./src/**/*.test.ts --testPathIgnorePatterns=ignore",
+    "test:postbuild": "yarn build && yarn test:unit:postbuild",
```
cscheffauer, line 37: postbuild in the pipeline runs "after" the build automatically, there is no need to build here

packages/@momentum-design/automation/src/commands/increment-packages/constants.ts
```diff
@@ -4,6 +4,10 @@ import GetPackages from '../get-packages';
 
 const OPTIONS: CommandOptions = {
   ...GetPackages.CONSTANTS.OPTIONS,
+  dryRun: {
+    describe: 'dry run mode, does not write changes to disk',
```
cscheffauer, line 8: i would add `or sends requests to NPM, other services`

packages/@momentum-design/automation/src/commands/main/main.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import { figmaPackage, packagesList, packageVersionIncrementList } from './main.fixture';
+
+describe('Main', () => {
+  describe('run on all packages', () => {
+    it('get list of all existing packages', async () => {
+      const results = await Execute.run('yarn md-automation -- --command get-packages');
+      expect(results).toBe(JSON.stringify(packagesList));
+    });
+
+    it('increment version count for all packages', async () => {
```
cscheffauer, line 11: what would be great here is to also have tests for "create-release" command

packages/@momentum-design/automation/src/commands/main/main.test.ts
```diff
@@ -0,0 +1,34 @@
+import { Execute } from '../../utils';
```
cscheffauer, line 1: having all those tests in one test file is not scalable, i think it would be better to create for each command a test file, like in get-packages, increment-packages folder etc.
You can name the test file like `XXX.postbuild-test.ts` and in the jest.config just target those test files (makes it more scalable)

cscheffauer (changes requested):

arjunkesava:
Closing this PR, as I've opened a new one at #803

PR #736, opened 2024-08-07
feat: add ios webex format tokens

brunnerjosh:
# Description

* Adds theme & core token output in the format that Webex for iOS requires 
* Output folder: `dist/ios-webex`


packages/@momentum-design/token-builder/src/models/token-builder/token-builder.ts
```diff
@@ -13,7 +13,7 @@ import {
 import { SomeJSONSchema } from 'ajv/dist/types/json-schema';
```
cscheffauer, line 13: the tests are failing, please fix them to get the pipeline green.

cscheffauer (changes requested):

cscheffauer (approved):

PR #737, opened 2024-08-08
fix(ci): types node to match node version

cscheffauer:
# Description

- Fix @types/node to match our used node engine version (20.15.1 -> @types/node: 20.14.9)
- Updated synpackrc config to enforce across packages


2020Deeya (approved):
LGTM

PR #738, opened 2024-08-08
feat(md): add steps to contribute for components pkg

arjunkesava:
# Description

- Add steps to contribute `@momentum-design/components` package.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,8 +19,19 @@ More specific information is provided in each sub-package.
 
 1. Fork the repository
 2. Clone the forked repository
-3. Verify that your forked repos are set up with the correct remote references.
-    Running git remote -v in your repository directory should return settings like these:
+    ```
+    git clone https://github.com/{username}/momentum-design.git
+    ```
+3. Navigate to the root of the repo.
+    ```
+    cd momentum-design
+    ```
+4. Setup upstream remote references in your local
+    ```
+    git remote add upstream https://github.com/momentum-design/momentum-design.git
+    ```
+5. Verify that your forked repos are set up with the correct remote references.
+    Running `git remote -v` in your repository directory should return settings like these:
 
     ```bash
     origin git@github.com:{username}/momentum-design.git (fetch)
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -52,6 +52,34 @@ Steps for creating a PR (after [First time setup](#first-time-setup) has been do
 6. **Important: Add a proper description and title to the PR - it should be formatted, human-readable and also not include the description template text anymore.**
 7. Set the *validated* label on the PR to kick off the pipeline if you have the access rights for it.
 
+## Steps to contribute for components package
```
cscheffauer, line 55: i would recommend to put this into the components package itself.

CONTRIBUTING.md
```diff
@@ -52,6 +52,34 @@ Steps for creating a PR (after [First time setup](#first-time-setup) has been do
...
+
+- Clone this repository locally:
```
cscheffauer, line 57: this is already covered by the Contributing First time setup

CONTRIBUTING.md
```diff
@@ -52,6 +52,34 @@ Steps for creating a PR (after [First time setup](#first-time-setup) has been do
...
+    ```
+    git clone https://github.com/momentum-design/momentum-design.git
+    ```
+- Go to the project and run yarn
```
cscheffauer, line 61: this is already covered by the Contributing First time setup (if not, add it)

CONTRIBUTING.md
```diff
@@ -52,6 +52,34 @@ Steps for creating a PR (after [First time setup](#first-time-setup) has been do
...
+- Install dependencies before we start components package
+    ```
+        yarn @momentum-design/token-builder build
+        yarn @momentum-design/tokens build
+        yarn @momentum-design/fonts build
+        yarn @momentum-design/common build
+        yarn @momentum-design/builder build
+        yarn @momentum-design/icons build
+    ```
+- Finally, you can run dev
```
cscheffauer, line 76: only this should be added inside of the Contributing guide of the components package itself.
also: this starts storybook, we should be very specific about what things are for. Saying `you can run dev` and `Done` is not really explaining what actually happens. For example it doesn't explain how developers can run tests, can build etc etc. Therefore i would recommend specifity

CONTRIBUTING.md
```diff
@@ -52,6 +52,34 @@ Steps for creating a PR (after [First time setup](#first-time-setup) has been do
...
+    ```
+        yarn @momentum-design/components dev
+    ```
+- Your localhost will fire up: http://localhost:6006/
```
cscheffauer, line 80: adding the localhost:6006 here i think should be avoided, since Ports can change over time. Its better to say: `Storybook will open up in your browser.`
or something similar

cscheffauer (changes requested):

cscheffauer (approved):

PR #742, opened 2024-08-09
fix(icons): clean-css-cli by moving to subpackage

cscheffauer:
# Description

- clean-css-cli is installed in root, should be installed in subpackage.

This will have no affect on the output of icons, purely development related.


2020Deeya (approved):
LGTM

PR #764, opened 2024-08-20
fix: code clean of icons and illustrations src path

2020Deeya:
# Description

Updated icons and illustrations src path with with better approach

### DIFF
--- a/packages/@momentum-design/docs/astro.config.mjs
+++ b/packages/@momentum-design/docs/astro.config.mjs
@@ -7,6 +7,9 @@ import copy from 'rollup-plugin-copy';
 import mdx from '@astrojs/mdx';
 import path from 'path';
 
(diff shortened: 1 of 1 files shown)


packages/@momentum-design/docs/astro.config.mjs
```diff
@@ -7,6 +7,9 @@ import copy from 'rollup-plugin-copy';
 import mdx from '@astrojs/mdx';
 import path from 'path';
 
+const iconsDistFolder = path.dirname(import.meta.resolve('@momentum-design/icons/dist/manifest.json'));
```
cscheffauer, line 10: would be great to have a documentation here what import.meta.resolve does

cscheffauer (dismissed):

packages/@momentum-design/docs/astro.config.mjs
2020Deeya: Done.

cscheffauer (approved):

PR #798, opened 2024-08-30
feat: add unit tests for High contrast tokens

2020Deeya:
# Description

Added unit tests for High contrast tokens.

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-54


2020Deeya:
image

The PR will fail when executing the test cases because the description for the WindowTextColor token does not follow the expected format. For instance, 
Current Description: "Text main color"
Expected Description: Should start with the format SystemColorXXX, where XXX refers to the token name. It should be SystemColorWindowTextColor in this case.

packages/assets/tokens/src/core/value.test.ts
```diff
@@ -0,0 +1,36 @@
+const values = require('./value.json');
+const expectedColors = require('./mappingTable');
```
arjunkesava, line 2: I sense a problem with this approach.
- As the `value.json` file is generated by the `token.studio` figma plugin, if there is any update in the Tokens Figma file, then these specs get failed.
- Instead of validating the system color postfix, we can make a new regexp and expect the entire data in value.json file to match the regexp everytime.

arjunkesava (changes requested):

packages/assets/tokens/src/core/value.test.ts
cscheffauer: the point of this unit test is to make sure that the changes made by token studio are actually what we expect. Typos or unwanted changes for these SystemColors should be avoided, since Windows requires them to be accurate.
Thats why we write this test originally.

Any other changes in the tokens themselves should not be checked.

To your point 2, @arjunkesava : we only want the HighContrast Core Tokens to be validated.

arjunkesava: Ohk, if the requirement is only to validate HighContrast Core Tokens, then we are good.

arjunkesava (dismissed):

packages/assets/tokens/src/core/value.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// Validate the 'SystemColorXXX' format in descriptions
+const validateSystemColorPrefix = (description, tokenName) => {
+    const expectedPrefix = `SystemColor${tokenName}`;
+    return description.startsWith(expectedPrefix);
+};
+
+describe('High Contrast Colors', () => {
+  test('Black high contrast tokens should match expected values and descriptions', () => {
```
cscheffauer, line 11: nitpick: in other unit tests we have been using `it` to follow more the english wording, i.e. `it("should do ..."`.
maybe worth creating a ESLint rule for all packages to check this.

packages/assets/tokens/src/core/value.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    const blackColors = values.color.highcontrast.black;
+    const expectedBlackColors = expectedColors.black;
+    
+    Object.keys(expectedBlackColors).forEach(key => {
+      expect(blackColors[key].value).toBe(expectedBlackColors[key].value);
+      expect(blackColors[key].description).toBe(expectedBlackColors[key].description);
```
cscheffauer, line 17: this assertion is more specific than the assertion in line 20. We do not need to check that the rest of the description matches our expectation, but only the start with the SystemColor is important. This line can be deleted

packages/assets/tokens/src/core/value.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      // Validate that the description starts with 'SystemColor' followed by the token name
+      const tokenName = key;
```
cscheffauer, line 19: no need to create this constant, key can be used directly

packages/assets/tokens/src/core/value.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    });
+  });
+
+  test('White high contrast tokens should match expected values and descriptions', () => {
+    const whiteColors = values.color.highcontrast.white;
+    const expectedWhiteColors = expectedColors.white;
+    
+    Object.keys(expectedWhiteColors).forEach(key => {
+      expect(whiteColors[key].value).toBe(expectedWhiteColors[key].value);
+      expect(whiteColors[key].description).toBe(expectedWhiteColors[key].description);
```
cscheffauer, line 30: same comment as above

packages/assets/tokens/src/core/value.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      // Validate that the description starts with 'SystemColor' followed by the token name
+      const tokenName = key;
```
cscheffauer, line 32: same comment as above

cscheffauer (changes requested):

packages/assets/tokens/src/core/value.test.ts
2020Deeya: Resolved. 
For creating a ESLint rule for all packages to check this, let's work on separate PR.

packages/assets/tokens/src/core/value.test.ts
2020Deeya: Agreed. Checking Prefix is enough.

cscheffauer (approved):

PR #802, opened 2024-09-02
fix(components): components prepublish

cscheffauer:
# Description

- To avoid yarn workspaces linking still being present in the package.json, remove it on prepublish
- Cleanup of token builder package.json formatting


2020Deeya (approved):
LGTM

PR #806, opened 2024-09-04
feat: remove devDependencies and scripts before publishing components package to npm

2020Deeya:
# Description

- Remove devDependencies and scripts from package.json before publishing components package to npm.
- Execute the prepareForPublish.js script manually to verify that it modifies package.json as intended.
   - `node scripts/prepareForPublish.js`

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-249

### DIFF
--- a/packages/components/scripts/prepareForPublish.js
+++ b/packages/components/scripts/prepareForPublish.js
@@ -7,7 +7,6 @@ const main = async () => {
   const packageJson = require(packageJsonPath);
 
   const dependencies = packageJson.dependencies || {};
-  const devDependencies = packageJson.devDependencies || {};
   const peerDependencies = packageJson.peerDependencies || {};
 
   for (const dep in dependencies) {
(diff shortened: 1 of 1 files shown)


packages/components/scripts/prepareForPublish.js
```diff
@@ -28,6 +28,9 @@ const main = async () => {
     }
   }
 
+  delete packageJson.scripts;
+  delete packageJson.devDependencies;
```
arjunkesava, line 32: If we delete devDependencies, then we don't need to update them with "*" [over here](https://github.com/2020Deeya/momentum-design/blob/f9254176879f64dda21136c8f56ecd983ca02337/packages/components/scripts/prepareForPublish.js#L19-L23).
This can be safely removed.


```
  for (const dep in devDependencies) {
    if (devDependencies[dep].startsWith('workspace:')) {
        devDependencies[dep] = `*`;
    }
  }
  ```

2020Deeya: Absolutely agreed. We should remove this part of code. Thanks!

arjunkesava (approved):

PR #823, opened 2024-09-09
ci: show playwright report as artifact for github actions

2020Deeya:
# Description

In order to save the artifact of Playwright:

- Upgraded pull-request.yml to upload the playwright-report at the end of running the E2E tests.
- Added @estruyf/github-actions-reporter which reports test results from Playwright to GitHub summaries.
- This doesn't have any impact to the output

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-267


supminn (dismissed):
LGTM

.github/workflows/pull-request.yml
```diff
@@ -343,3 +343,11 @@ jobs:
         
       - name: Test - Postbuild
         run: yarn workspace ${{ matrix.package }} test:postbuild
+
+      - name: Upload Playwright Report
```
cscheffauer, line 347: i think it would be great to unzip the report first, then upload it, such that we can directly go to the html which is inside of the zip (with that we wouldn't need to download)

.github/workflows/pull-request.yml
```diff
@@ -343,3 +343,11 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        if: ${{ !cancelled() }}
```
cscheffauer, line 349: at the moment this step runs in all packages, i.e. also runs if tokens Test - Postbuild runs.
We only want this step to run for components package - we could extend this if statement with a matrix.package comparison to make sure to only run this upload if the matrix.package is the components one.

cscheffauer (changes requested):

.github/workflows/pull-request.yml
2020Deeya: As discussed, @estruyf/github-actions-reporter is added as a dependency in components, which reports test results from Playwright to GitHub summaries.

Reference PR for sanity testing: 
https://github.com/momentum-design/momentum-design/actions/runs/10808908015/job/29983068894

packages/components/package.json
```diff
@@ -50,6 +50,7 @@
     "generate:component": "yarn build:plop && plop --plopfile config/plop/dist/plopfile.mjs"
   },
   "dependencies": {
+    "@estruyf/github-actions-reporter": "^1.9.2",
```
cscheffauer, line 53: this has been added as a dependency - it should be in "devDependencies". When adding dependencies make sure to add `--dev` when using yarn to install dependency in a package - or move it in the right dependency group.

packages/components/config/playwright/playwright.config.ts
```diff
@@ -33,7 +33,10 @@ const config: PlaywrightTestConfig = {
   /* Opt out of parallel tests on CI. */
   workers: process.env.CI ? 1 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: 'html',
+  reporter: process.env.CI ? [
+    ['html'],
+    ['@estruyf/github-actions-reporter']
```
cscheffauer, line 38: there is a bit of config i would suggest adding for this actions reporter here, which is:
```
import type { GitHubActionOptions } from '@estruyf/github-actions-reporter';

...
['@estruyf/github-actions-reporter', <GitHubActionOptions>{
      title: 'Playwright E2E Test results',
      useDetails: true,
      showTags: true,
      showError: true,
      includeResults: ['fail', 'flaky']
    }]
```

We can have a chat about how i ended up with this config and why.

cscheffauer (changes requested):

packages/components/package.json
2020Deeya: Sure, have added it as a dev dependency.

packages/components/config/playwright/playwright.config.ts
2020Deeya: Done, have incorporated these config options.

packages/components/config/playwright/playwright.config.ts
```diff
@@ -9,6 +10,15 @@ import { port } from '../esbuild/configs/e2e';
 
 const url = `http://localhost:${port}`;
 
+// GitHub Actions reporter options
```
cscheffauer, line 13: nitpick: comment is a duplicate of the code, please remove

cscheffauer (dismissed):
small comment, other than that looking good

packages/components/config/playwright/playwright.config.ts
2020Deeya: Done.

cscheffauer (approved):

PR #824, opened 2024-09-10
chore: do not enforce yarn version in consuming projects

NotNestor:
# Description

"packageManager": "yarn@3.2.4" ensures that the specified version of Yarn is used for the current project.
"engines": { "yarn": ">=3.0.0" } specifies the minimum version of Yarn required for consuming projects.

momentum-design should not dictate the the yarn version of consumers

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-268

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -3,8 +3,7 @@
   "packageManager": "yarn@3.2.4",
   "engines": {
     "node": ">=20.0.0",
-    "npm": ">=8.0.0",
-    "yarn": ">=3.0.0"
+    "npm": ">=8.0.0"
   },
   "main": "./dist/module/index.js",
   "types": "./dist/types/index.d.ts",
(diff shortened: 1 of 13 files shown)


packages/components/package.json
```diff
@@ -3,8 +3,7 @@
   "packageManager": "yarn@3.2.4",
   "engines": {
     "node": ">=20.0.0",
-    "npm": ">=8.0.0",
-    "yarn": ">=3.0.0"
+    "npm": ">=8.0.0"
```
cscheffauer, line 6: as discussed, lets remove it just in the prepublish script, but keep it in the package.json.
We want to enforce this for contributors, locally developing.

We want to remove it when publishing

cscheffauer (changes requested):

packages/components/package.json
NotNestor: no it should be removed in engines packageManager already tells what version should be used when building the project. yarn should be removed from engines

cscheffauer (approved):

PR #830, opened 2024-09-12
ci: make Visual Regression Testing work again

2020Deeya:
# Description

In order to make the Visual regression testing work in pipeline:

- Adjust the threshold for now

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-271


packages/components/src/components/avatar/avatar.e2e-test.ts
```diff
@@ -28,11 +28,11 @@ test.fixme('mdc-avatar', async ({ componentsPage }) => {
    */
 
   // TODO: fix visual regression test on CI
```
cscheffauer, line 30: suggestion: remove TODO comment as well given it works for now on CI (applies for all e2e tests)

packages/components/config/playwright/setup/constants.ts
```diff
@@ -15,8 +15,8 @@ const ACCESSIBILITY = {
 };
 
 const VISUAL_REGRESSION = {
-  THRESHOLD: 0,
-  MAX_DIFF_PIXELS_RATIO: 0.03,
+  THRESHOLD: 0.1,
```
cscheffauer, line 18: maybe worth adding a comment here that we keep the threshold, max_diff_pixels_ratio at a higher value until we have pixel perfect visual regression testing enabled.

cscheffauer (changes requested):

packages/components/config/playwright/setup/constants.ts
2020Deeya: Sure, it'll keep other developers updated.

packages/components/src/components/avatar/avatar.e2e-test.ts
2020Deeya: Done.

cscheffauer (approved):

PR #834, opened 2024-09-12
fix(spec): add req packages to build for test setup

arjunkesava:
# Description

- Problem: The PR pipeline fails randomly due to a weird error. Ex: [failure action](https://github.com/momentum-design/momentum-design/actions/runs/10683196872/job/29610843997)
 image

- After research, we found that to run this below script, we need the necessary packages to be `build` before to run this.
- https://github.com/momentum-design/momentum-design/blob/0e45a7bf8f646476012297d8c22a0507ebfd6901/packages/components/package.json#L43
- As github workflows run parallelly, there is no guarantee that the build will execute just before this script gets started.
- This PR will fix this issue by adding a mandatory `build step` so that it will build all the needed packages to run before `playwright` runs the above step.
- Here is the successful output:
Screenshot 2024-09-12 at 9 24 48 PM

## Links

- [JIRA Ticket](https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-250)

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -347,7 +347,14 @@ jobs:
         run: yarn
         
       - name: Test - Postbuild
-        run: yarn workspace ${{ matrix.package }} test:postbuild
+        run: |
+          if [ ${{ matrix.package }} == '@momentum-design/components' ]; then
+            yarn workspaces foreach -pR --topological-dev --from @momentum-design/components run build
+            yarn workspace @momentum-design/components test:postbuild
+          else
+            yarn workspace ${{ matrix.package }} test:postbuild
+          fi
+
 
       - name: Upload Playwright Report
         uses: actions/upload-artifact@v4


packages/components/config/playwright/playwright.config.ts
```diff
@@ -97,7 +97,7 @@ const config: PlaywrightTestConfig = {
 
   /* Run your local dev server before starting the tests */
   webServer: {
-    command: 'yarn test:e2e:setup',
+    command: 'yarn build:test-deps && yarn test:e2e:setup',
```
arjunkesava, line 100: FYI: Instead of the script [build:test-deps](https://github.com/momentum-design/momentum-design/blob/8350b6066808af7b6ff9fe495ddac08c2051a112/packages/components/package.json#L26), we can `cd` to the project root directory and run `yarn build` and again come back over here.

packages/components/package.json
```diff
@@ -23,6 +23,7 @@
     "dev:generate-manifest": "custom-elements-manifest analyze --config config/custom-elements-manifest.config.js",
     "dev:storybook:setup": "concurrently --names 'copy:icons,copy:fonts' -c 'cyan.bold,blue.bold' 'yarn copy:icons storybook' 'yarn copy:fonts storybook'",
     "dev:storybook": "yarn dev:storybook:setup && storybook dev -p 6006 -c config/storybook",
+    "build:test-deps": "yarn workspace @momentum-design/telemetry build && yarn workspace @momentum-design/token-builder build && yarn workspace @momentum-design/tokens build && yarn workspace @momentum-design/common build && yarn workspace @momentum-design/builder build && yarn workspace @momentum-design/fonts build && yarn workspace @momentum-design/icons build",
```
arjunkesava, line 26: These are all the scripts which were needed to run "copy:tokens, copy:icons and copy:fonts scripts".

packages/components/config/playwright/playwright.config.ts
cscheffauer: this command runs locally as well - we do not want that.
Instead i would recommend to edit the pipeline (pull-request.yml) itself

packages/components/package.json
cscheffauer, line 26: this is a dependency in the component package on other dependencies - yarn workspaces can figure that dependency tree out - i would vote for the pull-request.yml file approach though, which is not having a script like u have but just run yarn build as part of the test:post-build directly again.

cscheffauer (changes requested):

cscheffauer (approved):

PR #842, opened 2024-09-16
feat: typography tokens (variables)

shannkha:
# Description
feat: typography tokens

- Created application typography tokens (body, heading, and headline) for global web library.
- Created typography tokens like font family, font size, font weight, which are used in the application tokens.

Outstanding work: 
- Export sizes, weights, heights in rem, instead of numbers (i.e. 16 => 1rem)
- Make use of the type tokens in Components.

## Links
https://www.figma.com/design/QNRCZ4UUw1SOVbCUeL8HbB/%F0%9F%92%8E-Core---Type%3A-Applications?m=auto&node-id=2-47&t=f586ETmU3msLAUCu-1


clairelikesemojis (commented):
@cscheffauer Still not know where all redlines are coming from, only assumption is we updated our token storage location to the plugin so maybe that reflected here?

cscheffauer (approved):

PR #843, opened 2024-09-17
feat(types): add typescript transformer flow

chrisadubois:
# Description

Add a typescript transformer for the brand-visuals, illustrations, animations, and icons folder in momentum-design.

This will create Typescript types in dist/types for:
- Icons
- Illustrations
- Brand Visuals
- Animations


packages/tools/builder/src/assets/builder/transformer/types-transformer.ts
```diff
@@ -0,0 +1,50 @@
...
+    super(format, destination, 'TYPES');
+  }
+
+  /**
+     * Generates the types data from the config
+     * @returns object with types data
+     */
+
+  private async generateTypesData(): Promise<{data: string}> {
+    // const glyphData = this.getGlyphData();
```
cscheffauer, line 22: suggestion: commented code should not be committed, seems to be a left over from a copy i guess

packages/assets/illustrations/config/momentum.json
```diff
@@ -40,6 +40,20 @@
               "fileName": "manifest.json"
             }
           }
+        },
```
cscheffauer, line 43: ask: i know this is not in the scope of this PR, but would u mind fixing a bug in the illustrations?
I do see the "manifest.json" adds itself in the manifest.json as a entry - that leads to a type of "manifest". Can u fix this maybe? it seems like the target of the Manifest is ".dist/**/*.*" which picks all files in dist and adds them to the manifest.

packages/tools/builder/src/assets/builder/transformer/types-transformer.ts
chrisadubois: oops

packages/tools/builder/src/assets/builder/transformer/types-transformer.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+class TypesTransformer extends Transformer {
+  constructor(format: Formats, destination: string) {
+    super(format, destination, 'TYPES');
+  }
+
+  /**
+     * Generates the types data from the config
+     * @returns object with types data
+     */
```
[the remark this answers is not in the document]
chrisadubois: sure

packages/assets/illustrations/config/momentum.json
chrisadubois: done

PR #846, opened 2024-09-18
docs: updated component readme

cscheffauer:
# Description

- Updated the README, Contributing, Scripts markdown files in Components for more information.
- This has only an affect on the docs, no change to the output


packages/components/CONTRIBUTING.md
```diff
@@ -2,4 +2,34 @@
...
+
+```sh
+yarn telemetry build
+yarn token-builder build
+yarn tokens build
+yarn common build
+yarn builder build
+yarn fonts build
+yarn icons build
+```
```
supminn, line 29: question: should we run each command separately? If all the builds need to be updated, could we have a single command to do that?

supminn (approved):

packages/components/CONTRIBUTING.md
cscheffauer: good point, yarn build is doing that - will adjust

supminn (approved):

PR #848, opened 2024-09-19
fix: host storybook in astro docs

2020Deeya:
# Description

In order to expose the storybook within our components package to consumers, 

- Include storybook code into the existing astro docs

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-283


packages/assets/fonts/src/css/fonts.css
```diff
@@ -1,4 +1,4 @@
 @font-face {
     font-family: "Momentum";
-    src: url("/fonts/Inter.var.woff2") format("woff2");
+    src: url("../fonts/Inter.var.woff2") format("woff2");
```
cscheffauer, line 3: the path here is broken in the dist of the fonts package - we should set it such that it works as it is in the fonts package and not change it for other packages.
image

packages/components/package.json
```diff
@@ -14,7 +14,8 @@
       "types": "./dist/index.d.ts"
     },
     "./module": "./dist/components",
-    "./browser": "./dist/browser"
+    "./browser": "./dist/browser",
+    "./storybook-static": "./dist/storybook-static"
```
cscheffauer, line 18: issue: this export should not be here - we should not include the storybook-static in the NPM package.

packages/components/package.json
```diff
@@ -30,7 +31,8 @@
     "build:browser:e2e": "node config/esbuild/esbuild-e2e.config.js",
     "build:browser": "node config/esbuild/esbuild.config.js",
     "build:module": "tsc --project \"./tsconfig.module.json\"",
-    "build": "yarn build:module && yarn build:browser",
+    "build-storybook": "storybook build -c config/storybook -o dist/storybook-static",
+    "build": "yarn build:module && yarn build:browser && yarn build-storybook",
```
cscheffauer, line 35: issue: we should not include the build-storybook in the normal build, since the normal build is about creating the dist folder for the npm package release. Just keep build-storybook separate and we can modify the github actions for documentation to run this.

packages/components/package.json
```diff
@@ -30,7 +31,8 @@
     "build:browser:e2e": "node config/esbuild/esbuild-e2e.config.js",
     "build:browser": "node config/esbuild/esbuild.config.js",
     "build:module": "tsc --project \"./tsconfig.module.json\"",
-    "build": "yarn build:module && yarn build:browser",
+    "build-storybook": "storybook build -c config/storybook -o dist/storybook-static",
```
cscheffauer, line 34: suggestion: for the sake of consistency we could do `build:storybook` here for the script name

packages/components/config/storybook/provider/iconProvider.js
```diff
@@ -2,7 +2,7 @@ import { html } from 'lit';
 
 export const withIconProvider = (story) =>
   html`<mdc-iconprovider
-        url="/icons/svg"
+        url="./icons/svg"
```
cscheffauer, line 5: issue: just tried the storybook in the docs locally and it seems like the icon component is broken.
Please check

did yarn docs build / yarn docs preview

cscheffauer (changes requested):

packages/components/package.json
2020Deeya: Done

packages/components/config/storybook/provider/iconProvider.js
2020Deeya: I did yarn docs build and then, yarn docs preview. It has worked for me. We can verify it offline.

packages/components/package.json
2020Deeya: We used this export approach to easily access the storybook-static folder through dynamic module resolution with import.meta.resolve. Since we don't want to include the storybook-static folder in the NPM package, I have now removed the 18th line that exports storybook-static.

packages/assets/fonts/src/css/fonts.css
2020Deeya: Resolved the relative paths of fonts.

packages/components/scripts/copyFonts.js
```diff
@@ -3,8 +3,8 @@ const fse = require('fs-extra');
 const path = require('path');
 const chalk = require('chalk');
 
-const fontsFolder = path.dirname(require.resolve('@momentum-design/fonts/dist/inter/Inter.var.woff2'));
 const root = process.cwd();
+const fontsFolder = path.join(root, '..', 'assets', 'fonts', 'dist');
```
cscheffauer, line 7: suggestion: in this case i would still vote for having the require.resolve

packages/documentation/astro.config.mjs
```diff
@@ -12,9 +12,14 @@ import path from 'path';
 // ensuring reliable path resolution for tasks like copying files, compared to using relative paths.
 const getAssetsFolder = (manifestEntryPoint) => path.dirname(fileURLToPath(import.meta.resolve(manifestEntryPoint)));
 
+// Using relative path approach for now as we don't keep storybook-static folder inside dist of components
+const getStorybookFolder = () => path.join(path.dirname(fileURLToPath(import.meta.url)), '..', 'components');
```
cscheffauer, line 16: suggestion: this does not need to be a function

cscheffauer (changes requested):
some small comments, otherwise looking good

packages/components/scripts/copyFonts.js
2020Deeya: Done.

packages/documentation/astro.config.mjs
2020Deeya: Agreed.

cscheffauer (approved):

PR #857, opened 2024-10-01
test(e2etest): support all browser and update themeprovider e2e test

Jason20020:
# Description

- Enable support for all browsers in Playwright (Chrome, Edge, Firefox, Safari)
- Make sure all e2e tests run in all browsers
- Make sure visual regression screenshots are suffixed with the browserName to have different screenshots per browser 
- Make sure the pipeline runs all tests in all browsers successfully
- Adjust the existing ThemeProvider E2E tests (fix and make sure they work in all browsers) 

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-336


packages/components/config/esbuild/configs/browser.js
```diff
@@ -4,7 +4,7 @@ const projectPath = process.cwd();
 const outPath = join('dist', 'browser', 'index.js');
 
 // todo: define browser targets here:
-const browsers = ['chrome114', 'firefox114'];
+const browsers = ['chrome114', 'firefox114', 'safari13', 'edge93', 'ios13'];
```
cscheffauer, line 7: we don't need to test ios13 - like we do not officially support mobile safari. I think we can add that later if we need it, but i would keep it out for now given we are not focusing to it.

packages/components/config/esbuild/configs/browser.js
```diff
@@ -4,7 +4,7 @@ const projectPath = process.cwd();
 const outPath = join('dist', 'browser', 'index.js');
 
 // todo: define browser targets here:
```
cscheffauer, line 6: this comment can be removed now i think

packages/components/config/playwright/playwright.config.ts
```diff
@@ -43,7 +43,9 @@ const config: PlaywrightTestConfig = {
   /* Opt out of parallel tests on CI. */
   workers: process.env.CI ? 1 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: process.env.CI ? [['html'], ['@estruyf/github-actions-reporter', githubActionsReporterOptions]] : 'html',
+  reporter: process.env.CI
+    ? [['html'], ['@estruyf/github-actions-reporter', githubActionsReporterOptions]]
+    : [['list'], ['html']],
```
cscheffauer, line 48: why adding list here? i think we don't need that reporter

packages/components/config/playwright/playwright.config.ts
```diff
@@ -66,41 +68,41 @@ const config: PlaywrightTestConfig = {
...
-    //   },
-    // },
-    // {
-    //   name: 'Mobile Safari',
-    //   use: {
-    //     ...devices['iPhone 12'],
-    //   },
-    // },
+    {
+      name: 'Mobile Chrome',
```
cscheffauer, line 87: we are not supporting mobile for components yet, thus we can just keep them out of this PR

packages/components/config/playwright/playwright.config.ts
```diff
@@ -111,7 +113,7 @@ const config: PlaywrightTestConfig = {
     command: 'yarn test:e2e:setup',
     url,
     timeout: 240 * 1000,
-    reuseExistingServer: true,
+    reuseExistingServer: !process.env.CI,
```
cscheffauer, line 116: why was this change necessary? this was working before

packages/components/package.json
```diff
@@ -46,7 +46,8 @@
     "test:postbuild": "yarn test:e2e",
     "test:prebuild": "echo \"script 'test:prebuild' has not been implemented\"",
     "test:e2e:report": "yarn playwright show-report",
-    "test:e2e:setup": "concurrently --names 'copy:tokens,copy:icons,copy:fonts' -c 'magenta.bold,cyan.bold,blue.bold' 'yarn copy:tokens' 'yarn copy:icons playwright' 'yarn copy:fonts playwright' && yarn build:browser:e2e",
+    "test:e2e:browser": "playwright install-deps && playwright install",
```
cscheffauer, line 49: the name of this script is not descriptive enough - i would recommend something like:
`test:e2e:install` or something like that.

packages/components/package.json
```diff
@@ -46,7 +46,8 @@
...
+    "test:e2e:setup": "yarn test:e2e:browser && concurrently --names 'copy:tokens,copy:icons,copy:fonts' -c 'magenta.bold,cyan.bold,blue.bold' 'yarn copy:tokens' 'yarn copy:icons playwright' 'yarn copy:fonts playwright' && yarn build:browser:e2e",
```
cscheffauer, line 50: test:e2e:setup is used locally as well - we should not install the browsers all the time when running it locally - i would recommend to keep the "test:e2e:setup" as it is for now and if someone needs to install the browsers, they can just fire "test:e2e:install" independently.

For CI: we would want to run this script `test:e2e:install` directly in the github actions workflow on install.

packages/components/src/components/themeprovider/themeprovider.e2e-test.ts
```diff
@@ -9,7 +9,6 @@ test.beforeEach(async ({ componentsPage, theme }) => {
   await componentsPage.mount({
     html: `
     <mdc-themeprovider class="themeWrapper" id="local" theme="${themeClass}">
-      <p>Current theme: ${themeClass}</p>
```
cscheffauer, line 12: why was this removed?

cscheffauer (changes requested):

packages/components/config/playwright/playwright.config.ts
Jason20020: I will remove this

packages/components/config/playwright/playwright.config.ts
```diff
@@ -66,19 +66,19 @@ const config: PlaywrightTestConfig = {
         channel: 'chrome',
       },
     },
-    // {
-    //   name: 'firefox',
-    //   use: {
-    //     ...devices['Desktop Firefox'],
-    //   },
-    // },
+    {
```
cscheffauer, line 69: forgot mentioning: can u please rename all projects here to have a lowercase name i.e. "chrome", "firefox", "webkit", "edge"

with that the snapshotPath directory will always be lowercase and will not have any spaces.

cscheffauer (approved):
one comment which needs to be addressed, otherwise looking good

PR #864, opened 2024-10-02
test(e2e): add playwright install to support browser install on CI

Jason20020:
# Description
- Add test:e2e:install to package.json to support browsers install on workflow CI.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -57,6 +57,7 @@
     "test": "yarn test:prebuild && yarn test:postbuild",
     "test:postbuild": "yarn test:e2e",
     "test:prebuild": "echo \"script 'test:prebuild' has not been implemented\"",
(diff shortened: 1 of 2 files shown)


packages/components/package.json
```diff
@@ -57,6 +57,7 @@
     "test": "yarn test:prebuild && yarn test:postbuild",
     "test:postbuild": "yarn test:e2e",
     "test:prebuild": "echo \"script 'test:prebuild' has not been implemented\"",
+    "test:e2e:install": "playwright install-deps && playwright install",
```
cscheffauer, line 60: the pipeline doesn't pick this up yet - you would need to add this to the initialize step in the pull-request.yml workflow

cscheffauer (changes requested):

cscheffauer (approved):

PR #868, opened 2024-10-03
chore: Icon provider component

NotNestor:
# Description

* Update "scale" to "size" across project
* Add size to icon provider
* Add allow list of length units
* Update yarn lock file
* Fix various warnings

## Links

N/A


packages/components/config/playwright/setup/Components.page.ts
```diff
@@ -84,7 +84,7 @@ class ComponentsPage {
    * - Await till page has been loaded
    */
   async navigate(url?: string) {
-    await this.page.goto(url || '');
+    await this.page.goto(url ?? '');
     await expect(this.page).toHaveTitle(componentsDevPageTitle);
```
NotNestor, line 88: sonarlint prefers nullish coalescing over logical or

packages/components/config/playwright/setup/utils/accessibility.ts
```diff
@@ -26,7 +26,7 @@ class Accessibility {
    * File will be deleted after being attached, since after attaching to test report,
    * the file is not needed anymore locally
    *
-   * @param fileName name of the file to attach
+   * @param fileName - name of the file to attach
    */
```
NotNestor, line 30: jsdocs expects - after param

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -28,14 +28,14 @@ class Avatar extends Component {
    * Scale of the avatar
    */
   @property({ type: Number })
-  scale?: number = DEFAULTS.SCALE;
+  size?: number = DEFAULTS.SIZE;
```
NotNestor, line 32: update scale to size across components

packages/components/src/components/badge/badge.component.ts
```diff
@@ -1,10 +1,12 @@
 import { html } from 'lit';
 import { property } from 'lit/decorators.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
 import { StyleInfo, styleMap } from 'lit/directives/style-map.js';
 import { Component } from '../../models';
 import { DEFAULTS, WARNING_ICON_NAME } from './badge.constants';
 import styles from './badge.styles';
 import type { BadgeType } from './badge.types';
+import '../icon';
```
cscheffauer, line 9: i think we need to look at this closer - this line will register the icon component, but if someone just imports the Badge as a Class, we do not want the Icon component to be registered i think.
A possibly solution for this would be to move this import to the Badge index file.
This feels like we should come up with a convention for it.

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -6,6 +6,7 @@ const styles = [
   css`
     :host {
       --mdc-badge-icon-background-color: var(--mds-color-theme-background-accent-normal);
+      --mdc-badge-icon-color: var(--mds-color-theme-common-text-primary-normal);
```
cscheffauer, line 9: i will not review these lines here given we might change this later and its not really in scope of the iconprovider

packages/components/src/components/iconprovider/iconprovider.constants.ts
```diff
@@ -3,10 +3,17 @@ import utils from '../../utils/tag-name';
...
+const ALLOWED_LENGTH_UNITS = ['em', 'rem', 'px'];
 
-const DEFAULTS = {
+const DEFAULTS:Record<string, string> = {
   FILE_EXTENSION: 'svg',
   LENGTH_UNIT: 'em',
 };
 
-export { TAG_NAME, DEFAULTS, ALLOWED_FILE_EXTENSIONS };
+const LENGTH_UNIT_DEFAULT_SIZE:Record<string, number> = {
```
cscheffauer, line 13: nitpick: i think this should be part of the DEFAULTS object for keeping all DEFAULTS together

packages/components/src/components/iconprovider/iconprovider.component.ts
```diff
@@ -43,15 +47,28 @@ class IconProvider extends Provider<IconProviderContext> {
    * Length unit used for sizing of icons, default: 'em'
    */
   @property({ type: String, attribute: 'length-unit', reflect: true })
-  lengthUnit?: string = DEFAULTS.LENGTH_UNIT;
+  lengthUnit: string = DEFAULTS.LENGTH_UNIT;
+
+  /**
+   * The default size of the icon.
+   * This property can be set via the 'size' attribute.
```
cscheffauer, line 54: this line is just repeating what is clear, given the TSDoc here is for the size attribute

cscheffauer (changes requested):

packages/components/src/components/badge/badge.styles.ts
NotNestor: no but when testing i want the icon colour to change with the background colour correctly for demoing purposes

packages/components/src/components/iconprovider/iconprovider.component.ts
NotNestor: 👍 will update

packages/components/src/components/icon/icon.component.ts
```diff
@@ -117,6 +120,10 @@ class Icon extends Component {
     }
   }
 
+  private get computedIconSize() {
+    return this.size ?? this.sizeFromContext ?? DEFAULTS.SIZE;
```
supminn, line 124: question: do we need to add `DEFAULT.SIZE` here? Line52 would set the size to default value right? 
I'm not sure what value would this resolve to, since `this.size` would either have a value or take the `DEFAULT.SIZE`. This would never apply `this.sizeFromContext`. Is my understanding correct or does it work in a different way?

NotNestor: Updated to remove default value from size property. Now if not set it will fall back

cscheffauer (approved):

PR #874, opened 2024-10-07
test(e2e): added e2e tests for themeprovider

cscheffauer:
# Description

- Added themeprovider E2E tests, including more test cases, snapshots, etc.
- Enabled us to include "special components" for our e2e test build (modifying entrypoints in esbuild)
- Fixed playwright fixture to properly apply theme to page.
- Modified tsconfig to correctly apply typescript


Jason20020 (approved):

packages/components/config/playwright/setup/constants.ts
```diff
@@ -29,7 +29,7 @@ const DEFAULT_ACCESSIBILITY_SCAN_OPTIONS = {
 // we have pixel perfect visual regression testing enabled.
 const VISUAL_REGRESSION = {
   THRESHOLD: 0.1,
-  MAX_DIFF_PIXELS_RATIO: 0.08,
+  MAX_DIFF_PIXELS_RATIO: 0.15,
```
supminn, line 32: question: why did we change this value? Do we have a particular method to calculate this value?

cscheffauer: this was necessary cause screenshots were not matching on CI. There is a difference between running Browsers on different OS (anti-aliasing for example), which will mean that screenshots will be slightly different between were u take them. This is a tech debt - at some point we would want to switch to a docker image, which allows us to take the same screenshots locally and CI, which will help to make our regression tests pixel perfect.
At the moment its really just a "doesn't work on CI (because of browser rendering)" -> bump the ratio to a higher level to make it pass.

supminn (dismissed):

packages/components/config/playwright/setup/constants.ts
supminn: Thanks for the clarification. Rest of the changes looks good, I'm approving it.

supminn (approved):

PR #878, opened 2024-10-08
fix: provider stylesheet inheritance and display format

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to update the `display` of the `<mdc-themeprovider />` component to be `contents`, [which should be used for non-block-impacting components](https://developer.mozilla.org/en-US/docs/Web/CSS/display#contents).

Additionally, Stylesheet inheritance was being misrepresented from the `Provider` model into the `ThemeProvider` class definition.

Storybooks updated accordingly

# Links

* [Developer documentation on display contents]((https://developer.mozilla.org/en-US/docs/Web/CSS/display#contents).)


packages/components/src/components/themeprovider/themeprovider.component.ts
```diff
@@ -95,7 +96,7 @@ class ThemeProvider extends Provider<ThemeProviderContext> {
     }
   }
 
-  public static override styles = styles;
+  public static override styles: Array<CSSResult> = [...Provider.styles, styles];
```
chrisadubois, line 99: good catch

packages/components/src/models/provider/provider.component.ts
```diff
@@ -53,7 +53,7 @@ abstract class Provider<C> extends Component {
   /**
    * Styles associated with this Provider Component.
    */
-  public static override styles: CSSResult | Array<CSSResult> = styles;
+  public static override styles: Array<CSSResult> = [styles];
```
chrisadubois, line 56: noice

packages/components/src/models/provider/provider.styles.ts
```diff
@@ -2,12 +2,7 @@ import { css } from 'lit';
 
 const styles = css`
   :host {
-    border: 0;
-    display: block;
-    height: 100%;
-    margin: 0;
-    padding: 0;
-    width: 100%;
+    display: contents;
```
chrisadubois, line 5: 👍

chrisadubois (approved):
lgtm, but would want @cscheffauer approval.

tests are also failing.

packages/components/src/components/iconprovider/iconprovider.styles.ts
```diff
@@ -0,0 +1,11 @@
+import { css } from 'lit';
+
+const styles = css`
```
cscheffauer, line 3: @Jason20020 this is against what Tim did with his change - why exactly did you need to set this? Where the screenshots off? or what is the impact if we don't set this style here?

cscheffauer (changes requested):

cscheffauer (approved):

PR #879, opened 2024-10-09
chore(component): update existing icon component

arjunkesava:
# Description

- Update icons component to match with all the criteria as described on Icon tech [confluence](https://confluence-eng-gpk2.cisco.com/conf/pages/viewpage.action?pageId=562064947) page.
- Refactor dynamic SVG import code from promise pattern to async await pattern.
- Provide the option for consumers to add styles and classes.

| Before |
|---|
| image |

| After |
|---|
| image| 

#### Screen recording which explains this commit: 05e901e7cbb9f7b777cd875fece603a97f716d63
 
[story-background-screencast.webm](https://github.com/user-attachments/assets/7d6df6fa-197e-4bef-8539-bcfb10eb2e92)

The existing "Text" component is also not visible due to the white background which is set as a default.
This PR will add the "dark" background of gray color, due to which the text component can be visible.

## Links

JIRA ticket: [MOMENTUM-357](https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-357)


supminn (dismissed):

packages/components/config/storybook/preview.js
```diff
@@ -23,7 +23,7 @@ const preview = {
     },
     actions: { argTypesRegex: '^on[A-Z].*' },
     backgrounds: {
-      disable: true,
+      default: 'dark',
```
cscheffauer, line 26: this is hard coding it to dark - we would want to set the background based on the theme. Use css overrides & see storybook for how to do this.

packages/components/src/components/icon/icon.component.ts
```diff
@@ -24,7 +24,9 @@ import { DEFAULTS } from './icon.constants';
  * if `size = 1` and `length-unit = 'em'`, the size of the icon will be
  * `width: 1em; height: 1em`.
  *
- * For accessibility the `role` and `aria-label` of the icon can be set.
+ * For accessibility:
+ * - the `role` will always be 'img'.
```
cscheffauer, line 28: issue: the role should not always be `img` - we discussed that in the breakdown. 
`aria-label` determines if a icon gets decorative or informative - only if the aria-label is set, we will set the role img.
This text here is not aligned with what we discussed.

packages/components/src/components/icon/icon.component.ts
```diff
@@ -103,14 +100,18 @@ class Icon extends Component {
   }
 
   private setRoleOnIcon() {
```
cscheffauer, line 102: issue: this function should also be fired when the ariaLabel changes. For example if the icon is loaded and the aria-label changes afterwards, we want the role="img" to be set (or unset).

packages/components/src/components/icon/icon.component.ts
```diff
@@ -134,10 +135,6 @@ class Icon extends Component {
       });
```
cscheffauer, line 135: the console.error in line 134 has a warning - please put a eslint-ignore for this line to actively allow the console error here.

packages/components/src/components/icon/icon.component.ts
```diff
@@ -24,7 +24,9 @@ import { DEFAULTS } from './icon.constants';
  * if `size = 1` and `length-unit = 'em'`, the size of the icon will be
  * `width: 1em; height: 1em`.
  *
- * For accessibility the `role` and `aria-label` of the icon can be set.
+ * For accessibility:
+ * - the `role` will always be 'img'.
+ * - `aria-label` of the icon can be set.
  *
  * @tag mdc-icon
  * @tagname mdc-icon
```
NotNestor, line 32: can we remove the @tag to fix the warnings. we should only use @tagname

packages/components/src/components/icon/icon.stories.ts
```diff
@@ -24,15 +39,18 @@ export const Primary: StoryObj = {
   args: {
     name: 'accessibility-regular',
     size: 1,
+    class: 'custom-classname',
+    style: 'display: block;',
```
cscheffauer, line 43: suggestion: to not confuse consumers, i would vote for a styling override which is less "impactful" and more obvious that its done through storybook, for example margin or so

packages/components/src/components/icon/icon.utils.ts
```diff
@@ -1,13 +1,14 @@
...
-const dynamicSVGImport = async (url: string, name: string, fileExtension: string): Promise<Element> =>
-  fetch(`${url}/${name}.${fileExtension}`)
-    .then((response) => {
-      if (!response.ok) {
-        throw new Error('There was a problem while fetching the icon!');
-      }
-      return response.text();
-    })
-    .then((iconResponse) => new DOMParser().parseFromString(iconResponse, 'text/html').body.children[0]);
+const dynamicSVGImport = async (url: string, name: string, fileExtension: string): Promise<Element> => {
```
cscheffauer, line 3: nitpick: this function could benefit from a TSDoc

packages/components/src/components/icon/icon.stories.ts
```diff
@@ -1,10 +1,22 @@
 import type { Meta, StoryObj, Args } from '@storybook/web-components';
 import '.';
 import { html } from 'lit';
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
 
-const render = (args: Args) => html` <mdc-icon name="${args.name}" size="${args.size}"></mdc-icon> `;
+const render = (args: Args) => html`
```
cscheffauer, line 6: Nothing in storybook talks about decorative vs informative icons. We want our consumers to be able to understand when to use which.
I would recommend:
- Change story names to match Decorative / Informative
- Create additional documentation for Accessibility (a separate docs page in this components folder) to document Accessibility more thoroughly.

cscheffauer (changes requested):
first initial review, did not test locally yet - and also something additional:
All what we defined about css properties, styling etc is missing in this PR - please add.

packages/components/src/components/icon/icon.component.ts
arjunkesava: Done.

packages/components/src/components/icon/icon.stories.ts
```diff
@@ -15,24 +29,32 @@ const meta: Meta = {
...
   args: {
-    name: 'accessibility-regular',
-    size: 1,
-    role: 'graphics-document',
     'aria-label': 'This is the accessibility icon',
+    name: 'accessibility-regular',
+    size: 2,
+    class: 'custom-classname',
+    style: 'margin: 0px;',
+    role: 'img',
```
cscheffauer, line 58: issue: passing the role in here provides the user with wrong information. Role is not a attribute on the component and should also not be set by the user. Please remove this from here.

packages/components/src/components/icon/icon.component.ts
```diff
@@ -24,10 +26,23 @@ import { DEFAULTS } from './icon.constants';
  * if `size = 1` and `length-unit = 'em'`, the size of the icon will be
  * `width: 1em; height: 1em`.
  *
- * For accessibility the `role` and `aria-label` of the icon can be set.
+ * In terms of `accessibility`, there are two types of icons, decorative and informative.
```
cscheffauer, line 29: suggestion: to make this more clear i would vote for explicitly saying "if `aria-label` is passed in, the Icon will be a informative icon, otherwise its a decorative icon" or something like that. This way a consumer will understand it. With the description you have its not clear for a consumer that the aria-label actually determines if its a decorative icon or a informative icon.

cscheffauer (changes requested):

packages/components/src/components/icon/icon.component.ts
supminn, line 1: Overall, the documentation covers everything that is required. I have rephrased a few instances for better understanding. I hope this helps.

cscheffauer (approved):

PR #881, opened 2024-10-09
fix(themeprovider): make use of typography tokens

cscheffauer:
# Description

- Make use of the new Typography tokens in Themeprovider component (fontFamily)
- Updated storybook to include new typography tokens
- Updated Playwright copy script to include new typography tokens
- Updated Consumption.mdx 
- Removed core theme tokens from copy scripts, since only Theme tokens should be used for colors


supminn (approved):
Thanks for the quick addition.

Jason20020 (approved):

PR #887, opened 2024-10-11
feat: New transparency graphic

morgan1bathe:
- transparency-graphic: new asset

No visual changes to the following:
- cisco-ai-assistant-bw-gradient.svg
- cisco-ai-assistant-color-gradient.svg
- google-meet-color.svg
- ms-teams-color.svg


cscheffauer (changes requested):
@morgan1bathe few questions: instead of a transparency graphic, wouldn't it be better to facilitate this through a token? a transparency graphic might be good for non css consumers, but web consumers can also just consume a token and apply the wanted transparency effect through that no?
My concern in particular is: This is another svg which needs to be downloaded, whereas a token is a simple line of CSS, which is more performant for consumers.
We can discuss this internally.

PR #890, opened 2024-10-14
fix: update iconProvider storybook

2020Deeya:
# Description

Update storybook for different icon provider configurations

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-349


packages/components/src/components/iconprovider/iconprovider.stories.utils.ts
```diff
@@ -14,6 +14,14 @@ class SubComponentIconProvider extends Component {
       <p>File Extension: ${this.iconProviderContext.value?.fileExtension}</p>
       <p>Length Unit: ${this.iconProviderContext.value?.lengthUnit}</p>
       <p>Default Size: ${this.iconProviderContext.value?.size}</p>
+      <mdc-iconprovider
```
cscheffauer, line 17: question: is there a reason to have another IconProvider in the Iconprovider here?
Like the "SubComponentIconProvider" component here will be rendered already in the actual IconProvider from the story.

cscheffauer (approved):
one question, otherwise looking good

packages/components/src/components/iconprovider/iconprovider.stories.utils.ts
2020Deeya: You're right. It's something redundant to have another IconProvider. We could remove it and keep only the icon component.

PR #891, opened 2024-10-14
feat: text component updated

supminn:
# Description

The text component has been updated based on the [Tech Breakdown](https://confluence-eng-gpk2.cisco.com/conf/pages/viewpage.action?pageId=563880910)

## Links

[MOMENTUM-365](https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-365)


packages/components/src/components/text/text.component.ts
```diff
@@ -2,46 +2,96 @@ import { html } from 'lit';
...
+      case 'h4': return html`<h4><slot></slot></h4>`;
+      case 'h5': return html`<h5><slot></slot></h5>`;
+      case 'h6': return html`<h6><slot></slot></h6>`;
+      case 'p': return html`<p><slot></slot></p>`;
+      case 'small': return html`<small><slot></slot></small>`;
+      case 'div': return html`<div><slot></slot></div>`;
+      case 'span': return html`<span><slot></slot></span>`;
+      case 'section': return html`<section><slot></slot></section>`;
+      default: return html`<p><slot></slot></p>`;
+    }
```
supminn, line 94: I wanted to create this section of the code as follows:
```html
 return html`<${this.tagname}><slot></slot></${this.tagname}>`
 ```
This is giving an error
Bindings cannot be used in place of tag names.
Template contained invalid HTML syntax, error was: invalid-first-character-of-tag-name

Alternatively, I worked on creating a utils function 
```javascript
import { html } from 'lit';
import { VALUES } from './text.constants';
import { ValidTextTags } from './text.types';

export function getTagNameTemplate(tagname: ValidTextTags | undefined): ReturnType<typeof html> {
  if (tagname && VALUES.TAGNAME.includes(tagname)) {
    // eslint-disable-next-line lit/binding-positions, lit/no-invalid-html
    return html`<${tagname}><slot></slot></${tagname}>`;
  }
  return html`<p><slot></slot></p>`;
}
```
But, this isn't helpful too. Let me know how I could improve this section of the code.

packages/components/src/components/text/fonts.styles.ts
supminn, line 1: I've noticed that when we change the `tagname`, the default styles from user agent stylesheets are overriding our defined styles based on `type`.

packages/components/src/components/text/text.component.ts
```diff
@@ -2,46 +2,96 @@ import { html } from 'lit';
...
+  public override render() {
+    if (!this.type) {
+      console.error('Type attribute is required for mdc-text component');
+      return null;
     }
-  }
 
-  public override render() {
-    return html`<slot></slot>`;
+    switch (this.tagname) {
```
cscheffauer, line 81: suggestion: this is a big switch case statement, include all of the html details. I would vote for replacing the tagname directly, like here: https://github.com/shoelace-style/shoelace/blob/next/src/components/button/button.component.ts#L264

packages/components/src/components/text/text.constants.ts
```diff
@@ -1,27 +1,57 @@
...
+    'p',
+    'small',
+    'span',
+    'div',
+    'section',
   ],
 };
 
+const DEFAULTS = {
+  TAGNAME: VALUES.TAGNAME[6],
```
cscheffauer, line 54: issue: this has the potential to be broken, in particular if a new TagName gets added to the array.
Either use a object to keep track of the tagnames or hard code the value here - we shouldn't rely on array indexes given the potential to introduce a bug.

cscheffauer (changes requested):
first initial review, haven't tested locally yet

packages/components/src/components/text/text.component.ts
supminn: We decided not to implement this change as per our discussion. We will continue to use switch case with hardcoded values to avoid using static-html from lit.

cscheffauer (dismissed):

packages/components/src/components/text/text.component.ts
```diff
@@ -2,46 +2,96 @@ import { html } from 'lit';
...
+   * Note that the styling is determined by the `type` attribute.
+   */
+  @property({ attribute: 'tagname', reflect: true, type: String })
+  public tagname?: ValidTextTags = DEFAULTS.TAGNAME;
 
-    if (changedProperties.has('type')) {
-      this.handleTypeChanged();
+  public override render() {
+    if (!this.type) {
+      console.error('Type attribute is required for mdc-text component');
```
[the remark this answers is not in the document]
supminn: Now that we are using a default value for type, I am getting this part of the code removed.

packages/components/src/components/text/text.component.ts
```diff
@@ -2,46 +2,97 @@ import { CSSResult, html } from 'lit';
...
+   * - 'headline-small-regular'
    */
   @property({ attribute: 'type', reflect: true, type: String })
-  public type?: FontType;
-
-  protected handleTypeChanged(): void {
-    this.role = getRole(this.type);
-    this.ariaLevel = getAriaLevel(this.type);
-  }
+  public type: FontType | null = null;
```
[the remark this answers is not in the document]
supminn: Sure, we can do this!
Since we want `type` to be a required attribute, I had set it to `null` and conditionally do not render the text component if `type` was not passed.

To keep it in-sync with how momentum-react does, I shall get the default value added. Thanks.

arjunkesava (approved):

PR #896, opened 2024-10-15
fix(components): base styling to inherit box sizing

cscheffauer:
# Description

There is an issue in our components where the box sizing isn't inherited properly through when using Higher Order Components like the ThemeProvider. This PR fixes this.
- Adding styling to base Component class
- Using the subclass styling in all components (except IconProvider since IconProvider doesn't have a separate component style)
- Tested locally and with E2E tests, all fine.

## Links

More information about it:
https://dev.to/konnorrogers/revisiting-box-sizing-best-practices-3del
In general, the shadow dom is not inheriting the box-sizing as other native elements do, thats why we need to set a standard box-sizing style on each component, which in our case achieved similar to shoelace:
https://github.com/shoelace-style/shoelace/blob/next/src/styles/component.styles.ts


supminn (approved):
LGTM

PR #897, opened 2024-10-16
feat: util function that converts px to rem for tokens

supminn:
# Description

- Created a local transform in token-builder package that converts line-heights and font-sizes from 'px' to 'rem'.

## Links

[MOMENTUM-364](https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-364)

### DIFF
--- a/packages/tools/token-builder/src/transforms/pxToRem/pxToRem.ts
+++ b/packages/tools/token-builder/src/transforms/pxToRem/pxToRem.ts
@@ -1,0 +1,67 @@
+import type { Named, Transform as SDTransform } from 'style-dictionary';
+
+import type { Config, Matcher, Token, Transformer, Type } from './types';
+
+import CONSTANTS from './constants';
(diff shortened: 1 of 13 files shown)


packages/tools/token-builder/src/transforms/pxToRem/pxToRem.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  public get type(): Type {
+    return 'value';
+  }
+
+  public get sdConfig(): Named<SDTransform> {
+    return {
+      name: this.name,
+      type: 'value',
```
cscheffauer, line 55: nitpick: i think here we want to use this.type as well, otherwise the getter in line 48 doesn't make sense.

cscheffauer (changes requested):

cscheffauer (approved):

PR #899, opened 2024-10-16
style: add shared styles for focus ring

2020Deeya:
# Description

Add shared CSS styles for the focus ring

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-373


packages/components/src/utils/styles/index.ts
```diff
@@ -10,4 +10,49 @@ const hostFitContentStyles = css`
...
+    --mdc-focus-ring-middle-width: 0.125rem;
+    --mdc-focus-ring-outer-width: 0.0625rem;
+
+    --mdc-focus-ring-middle-offset: var(--mdc-focus-ring-inner-width);
+    --mdc-focus-ring-outer-offset: calc(var(--mdc-focus-ring-inner-width) + var(--mdc-focus-ring-middle-width));
+
+    --mdc-focus-ring-z-index: 1; 
+  }
+
+  :host([focusRingVisible]) {
```
arjunkesava, line 29: Q: can you help me explain how does this work?

The variable `focusRingVisible` is not imported from anywhere and neither declared anywhere aswell.

arjunkesava (dismissed):
Please attach a gif / movie / screenshot explaining the hover (i.e, focus) state of this new css.

cscheffauer (dismissed):
code looks good to me - gonna test it in detail once we have a storybook

packages/components/src/utils/styles/index.ts
2020Deeya: Please refer this PR related to Mixin. Styles will be applied to web-component when this attribute will be set to true by Mixin.
https://github.com/momentum-design/momentum-design/pull/901/

supminn (approved):
Good to merge if the updated color tokens are available.

packages/components/src/utils/styles/index.ts
```diff
@@ -10,4 +10,49 @@ const hostFitContentStyles = css`
...
+    --mdc-focus-ring-middle-offset: var(--mdc-focus-ring-inner-width);
+    --mdc-focus-ring-outer-offset: calc(var(--mdc-focus-ring-inner-width) + var(--mdc-focus-ring-middle-width));
+
+    --mdc-focus-ring-z-index: 1; 
+  }
+
+  :host(.focus-ring-visible) {
+    position: relative;
+    outline: var(--mdc-focus-ring-inner-width) solid var(--mdc-focus-ring-inner-color);
+    z-index: var(--mdc-focus-ring-z-index)
```
cscheffauer, line 32: sorry for the confusion, i think the z-index here doesn't make sense - cause the host is our actual component itself (cause we use a mixin) - so in this case setting the z-index makes the whole component to have the z-index.
I think we actually don't need it - sorry for changing my point on this.

packages/components/src/utils/styles/index.ts
```diff
@@ -10,4 +10,49 @@ const hostFitContentStyles = css`
...
+    --mdc-focus-ring-middle-width: 0.125rem;
+    --mdc-focus-ring-outer-width: 0.0625rem;
+
+    --mdc-focus-ring-middle-offset: var(--mdc-focus-ring-inner-width);
+    --mdc-focus-ring-outer-offset: calc(var(--mdc-focus-ring-inner-width) + var(--mdc-focus-ring-middle-width));
+
+    --mdc-focus-ring-z-index: 1; 
+  }
+
+  :host(.focus-ring-visible) {
```
cscheffauer, line 29: i have tried this in a codesandbox - everything looks fine, except for 2 particular cases:
- Try to set a "border" on the button or element you apply the focusRing to - in this case the positioning doesn't work. We have to consider this and work it out.
- I tried it on Firefox and the outline positioning doesn't work.
I have played around with it to see how to make it work better, and this is the solution i came up with:
```
content: "";
  position: absolute;
  top: 0;
  left: 0;
  right: 0;
  bottom: 0;
  outline: var(--mdc-focus-ring-middle-width) solid var(--mdc-focus-ring-middle-color);
  border-radius: inherit;
  outline-offset: var(--mdc-focus-ring-middle-offset);
```
In particular: set the top, left, right, bottom to 0 and use "outline-offset" instead

cscheffauer (changes requested):

packages/components/src/utils/styles/index.ts
cscheffauer, line 29: instead of a class, i think we can simply do this:
https://github.com/shoelace-style/shoelace/blob/next/src/components/button/button.styles.ts#L42
`:focus-visible` selector.
With that we wouldn't need the focus-ring-visible class logic, the protected property and also the consuming component can control it (by setting the "focus" programmatically on the component).

cscheffauer (changes requested):

packages/components/src/utils/styles/index.ts
2020Deeya: Agreed. 
Reverted it back.

packages/components/src/utils/styles/index.ts
2020Deeya: Yes, 
Also, Instead of set the top, left, right, bottom to 0, we could make use of inset property.

2020Deeya: Agreed. Fixed it.

cscheffauer: issue: the first bullet point here has not been addressed, still an issue when setting the border on a element.

2020Deeya: There’s a limitation in our current logic for setting the border on an element. As an alternative, we will have to fall back on the box-shadow approach for now.

packages/components/src/utils/styles/index.ts
```diff
@@ -10,4 +10,27 @@ const hostFitContentStyles = css`
...
 
-export { hostFitContentStyles };
+const hostFocusRingStyles = css`
+  :host {
+    --mdc-focus-ring-inner-color: var(--mds-color-theme-focus-0);
+    --mdc-focus-ring-middle-color: var(--mds-color-theme-focus-1);
+    --mdc-focus-ring-outer-color: var(--mds-color-theme-focus-2;
+
+    --mdc-focus-ring-inner-width: 0.125rem;
+    --mdc-focus-ring-middle-width: calc(2 * var(--mdc-focus-ring-inner-width));
```
cscheffauer, line 20: suggestion: i think here we would just do 0.125rem as well and on the other one the 0.0625rem, cause in the box-shadow we can do the offset, see comment below.

packages/components/src/utils/styles/index.ts
```diff
@@ -10,4 +10,27 @@ const hostFitContentStyles = css`
...
+    --mdc-focus-ring-outer-width: calc(0.0625rem + var(--mdc-focus-ring-middle-width));
+
+    --mdc-focus-ring-middle-offset: var(--mdc-focus-ring-inner-width);
+    --mdc-focus-ring-outer-offset: calc(var(--mdc-focus-ring-inner-width) + var(--mdc-focus-ring-middle-width));
+  }
+  
+  :host(:focus) {
+    position: relative;
+    box-shadow: 
+        0 0 0 var(--mdc-focus-ring-inner-width) var(--mdc-focus-ring-inner-color),
```
cscheffauer, line 30: https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow
if u follow this, we can build it with proper offsets here (instead of 0 0 0).

cscheffauer (changes requested):

cscheffauer (approved):

PR #901, opened 2024-10-17
feat: add mixin for focus ring

2020Deeya:
# Description

Add Mixin for the focus ring

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-374

### DIFF
--- a/packages/components/src/utils/mixins/FocusRingMixin.ts
+++ b/packages/components/src/utils/mixins/FocusRingMixin.ts
@@ -1,0 +1,38 @@
+import { LitElement } from 'lit';
+import { property } from 'lit/decorators.js';
+
+type Constructor<T = {}> = new (...args: any[]) => T;
+
+export interface FocusRingMixinInterface {}
+
+export const FocusRingMixin = <T extends Constructor<LitElement>>(superClass: T) => {
+  class FocusRingClass extends superClass {
+    @property({ type: Boolean, reflect: true })
+    protected focusRingVisible = false;
+
+    private handleFocusBound = this.handleFocus.bind(this);
+    private handleBlurBound = this.handleBlur.bind(this);
+
+    constructor(...args: any[]) {
+      super(...args);
+      this.addEventListener('focusin', this.handleFocusBound);
+      this.addEventListener('focusout', this.handleBlurBound);
+    }
+
+    handleFocus() {
+      this.focusRingVisible = true; 
+    }
+
+    handleBlur() {
+      this.focusRingVisible = false; 
+    }
+
+    override disconnectedCallback() {
+      super.disconnectedCallback();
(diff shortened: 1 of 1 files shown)


packages/components/src/utils/mixins/FocusRingMixin.ts
```diff
@@ -0,0 +1,39 @@
+import { LitElement } from 'lit';
+import { property } from 'lit/decorators.js';
+
+type Constructor<T = {}> = new (...args: any[]) => T;
```
cscheffauer, line 4: suggestion: we can move that into a shared type, given this is the same type in every mixin.

packages/components/src/utils/mixins/FocusRingMixin.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface FocusRingMixinInterface {}
+
+export const FocusRingMixin = <T extends Constructor<LitElement>>(superClass: T) => {
+  class FocusRingClass extends superClass {
+    @property({ type: Boolean, reflect: true })
```
cscheffauer, line 10: suggestion: this doesn't need to be here, since we only want to have a focusRingVisible class property, not a decorator. Just remove this line.

packages/components/src/utils/mixins/FocusRingMixin.ts
```diff
@@ -0,0 +1,39 @@
...
+    protected focusRingVisible = false;
+
+    private handleFocusBound = this.handleFocus.bind(this);
```
cscheffauer, line 13: question: is the binding necessary?

cscheffauer (changes requested):

packages/components/src/utils/mixins/FocusRingMixin.ts
2020Deeya: it's good to maintain the context of 'this'. But In our case, it's an overhead. We could remove it.

packages/components/src/utils/mixins/FocusRingMixin.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Handles focus and blur events to show/hide the focus ring.
+     */
+    private handleEvent = (event: Event) => {
+      if (event instanceof FocusEvent) {
+        switch (event.type) {
+          case 'focus':
+            this.focusRingVisible = true;
+            this.setAttribute('focusRingVisible', ''); // will be removed
```
2020Deeya, line 40: Temporarily added this line. Will be removed.

2020Deeya:
As discussed, we don't have any specific requirements for handling events via mixins, so we are closing this PR for now. It will be considered as a reference later.

PR #903, opened 2024-10-17
feat: Focus ring effect token updates

clairelikesemojis:
# Description
[dark/light/hc_black/hc_white] New theme token added for focus ring effect: 
- color-theme-focus-0
- color-theme-focus-1
- color-theme-focus-2

## Links
https://www.figma.com/design/yYfvFEEr3UkW1yUKkW7zEH/%F0%9F%92%8E-Theme---Color?m=auto&node-id=0-1&t=eaX0SQnrEQqxXjj0-1


packages/assets/tokens/src/theme/stable/dark.json
```diff
@@ -2594,6 +2594,32 @@
...
+        },
+        "2": {
+          "value": "{color.core.blue.40}",
+          "type": "color",
+          "$extensions": {
+            "studio.tokens": {
+              "modify": {
+                "type": "alpha",
+                "value": "0.35",
+                "space": "lch"
```
cscheffauer, line 2617: We need to test if this works in our tooling. @2020Deeya

cscheffauer (changes requested):

packages/assets/tokens/src/theme/stable/light.json
```diff
@@ -2611,6 +2611,32 @@
             "description": "linear-gradient(90deg, #932099 0%, #932099 90%, #932099 90%, #fd884e 0%)\n"
           }
         }
+      },
+      "focus": {
+        "0": {
+          "value": "{color.core.black-alpha.100}",
```
cscheffauer, line 2617: @clairelikesemojis i think this is a bug - like in light theme i don't think this should be black-alpha? can u check please?

cscheffauer (changes requested):

packages/assets/tokens/src/theme/stable/light.json
clairelikesemojis: Yes! Just fixed it! Thanks

cscheffauer (approved):

PR #910, opened 2024-10-21
chore(e2e): add tagname attribute test in text component

arjunkesava:
# Description

- This PR adds attribute check for `text component`.

## Links

Ticket: https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-366


packages/components/src/components/text/text.e2e-test.ts
```diff
@@ -63,4 +68,14 @@ test.describe('mdc-text', () => {
...
   }
+
+  for (const tagname of tagnameToTest) {
+    test(tagname, async ({ componentsPage }) => {
+      const text = await setup({ componentsPage, type: DEFAULTS.TYPE, tagname, children: textContent });
+
+      await test.step('tagname should be present on component', async () => {
+        expect(await text.getAttribute('tagname')).toBe(tagname);
+      });
+    });
```
supminn, line 79: question: We should test for a11y and visual regression when the tag name changes as well right? We'd need to ensure that the styles are intact but the HTML tags are changed.

arjunkesava: Done.

supminn (approved):

supminn (approved):

PR #911, opened 2024-10-21
fix(text): we added Text

arjunkesava:
# Description

Component "Text" is ready for consumption!
Text component is ready for consumption and considered stable.
See more details in [Storybook](https://momentum-design.github.io/momentum-design/en/components/).

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-371


packages/components/src/components/text/text.stories.ts
```diff
@@ -10,12 +10,12 @@ const render = (args: Args) => html`
 `;
```
cscheffauer, line 10: @arjunkesava as discussed, this PR already makes the feat (minor version) bump for the text:
https://github.com/momentum-design/momentum-design/pull/891
We don't want another minor bump, thus in this case set "fix(text):" or something else in the PR Title.

cscheffauer (changes requested):

supminn (approved):

cscheffauer (approved):

PR #914, opened 2024-10-22
chore(badge): build badge component of all types

arjunkesava:
# Description

- There are 4 types of badges:
  - notifcation
  - icons
  - counter
  - text
- The notification badge contains an icon of circle shape.
- Only the first 4 characters of the text will be displayed, rest will be hidden.
- The `counter` and `text` types were build using `mdc-text` component.
- The `notification` and `icons` types were build using `mdc-icon` component.

## Links

JIRA: [MOMENTUM-381](https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-381)

## Screenshots:
- All screenshot were shot in 400x zoom.

#### notification type:
image

#### icons type:
image

#### counter type:
image

#### text type:
image

### DIFF
--- a/packages/components/src/components/badge/badge.component.ts
+++ b/packages/components/src/components/badge/badge.component.ts
@@ -1,44 +1,22 @@
 import { CSSResult, html } from 'lit';
 import { property } from 'lit/decorators.js';
-import { ifDefined } from 'lit/directives/if-defined.js';
-import { StyleInfo, styleMap } from 'lit/directives/style-map.js';
 import { Component } from '../../models';
-import { DEFAULTS, WARNING_ICON_NAME } from './badge.constants';
+import { DEFAULTS } from './badge.constants';
 import styles from './badge.styles';
-import type { BadgeType } from './badge.types';
+import type { BadgeType, BadgeVariant } from './badge.types';
 
 /**
- * @slot - This is a default/unnamed slot
- *
- * @summary This is MyElement
- *
  * @tagname mdc-badge
  */
 class Badge extends Component {
   /**
    * Type of the badge
-   * Can be `regular`, `icon`, `text` or `warning`
+   * Can be `notification`, `icon`, `counter` and `text`
    *
-   * Default: `regular`
+   * Default: `notification`
    */
   @property({ type: String, reflect: true })
   type?: BadgeType = DEFAULTS.TYPE;
-
-  /**
-   * Scale of the badge (works in combination with length unit)
-   *
-   * Default: `1`
-   */
-  @property({ type: Number })
-  size?: number = DEFAULTS.SIZE;
-
-  /**
-   * Length unit attribute for scale
-   *
-   * Default: `rem`
-   */
-  @property({ type: String, attribute: 'length-unit' })
-  lengthUnit?: string = DEFAULTS.LENGTH_UNIT;
 
   /**
    * If `type` is set to `icon`, attribute `iconName` can
@@ -56,63 +34,40 @@ class Badge extends Component {
   @property({ type: String })
   text?: string;
 
-  private updateSize() {
-    // if (this.scale && this.lengthUnit) {
-    //   const value = `${this.scale}${this.lengthUnit}`;
-    //   this.style.height = value;
-    //   if (this.type !== 'text') {
-    //     this.style.width = value;
-    //   }
-    // }
-  }
+  /**
+   * badge variant
+   */
+  @property({ type: String })
+  variant?: BadgeVariant;
 
-  override updated(changedProperties: Map<string, any>) {
-    super.updated(changedProperties);
-    if (changedProperties.has('size') || changedProperties.has('lengthUnit')) {
-      this.updateSize();
+  @property({ type: Number })
+  counter?: number;
+
+  @property({ type: Number, attribute: 'max-counter' })
+  maxCounter?: number;
+
+  @property({ type: Boolean })
+  overlay?: boolean;
+
+  /**
+   * Aria-label attribute to be set for accessibility
+   */
+  @property({ type: String, attribute: 'aria-label' })
+  override ariaLabel: string | null = null;
+
+  private getBadgeContentBasedOnType() {
+    switch (this.type) {
+      default:
+        return html``;
     }
-  }
-
-  iconTemplate() {
-    return html`<div class="mdc-badge-icon-container">
-      <mdc-icon name=${ifDefined(this.iconName)} size="100" length-unit="%"></mdc-icon>
-    </div>`;
-  }
-
-  textTemplate() {
-    return html`${this.text}`;
-  }
-
-  warningTemplate() {
-    return html` <mdc-icon name="${WARNING_ICON_NAME}" class="mdc-badge-warning"></mdc-icon> `;
   }
 
   public override render() {
-    let content;
-    const size = `${this.size}${this.lengthUnit}`;
-    let sizeStyles: StyleInfo = { width: size, height: size };
-
-    switch (this.type) {
-      case 'regular':
-        content = html``;
-        break;
-      case 'icon':
-        content = this.iconTemplate();
-        break;
-      case 'text':
-        content = this.textTemplate();
-        // make width flexible when text -> only set the height:
-        sizeStyles = { height: sizeStyles.height };
-        break;
-      case 'warning':
-        content = this.warningTemplate();
-        break;
-      default:
-        content = html``;
-        break;
-    }
-
-    return html`<div class="mdc-badge-container" style=${styleMap(sizeStyles)}>${content}</div>`;
+    return html`
+      <div class="mdc-badge-container">
+        ${this.getBadgeContentBasedOnType()}
+      </div>
+    `;
   }
 
   public static override styles: Array<CSSResult> = [...Component.styles, ...styles];
--- a/packages/components/src/components/badge/badge.component.ts
+++ b/packages/components/src/components/badge/badge.component.ts
@@ -1,10 +1,12 @@
 import { CSSResult, html, TemplateResult } from 'lit';
+import { styleMap } from 'lit-html/directives/style-map.js';
+import { classMap } from 'lit-html/directives/class-map.js';
 import { property } from 'lit/decorators.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { Component } from '../../models';
 import { DEFAULTS } from './badge.constants';
 import styles from './badge.styles';
-import type { BadgeType, BadgeVariant } from './badge.types';
+import { BadgeType, BadgeVariant } from './badge.types';
 import '../icon';
 import '../text';
 
@@ -19,7 +21,7 @@ class Badge extends Component {
    * Default: `notification`
    */
   @property({ type: String, reflect: true })
-  type?: BadgeType = DEFAULTS.TYPE;
+  type: BadgeType = DEFAULTS.TYPE;
 
   /**
    * If `type` is set to `icon`, attribute `iconName` can
@@ -49,7 +51,7 @@ class Badge extends Component {
   @property({ type: Number, attribute: 'max-counter' })
   maxCounter: number = DEFAULTS.MAX_COUNTER;
 
-  @property({ type: String })
+  @property({ type: Boolean })
   overlay = false;
 
   /**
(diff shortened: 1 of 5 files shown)


packages/components/src/components/badge/badge.component.ts
```diff
@@ -1,44 +1,27 @@
...
+import '../icon';
+import '../text';
 
 /**
- * @slot - This is a default/unnamed slot
- *
- * @summary This is MyElement
- *
  * @tagname mdc-badge
  */
```
supminn, line 15: suggestion: could you please the description to badge component. We could mention about the various types that we offer.

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -5,22 +5,25 @@ const styles = [
...
     }
-
-    .mdc-badge-icon-container {
-      height: 80%;
-      width: 80%;
+    .mdc-badge-overlay {
+      border: 1px solid black;
+    }
+    .mdc-badge-text {
+      padding: 0 4px;
```
supminn, line 22: question: are we maintaining padding in px? I believe we decided to maintain it as rem during the discussion.

packages/components/src/components/badge/badge.component.ts
arjunkesava: Doesn't that fall in to storybook task?

packages/components/src/components/badge/badge.component.ts
```diff
@@ -56,63 +39,124 @@ class Badge extends Component {
...
+  ): TemplateResult {
+    const colorVariant = variant === undefined ? BadgeVariant.SECURE : variant;
+    const iconStyles = {
+      color: `var(--mds-color-theme-indicator-${colorVariant})`,
+    };
+    return html`
+      <mdc-icon
+        name="${ifDefined(iconName)}"
+        length-unit="${DEFAULTS.LENGTH_UNIT}"
+        size="${type === BadgeType.NOTIFICATION ? 1 : DEFAULTS.ICON_SIZE}"
```
supminn, line 100: question: Could you help me understand this conditional logic. `DEFAULTS.ICON_SIZE` is also set to `1`. Why do we need this condition?

packages/components/src/components/badge/badge.component.ts
```diff
@@ -56,63 +45,124 @@ class Badge extends Component {
...
+  private getBadgeIcon(
+    iconName: string,
+    ariaLabel: string | null,
+    variant: BadgeVariant,
+    type: BadgeType,
+    overlay: boolean,
+  ): TemplateResult {
+    const colorVariant = variant === undefined ? BadgeVariant.SECURE : variant;
+    const iconStyles = {
+      color: `var(--mds-color-theme-indicator-${colorVariant})`,
```
supminn, line 100: question: Do we support these colors for both `notification` and `icon` badges? As per the [Tech Breakdown](https://confluence-eng-gpk2.cisco.com/conf/pages/viewpage.action?pageId=563878053) my understanding is that we support the colors for notification badges only. Please correct me if mistaken.

packages/components/src/components/badge/badge.component.ts
```diff
@@ -56,63 +45,124 @@ class Badge extends Component {
...
+    return html`
+      <mdc-text
+        type="body-small-medium"
+        tagname="span"
+        class="mdc-badge-text ${classMap({ 'mdc-badge-overlay': overlay })}"
+      >
+        ${text}
+      </mdc-text>
+    `;
+  }
```
supminn, line 131: They need to support three predefined variants (default, attention, stable).

packages/components/src/components/badge/badge.component.ts
arjunkesava: Thank you for noticing it, it has to be 0.75, I will update it.

packages/components/src/components/badge/badge.component.ts
arjunkesava: - notification will always be blue (ie `secure` color code).
- for icons the consumer can change the variant (which will change the color).
- I will add an extra `type` check as well, to make it always blue for notification.

packages/components/src/components/badge/badge.component.ts
arjunkesava: As per the techbreakdown, the badge type `counter` and `text` don't support the variants.

packages/components/src/components/badge/badge.component.ts
arjunkesava: Fixed with https://github.com/momentum-design/momentum-design/pull/914/commits/ce2ad2be1a5b5dcc7ccf2b9cb4682b537dca2bfd

packages/components/src/components/badge/badge.component.ts
arjunkesava: Fixed with https://github.com/momentum-design/momentum-design/pull/914/commits/ce2ad2be1a5b5dcc7ccf2b9cb4682b537dca2bfd

packages/components/src/components/badge/badge.component.ts
supminn: Since we had doc comments w.r.t properties and attributes, I felt this was the only piece missing. 
I guess we could cover it as a part of storybook from next time! Thanks for the change.

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -5,22 +5,25 @@ const styles = [
...
-    .mdc-badge-icon-container {
-      height: 80%;
-      width: 80%;
+    .mdc-badge-overlay {
+      outline: 1px solid var(--mdc-badge-text-color);
+    }
+    .mdc-badge-text {
+      padding: 0 0.25rem;
+      border-radius: 6.25rem;
+      line-height: 1px;
```
supminn, line 24: nitpick: Can we use [unit less line-height](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height) value here?

packages/components/src/components/badge/badge.component.ts
```diff
@@ -1,44 +1,32 @@
...
 import { property } from 'lit/decorators.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
-import { StyleInfo, styleMap } from 'lit/directives/style-map.js';
 import { Component } from '../../models';
-import { DEFAULTS, WARNING_ICON_NAME } from './badge.constants';
+import { DEFAULTS } from './badge.constants';
 import styles from './badge.styles';
-import type { BadgeType } from './badge.types';
+import { BadgeType, BadgeVariant } from './badge.types';
+import '../icon';
```
cscheffauer, line 10: issue: this import statement needs to be moved to the index.ts file (same for text), cause this import registers the icon component.
When using the badge.component class only as a consumer, the icon shouldn't be registered automatically.
I think this would be beneficial to put into a convention.

packages/components/src/components/badge/badge.component.ts
```diff
@@ -1,44 +1,32 @@
...
+import { classMap } from 'lit-html/directives/class-map.js';
 import { property } from 'lit/decorators.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
-import { StyleInfo, styleMap } from 'lit/directives/style-map.js';
 import { Component } from '../../models';
-import { DEFAULTS, WARNING_ICON_NAME } from './badge.constants';
+import { DEFAULTS } from './badge.constants';
 import styles from './badge.styles';
-import type { BadgeType } from './badge.types';
+import { BadgeType, BadgeVariant } from './badge.types';
```
cscheffauer, line 9: nitpick: both are types, so we can do `import type` here

packages/components/src/components/badge/badge.component.ts
```diff
@@ -1,44 +1,32 @@
...
+ * Badges are often used to display notifications, counts, making them a useful tool for
+ * conveying information quickly without taking up much space.
  *
  * @tagname mdc-badge
  */
 class Badge extends Component {
   /**
    * Type of the badge
-   * Can be `regular`, `icon`, `text` or `warning`
+   * Can be `notification`, `icon` and `counter`
```
cscheffauer, line 24: docs: small description here would be useful of what those are. We can do it now, or in the storybook task - but just want to make sure this will get properly documented.

packages/components/src/components/badge/badge.component.ts
```diff
@@ -50,69 +38,122 @@ class Badge extends Component {
   iconName?: string;
 
   /**
-   * If `type` is set to `text`, attribute `text` can
-   * be used to pass in any text to be displayed in the badge.
+   * badge variant
```
cscheffauer, line 41: issue: this documentation is useless really - either we write it properly now or leave it to the storybook task.

packages/components/src/components/badge/badge.types.ts
```diff
@@ -1 +1,14 @@
-export type BadgeType = 'regular' | 'text' | 'icon' | 'warning';
+export enum BadgeType {
+  NOTIFICATION = 'notification',
+  ICON = 'icon',
+  COUNTER = 'counter'
+}
+
+export enum BadgeVariant {
```
cscheffauer, line 7: issue: there are a few issues with using enums in typescript:
https://thoughtbot.com/blog/the-trouble-with-typescript-enums

Maybe we should think about if we want to use enums or if its better to rely on a Object in this matter.
Might result in a convention as well.

packages/components/src/components/badge/badge.types.ts
```diff
@@ -1 +1,14 @@
-export type BadgeType = 'regular' | 'text' | 'icon' | 'warning';
+export enum BadgeType {
```
cscheffauer, line 1: suggestion: these are not types, but enums are values (i.e. constants) - we should move those into the constants file.
Also worth having this documented in a convention, like which files are used and whats the purpose of them. The purpose of the types file is to only have Typescript types, no values / constants what so ever.

packages/components/src/components/badge/badge.styles.ts
cscheffauer: issue: we should not override the line-height, but just use the existing variants of the text components.

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -5,22 +5,25 @@ const styles = [
...
+      color: var(--mdc-badge-text-color);
+      background-color: var(--mdc-badge-text-background-color);
```
cscheffauer, line 26: suggestion: this background color needs to be on the container, not the text.
The text part of the badge should never have a separate background

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -5,22 +5,25 @@ const styles = [
...
   css`
     :host {
-      --mdc-badge-icon-background-color: var(--mds-color-theme-background-accent-normal);
-      --mdc-badge-icon-color: var(--mds-color-theme-common-text-primary-normal);
+      --mdc-badge-text-color: var(--mds-color-theme-text-primary-normal);
+      --mdc-badge-text-background-color: var(--mds-color-theme-indicator-secure);
     }
-
     .mdc-badge-container {
+      max-height: 1rem;
```
cscheffauer, line 12: question: is there a reason to use max-height instead of height?

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -5,22 +5,25 @@ const styles = [
...
-      background-color: var(--mdc-badge-icon-background-color);
-      color: var(--mdc-badge-icon-color);
+      border-radius: 100%;
     }
-
-    .mdc-badge-icon-container {
-      height: 80%;
-      width: 80%;
+    .mdc-badge-overlay {
+      outline: 1px solid var(--mdc-badge-text-color);
```
cscheffauer, line 19: issue: this should be its own component token, cause the overlay color might be different than the text color.

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -5,22 +5,25 @@ const styles = [
...
+    }
+    .mdc-badge-text {
+      padding: 0 0.25rem;
+      border-radius: 6.25rem;
```
cscheffauer, line 23: question: why does the text have a border-radius applied as well? that doesn't really make sense to me, given the container does the border-radius styling.

packages/components/src/components/badge/badge.component.ts
```diff
@@ -50,69 +38,122 @@ class Badge extends Component {
...
-        content = html``;
-        break;
+        return html``;
     }
+  }
 
-    return html`<div class="mdc-badge-container" style=${styleMap(sizeStyles)}>${content}</div>`;
+  public override render() {
+    return html`
+      <div
```
cscheffauer, line 149: suggestion: this separate div is not necessary, we can just apply the role, aria-label and also styling on the host `<mdc-badge>` directly.

packages/components/src/components/badge/badge.component.ts
```diff
@@ -50,69 +38,122 @@ class Badge extends Component {
...
+   * @param overlay - whether the badge should have an overlay.
+   * @param counter - the number to be displayed on the badge
+   * @returns the template result of the text.
+   */
+  private getBadgeText(maxCounter: number, overlay: boolean, counter?: number): TemplateResult {
+    return html`
+      <mdc-text
+        type="body-small-medium"
+        tagname="span"
+        class="mdc-badge-text ${classMap({ 'mdc-badge-overlay': overlay })}"
```
cscheffauer, line 122: suggestion: this conditional overlay i think would be better to add to the container, instead of the text and icon separately.

packages/components/src/components/badge/badge.component.ts
```diff
@@ -50,69 +38,122 @@ class Badge extends Component {
...
+    `;
   }
 
-  public override render() {
-    let content;
-    const size = `${this.size}${this.lengthUnit}`;
-    let sizeStyles: StyleInfo = { width: size, height: size };
+  /**
+   * Method to generate the badge text and counter template.
+   * @param text - text to be displayed in the badge.
```
cscheffauer, line 112: docs: this line doesn't make sense anymore

packages/components/src/components/badge/badge.component.ts
```diff
@@ -50,69 +38,122 @@ class Badge extends Component {
...
+  /**
+   * If `type` is set to `counter` and if `counter` is greater than `maxCounter`,
+   * then it will return a string the maxCounter value as string.
+   * Otherwise, it will return a string representation of `counter`.
+   * If `counter` is not a number, it will return an empty string.
+   * @param maxCounter - the maximum limit which can be displayed on the badge
+   * @param counter - the number to be displayed on the badge
+   * @returns the string representation of the counter
+   */
+  private getCounterText(maxCounter: number, counter?: number): string {
```
cscheffauer, line 70: suggestion: both counter and maxCounter are set on the class itself, and can be get with `this.maxCounter` and `this.counter` - we don't need to pass them into this function as params.

packages/components/src/components/badge/badge.component.ts
```diff
@@ -50,69 +38,122 @@ class Badge extends Component {
...
+    const colorVariant = (variant === undefined || type === BadgeType.NOTIFICATION) ? BadgeVariant.SECURE : variant;
+    const iconStyles = {
+      color: `var(--mds-color-theme-indicator-${colorVariant})`,
+    };
+    return html`
+      <mdc-icon
+        name="${ifDefined(iconName)}"
+        length-unit="${DEFAULTS.LENGTH_UNIT}"
+        size="${type === BadgeType.NOTIFICATION ? DEFAULTS.NOTIFICATION_ICON_SIZE_IN_REM : DEFAULTS.ICON_SIZE}" 
+        aria-hidden="${ifDefined(ariaLabel ? 'true' : 'false')}"
```
cscheffauer, line 102: issue: if the ariaLabel is set, we make this icon aria-hidden? that doesn't work in terms of a11y

cscheffauer (changes requested):
first round of review, haven't tested it locally yet

packages/components/src/components/badge/badge.styles.ts
cscheffauer: we can set this as a rem value: 0.0625rem

packages/components/src/components/badge/badge.component.ts
cscheffauer: removing aria-label and aria-hidden from the icon will solve it.

packages/components/src/components/badge/badge.styles.ts
arjunkesava: The container class is removed totally.
So now, the border radius can only be set with text.

packages/components/src/components/badge/badge.component.ts
cscheffauer: sure - just to mention, you can just set the box-sizing css attribute and with that u can get it working.

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -5,22 +5,22 @@ const styles = [
...
-
-    .mdc-badge-container {
-      display: flex;
-      justify-content: center;
-      align-items: center;
-      border-radius: 9999px;
-      background-color: var(--mdc-badge-icon-background-color);
-      color: var(--mdc-badge-icon-color);
+    .mdc-badge-overlay {
+      outline: 0.063rem solid var(--mdc-badge-text-color);
```
cscheffauer, line 14: issue: this has been discussed before, the overlay is not the text color, but a different theme token (check Figma).
Also this whole outline line can be moved to a component token above.

image

cscheffauer, line 14: issue: 0.063rem is not correct, but 0.0625 is - we should not do any rounding etc. but stick to proper values if possible. In this case its not a infinite number, but a concrete 0.0625, thus we want to use it!

cscheffauer (changes requested):

packages/components/src/components/badge/badge.component.ts
```diff
@@ -50,69 +36,142 @@ class Badge extends Component {
...
+   */
+  private getCounterText(maxCounter: number, counter?: number): string {
+    if (counter === undefined || typeof counter !== 'number') {
+      return '';
+    }
+    if (counter > maxCounter) {
+      return `${maxCounter}+`;
+    }
+    // At any given time, the max limit should not cross 999.
+    if (counter > DEFAULTS.MAX_COUNTER_LIMIT) {
```
cscheffauer, line 76: issue: this needs to be placed before line 72 i think - if the maxCounter is 10000 and counter is 10001, it will return `10000+` - it ignores this line 76

packages/components/src/components/badge/badge.component.ts
```diff
@@ -50,69 +36,142 @@ class Badge extends Component {
...
+        name="${ifDefined(iconName)}"
+        length-unit="${DEFAULTS.LENGTH_UNIT}"
+        size="${DEFAULTS.ICON_SIZE}"
+      ></mdc-icon>
+    `;
   }
 
-  textTemplate() {
-    return html`${this.text}`;
+  private getIconClasses(overlay: boolean, iconVariant: string) {
```
cscheffauer, line 103: docs: every method requires a tsdoc, given that u mentioned method ts docs will be done in this PR, it should be also applied here.

packages/components/src/components/badge/badge.constants.ts
```diff
@@ -2,16 +2,42 @@ import utils from '../../utils/tag-name';
...
+  WARNING: 'warning',
+  ERROR: 'error',
+};
 
 const DEFAULTS = {
-  TYPE: 'regular' as const,
-  SIZE: 1,
+  TYPE: BADGE_TYPE.COUNTER,
   LENGTH_UNIT: 'rem',
+  MAX_COUNTER: 99999,
```
cscheffauer, line 31: why is the max_counter higher than the max_counter_limit? this is confusing - also i think 99999 is too big, like the maximum of the max_counter should be 999 as far as I understood.

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -5,22 +5,59 @@ const styles = [
   hostFitContentStyles,
   css`
     :host {
-      --mdc-badge-icon-background-color: var(--mds-color-theme-background-accent-normal);
-      --mdc-badge-icon-color: var(--mds-color-theme-common-text-primary-normal);
+      --mdc-badge-default-foreground-color: var(--mds-color-theme-common-text-primary-normal);
```
cscheffauer, line 8: this doesn't align in terms of naming with the existing constants, like the constants are "primary" and "secondary" - see the comment before, we first want to define the variants according to figma, then we can name here the component tokens accordingly.

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -5,22 +5,59 @@ const styles = [
...
+      --mdc-badge-secondary-foreground-color: var(--mds-color-theme-text-secondary-normal);
+      --mdc-badge-success-foreground-color: var(--mds-color-theme-text-success-normal);
+      --mdc-badge-warning-foreground-color: var(--mds-color-theme-text-success-normal);
```
cscheffauer, line 11: issue: in this line and in the line below there are copy paste errors. all 3 foreground colors from line 10-12 all point to theme-text-success-normal, which will result in this:
image

cscheffauer (changes requested):

packages/components/src/components/badge/badge.component.ts
arjunkesava: Done. I've updated it.

cscheffauer (approved):
lgtm 💯

PR #916, opened 2024-10-23
docs: add focus ring in storybook

2020Deeya:
# Description

Demonstrate focus ring behavior in different states (focus and blur) and on various elements.

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-375


packages/components/src/stories/FocusRing/subcomponent-focusring.styles.ts
```diff
@@ -0,0 +1,20 @@
+import { css } from 'lit';
+import { hostFitContentStyles, hostFocusRingStyles } from '../../utils/styles';
+
+const styles = [
+  hostFitContentStyles,
+  css`
+    :host {
```
cscheffauer, line 7: image

I am very confused on this - first:
- why is there a orange border? is this to show that the focus works on borders?
- I just tried it in storybook and the focus ring doesn't show up currently on the button at all - not on the outside or anything. Can u please check this?

packages/components/src/stories/FocusRing/subcomponent-focusring.constants.ts
```diff
@@ -0,0 +1,28 @@
...
+const TAG_NAME = utils.constructTagName('subcomponent-focusring');
+
+const DEFAULTS = {
+  SHAPE: 'mds-div' as const,
+};
+
+const FOCUS_RING_DESCRIPTION = `
+  **The Focus Ring provides a visual indication of focus for users navigating via keyboard or mouse.** 
+  
+  It is comprised of three distinct borders: black, blue, and light blue.
```
cscheffauer, line 13: suggestion: we don't need to mention this in particular, cause this could be misleading. Like if we say 3 borders, thats not necessarily true cause we use box shadows - also for a consumer this information doesn't really matter i think.

packages/components/src/stories/FocusRing/subcomponent-focusring.stories.utils.ts
```diff
@@ -0,0 +1,46 @@
...
+   */
+  @property({ type: String })
+  shape: string = DEFAULTS.SHAPE;
+
+  override render() {
+    switch (this.shape) {
+      case 'mds-div':
+        return html`<div tabindex="0" class="mds-div">Focusable Div</div>`;
+      case 'mds-anchor':
+        return html`<a href="#" class="mds-anchor">LinkText</a>`;
```
cscheffauer, line 25: issue: clicking on the link takes me out of storybook, i think this is not a great user behavior. I would vote for not having a href or so on the <a tag for now

packages/components/src/stories/FocusRing/subcomponent-focusring.stories.utils.ts
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * Default: 'mds-div'
+   */
+  @property({ type: String })
+  shape: string = DEFAULTS.SHAPE;
+
+  override render() {
+    switch (this.shape) {
+      case 'mds-div':
+        return html`<div tabindex="0" class="mds-div">Focusable Div</div>`;
```
cscheffauer, line 23: suggestion: i am not a big fan of this example, cause this is how components should NOT be built - like tabIndex=0 on a div is not accessible and has multiple issues. Not sure if we should show this here as a example.

packages/components/src/stories/FocusRing/subcomponent-focusring.stories.utils.ts
```diff
@@ -0,0 +1,46 @@
...
+      case 'mds-anchor':
+        return html`<a href="#" class="mds-anchor">LinkText</a>`;
+      case 'mds-button':
+        return html`<button class="mds-button">ClickMe</button>`;
+      case 'mds-radio':
+        return html`<input type="radio" id="radio" name="group" class="mds-radio" /><label for="radio"
```
cscheffauer, line 29: issue: when i choose this type and focus to it, the focus ring shows on the radio inside, but not on the outside as i would expect. Or maybe the orange border here is making it misleading - maybe we can remove the orange border?
image

packages/components/config/storybook/preview.js
```diff
@@ -116,7 +116,7 @@ const preview = {
     },
```
cscheffauer, line 116: issue on the PR title - this is not a feat, but maybe docs or something else?
we need to be careful in terms of deploying / naming the title if we want to follow a certain rythm in our release cadence.

cscheffauer (changes requested):

packages/components/src/stories/FocusRing/subcomponent-focusring.stories.ts
```diff
@@ -0,0 +1,37 @@
+import type { Meta, StoryObj, Args } from '@storybook/web-components';
+import { html } from 'lit';
+import './subcomponent-focusring.stories.utils';
+import { FOCUS_RING_DESCRIPTION } from './subcomponent-focusring.constants';
```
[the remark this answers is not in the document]
2020Deeya: Done.

packages/components/src/stories/FocusRing/subcomponent-focusring.stories.ts
```diff
@@ -0,0 +1,37 @@
...
+      control: { type: 'radio' },
+    },
+  },
+};
+
+export default meta;
+
+export const Primary: StoryObj = {
+  args: {
+    shape: SHAPES[0],
```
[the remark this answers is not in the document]
2020Deeya: Done.

packages/components/src/stories/FocusRing/subcomponent-focusring.styles.ts
2020Deeya: 1] Orange border is there in order to consider a scenario what if a border is applied to web component and how it behaves in focus state.
2] Color Modifier is not being considered properly due to which focus ring is not showing up. I overlooked that change locally while performing unit conversion, and it was not pushed. 

Fixed it. Focus ring should now be visible.

packages/components/src/stories/FocusRing/subcomponent-focusring.constants.ts
2020Deeya: Agreed. Removed it.

packages/components/src/stories/FocusRing/subcomponent-focusring.stories.utils.ts
2020Deeya: Removing the href attribute does not make the element focusable. Though, we could still make it work.
But I think, instead of removing href in anchor tag, we can show some another good example, ex: checkbox input

packages/components/src/stories/FocusRing/subcomponent-focusring.stories.utils.ts
2020Deeya: okay, let's not show div as an example.

packages/components/src/stories/FocusRing/subcomponent-focusring.stories.utils.ts
2020Deeya: As I mentioned above, The orange border illustrates how a border applied to the web component behaves in the focus state. 

We could remove it for now to avoid confusion.

packages/components/config/storybook/preview.js
2020Deeya: Named it docs.

packages/components/src/stories/FocusRing/subcomponent-focusring.styles.ts
```diff
@@ -0,0 +1,14 @@
...
+      font-family: sans-serif;
```
cscheffauer, line 8: nitpick: why setting the font-family here? no point in doing so given that we apply the font-family in the Themeprovider

packages/tools/token-builder/src/common/constants.ts
```diff
@@ -5,6 +5,7 @@ const SD_TRANSFORMS = {
   NAME_CTI_KABAB: 'name/cti/kebab',
   NAME_CTI_CAMEL: 'name/cti/camel',
   RESOLVE_MATH: 'ts/resolveMath',
+  COLOR_MODIFIER: 'ts/color/modifiers',
```
cscheffauer, line 8: suggestion: i understand that this change is necessary - you have to put this information properly in the PR description. Worth keeping in mind that since you are making changes to both components and tokens, that there will be now 2 version bumps / releases, one for components, one for tokens.
My preference would be to put this token change in a separate PR, but if we keep it in this one, we at least need a proper description in the PR to reflect whats there.

packages/components/src/stories/FocusRing/subcomponent-focusring.styles.ts
cscheffauer: issue: its like this now, which is also not correct - we want the "default" outline of the button etc to not be there, but just apply our own momentum focusring.
image

cscheffauer: issue: also the select looks very confusing, like consumers will definitely not be able to understand what "our" focus ring is, vs the browser ones.
image

cscheffauer (changes requested):

packages/components/src/stories/FocusRing/subcomponent-focusring.constants.ts
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable implicit-arrow-linebreak */
```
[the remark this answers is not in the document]
2020Deeya: It can be removed now. It was added due to some old code. Thanks for pointing it out. Missed it.

packages/components/src/stories/FocusRing/subcomponent-focusring.constants.ts
```diff
@@ -0,0 +1,12 @@
...
+import utils from '../../utils/tag-name';
+
+const TAG_NAME = utils.constructTagName('subcomponent-focusring');
+
+const DEFAULTS = {
+  SHAPE: 'mds-button' as const,
```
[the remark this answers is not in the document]
2020Deeya: Since it's a temporary component, we've chosen this tagname temporarily. 
Yes, If we want, we can keep standards names.

packages/components/src/stories/FocusRing/subcomponent-focusring.stories.utils.ts
```diff
@@ -0,0 +1,47 @@
+import { html, CSSResult } from 'lit';
+import { property } from 'lit/decorators.js';
+import { Component } from '../../models';
+import { DEFAULTS, TAG_NAME } from './subcomponent-focusring.constants';
+import styles from './subcomponent-focusring.styles';
+
+/**
+ * The Focus Ring provides a visual indication of focus for users navigating via keyboard or mouse.
```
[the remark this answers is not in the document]
2020Deeya: okay, Can we keep it like a comment? 
It was just to add short information about Focus ring.

packages/components/src/stories/FocusRing/subcomponent-focusring.styles.ts
2020Deeya: Sure. Removed it.

packages/tools/token-builder/src/common/constants.ts
2020Deeya: You're right! 
We should keep this change separate from this PR.

packages/components/src/stories/FocusRing/subcomponent-focusring.stories.ts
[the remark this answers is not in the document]
cscheffauer: this has not been addressed

cscheffauer (approved):
one leftover comment which needs to be addressed, other than that looking good.

cscheffauer (approved):

PR #917, opened 2024-10-23
docs: updated tsdoc to support few more properties

supminn:
# Description

Updated the convention docs for tsdoc to support a few more properties.

### DIFF
--- a/packages/components/conventions/component-tsdoc.md
+++ b/packages/components/conventions/component-tsdoc.md
@@ -25,6 +25,21 @@ Usage: can be multiple
 Describes the CSS Properties of the component, which can be used by consumers to override styling of the component.
 Usage: can be multiple
 
+### @csspart
+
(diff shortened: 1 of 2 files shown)


packages/components/conventions/component-tsdoc.md
```diff
@@ -25,6 +25,21 @@ Usage: can be multiple
...
+
+Describes the Shadow Parts of the component, allowing consumers to style specific elements within the shadow DOM.
+Usage: can be multiple
+
+### @dependency
+
+Describes the dependencies of the component, indicating which other packages or modules are required for its functionality.
+Usage: can be multiple
+
+### @event
```
cscheffauer, line 38: there is also a tsdoc file in src, which i think we need to update for avoiding Linting warnings.

cscheffauer (changes requested):
one small thing, other than that looking great.

cscheffauer (approved):

PR #918, opened 2024-10-29
docs: created a set of convention documentations

supminn:
# Description

Creating a list of convention documents that can be considered as guidelines while building our components.

## Links

[MOMENTUM-385](https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-385)

### DIFF
--- a/packages/components/conventions/component-code-structure.md
+++ b/packages/components/conventions/component-code-structure.md
@@ -20,10 +20,10 @@ When this command is run, a prompt will appear for the user to enter the compone
 
 ### Generated Structure
 
-Once the user enters the component name, a folder will be created within p`ackages/components/src/components/` with the specified name. The following files will be added inside this folder:
+Once the user enters the component name, a folder will be created within `packages/components/src/components/` with the specified name. The following files will be added inside this folder:
 
-1. `index.js`: This file serves as the entry point for the component. It typically exports the main component or other relevant modules, making them accessible for import in other parts of the application.
-1. `x.component.ts`: This file contains the presentational component. It should focus on props, avoiding selectors, dispatches, and side effects.
+1. `index.ts`: This file serves as the registration point for the web component, ensuring that the component is properly registered and available for use throughout the application.
+1. `x.component.ts`: This file contains the presentational component. It should focus on class defintion, reactive properties, render() and life cycle methods.
 1. `x.stories.ts`: This file is used for creating stories for the presentational component. The implementer can decide how to structure these stories.
 1. `x.e2e-test.ts`:This file contains end-to-end tests, including test hooks and utilities, as well as individual tests for the presentational component.
 1. `x.styles.ts`: This file is designated for adding CSS styling to the component.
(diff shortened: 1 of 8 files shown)


packages/components/conventions/component-code-structure.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Lowercase: The component name should be in lowercase.
+- No Prefixes: Avoid using any prefixes.
+- No Additional Hyphens: Do not include extra hyphens (e.g., avoid names like theme-provider).
+
+### Generated Structure
+
+Once the user enters the component name, a folder will be created within p`ackages/components/src/components/` with the specified name. The following files will be added inside this folder:
+
+1. `index.js`: This file serves as the entry point for the component. It typically exports the main component or other relevant modules, making them accessible for import in other parts of the application.
```
cscheffauer, line 25: we can mention here that in the index we register the web component, in the component file we define the class

packages/components/conventions/component-code-structure.md
```diff
@@ -0,0 +1,49 @@
...
+1. `x.e2e-test.ts`:This file contains end-to-end tests, including test hooks and utilities, as well as individual tests for the presentational component.
+1. `x.styles.ts`: This file is designated for adding CSS styling to the component.
+1. `x.constants.ts`: This file can be used for any constants that may be necessary for the component.
+
+### Additional Files (Optional)
+
+Developers may also include specific files for their components if needed:
+
+- `x.types.ts`: A types file for defining Typescript types.
+- `x.utils.ts`: A utilities file for complex functions, if necessary.
```
cscheffauer, line 37: possibly adding a fixtures file x.fixtures.ts

packages/components/conventions/constants-and-enums.md
```diff
@@ -0,0 +1,87 @@
...
+- **Simple Definition**: Union types offer a straightforward way to define sets of related constants.
+- **No Runtime Imports Needed**: There is no need to import any runtime values (like enums or objects).
+- **Structural Typing**: Supports structural typing without requiring additional linting rules.
+
+#### Disadvantages
+
+- **Limited to Serializable Values**: Only serializable values (like strings) can be used; functions or class instances cannot be defined as types.
+- **No Iteration or Reverse Mapping**: Union types do not support iteration over values or reverse mapping.
+
+## Conclusion
```
cscheffauer, line 85: suggestion: we need to say here what we concluded, like which approach are we following forward.

cscheffauer (changes requested):

packages/components/conventions/component-color-tokens.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Guidelines for Using Color Tokens
+
+1. **Consistent Usage**: Always utilize the defined theme and color tokens in your component designs. Avoid using custom colors or any values outside the provided tokens.
+1. **Component Specifications**: Ensure that every component specification includes a corresponding color token. This will help maintain uniformity and facilitate easier updates in the future.
+1. **Avoid Anomalies**: There should never be instances of using colors that are not defined in the token library. This helps prevent discrepancies and ensures that all components adhere to the design system.
+
+## Conclusion
+
+Adhering to the practice of using theme and color tokens within the Momentum Design System is crucial for maintaining consistency and ease of maintenance across our components. By ensuring every component specification references these tokens, we foster a cohesive design language that enhances both the developer experience and end-user satisfaction.
```
cscheffauer, line 17: same comment as above on color tokens

packages/components/conventions/component-css-rem-unit.md
```diff
@@ -0,0 +1,72 @@
...
+    .example {
+         font-size: 1.5rem; /* 1.5 * 16px = 24px */
+         padding: 1rem;    /* 1 * 16px = 16px */
+         margin: 2rem;     /* 2 * 16px = 32px */
+    }
+    ```
+
+3. **Avoid Using `px` Directly**:
+    - Refrain from using `px` for font sizes, margins, paddings, and other dimensions.
+    - Use `px` only when absolutely necessary, such as for border widths or media queries.
```
cscheffauer, line 63: i would recommend to not put border widths example here, cause we want to use REM there too - media queries makes sense yes.

packages/components/conventions/component-import.md
```diff
@@ -0,0 +1,83 @@
...
+declare global {
+    interface HTMLElementTagNameMap {
+        ['mdc-badge']: Badge;
+    }
+}
+
+export default Badge;
+```
+
+## Best Practices for Importing Components
```
cscheffauer, line 56: This section here is structured a bit weird, like its listing best practices, what we actually want to do and also advantages disadvantages.
Can we decluter this? Bullet point 1 is the Convention itself, like the Implementation guideline
Bullet 2 can be merged into bullet 2, like its only saying that we should NOT put the imports in the class - thats for me the same thing as number 1 - dependent imports only in index, not in class.

cscheffauer (changes requested):
few more things, otherwise looking good

cscheffauer (approved):

PR #921, opened 2024-10-29
test: add e2e and visual regression tests

2020Deeya:
# Description

Set up E2E tests and Visual regression tests for the Focus Ring

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-376


packages/components/src/stories/FocusRing/subcomponent-focusring.e2e-test.ts
```diff
@@ -0,0 +1,125 @@
...
+  });
+  const subComponent = componentsPage.page.locator('#firstComponent');
+  await subComponent.waitFor();
+  return subComponent;
+};
+
+// Applies focus to the component, handling WebKit-specific focus quirks
+const applyFocus = async (componentsPage: ComponentsPage, browserName: string) => {
+  if (browserName === 'webkit') {
+    await componentsPage.page.keyboard.down('Alt');
```
cscheffauer, line 28: suggestion: these 2 lines can be merged into:
`await componentsPage.page.keyboard.press('Alt+Tab')`
see here: https://playwright.dev/docs/api/class-keyboard#keyboard-press

packages/components/src/stories/FocusRing/subcomponent-focusring.e2e-test.ts
```diff
@@ -0,0 +1,125 @@
...
+  const boxShadow = await subComponent.evaluate((el: Element) => getComputedStyle(el).boxShadow);
+  await expect(boxShadow).toBe(expectedBoxShadow);
+};
+
+// Expected focus ring style for verification
+/* eslint-disable max-len */
+const focusRingStyle = 'rgb(0, 0, 0) 0px 0px 0px 2px, rgb(100, 180, 250) 0px 0px 0px 4px, rgba(100, 180, 250, 0.35) 0px 0px 0px 5px';
+
+// Checks if the component has the expected focus state (focused or not)
+const checkFocusState = async (subComponent: Locator, expectedState: boolean) => {
```
cscheffauer, line 46: suggestion: this util is not necessary, we can use `toBeFocused` instead
https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-focused

packages/components/src/stories/FocusRing/subcomponent-focusring.e2e-test.ts
```diff
@@ -0,0 +1,125 @@
...
+  const isFocused = await subComponent.evaluate((el) => document.activeElement === el);
+  expect(isFocused).toBe(expectedState);
+};
+
+// Applies blur to the component, with WebKit-specific handling for focus issues
+const applyBlur = async (componentsPage: ComponentsPage, subComponent: Locator, browserName: string) => {
+  if (browserName === 'webkit') {
```
cscheffauer, line 53: i think this is not necessary if we tab out of the element with Alt+Tab in webkit

packages/components/src/stories/FocusRing/subcomponent-focusring.e2e-test.ts
```diff
@@ -0,0 +1,125 @@
...
+    /* eslint-disable max-len */
+    html: `<mdc-subcomponent-focusring shape="${shape}" id='secondComponent' class="focus-ring"></mdc-subcomponent-focusring>`,
+  });
+  const subComponent = componentsPage.page.locator('#firstComponent');
+  await subComponent.waitFor();
+  return subComponent;
+};
+
+// Applies focus to the component, handling WebKit-specific focus quirks
+const applyFocus = async (componentsPage: ComponentsPage, browserName: string) => {
```
cscheffauer, line 26: this function name doesn't make sense - better would be `pressTab` or something.
cause just because we are pressing tab on the keyboard, it doesn't mean something has focus or not.

packages/components/src/stories/FocusRing/subcomponent-focusring.e2e-test.ts
```diff
@@ -0,0 +1,125 @@
...
+    // Test focus ring disappearance
+    await applyBlur(componentsPage, subComponent, browserName);
+    await checkFocusState(subComponent, false);
+    await checkFocusRing(subComponent, 'none');
+    await componentsPage.visualRegression.takeScreenshot(`focus-ring-disappearance-${shape}`, {
+      element: subComponent,
+    });
+
+    // Test rapid focus/blur interactions
+    await Promise.all(
```
cscheffauer, line 82: this has the potential for flakiness - Promise.all runs all the async tasks in parallel, which could lead to some unwanted states. I think it would be better to run these async tasks in series.

packages/components/src/stories/FocusRing/subcomponent-focusring.e2e-test.ts
```diff
@@ -0,0 +1,125 @@
...
+    );
+    await checkFocusRing(subComponent, 'none');
+    await componentsPage.visualRegression.takeScreenshot(`focus-ring-rapid-interaction-${shape}`, {
+      element: subComponent,
+    });
+  });
+};
+
+// Tests event propagation for focus ring within a parent-child structure to check interference
+const testEventPropagation = async (componentsPage: ComponentsPage, browserName: string, shape: string) => {
```
cscheffauer, line 96: i don't understand the purpose of this test - like in line 103 we are pressing tab, then the parent gets focused (which is the outer button). And thats it then? like i don't know what the subcomponent in the focusRing is supposed to do in this case?
I don't see a huge value in this test atm - can u please clean this up?

cscheffauer (changes requested):

packages/components/src/stories/FocusRing/subcomponent-focusring.e2e-test.ts
2020Deeya: Done

packages/components/src/stories/FocusRing/subcomponent-focusring.e2e-test.ts
2020Deeya: Agreed. 
The reason it was added was to resolve an ESLint issue that occurred when using a for loop instead of running the promises in parallel with Promise.all. 
Reverted it using eslint disable for this case.

packages/components/src/stories/FocusRing/subcomponent-focusring.e2e-test.ts
2020Deeya: As per investigation, It works fine for elements like button and checkbox, but it fails in Webkit due to inconsistent focus handling, , especially with elements like radio buttons. 
Handling it for radio button only.

cscheffauer (approved):

PR #923, opened 2024-10-30
fix: tech debt for icon component

arjunkesava:
# Description

- Disable all properties of icon component.
- Hide `aria-label` attribute only for "Decorative" story.

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-380

## Screenshots

- For decorative icon, the aria-label is hidden.
image

- For informative icon, the aria-label is visible:
image

- Disable all props for icon component.
image


cscheffauer (approved):
lgtm!

PR #925, opened 2024-10-30
chore: add storybook for badge component

arjunkesava:
# Description

- Add stories for badge component.
- Show `attributes` of badge based on type conditional checks.

#### Dot type:
image

#### Icon type:
image

#### Counter type:
image

#### Success type:
image

#### Warning type:
image

#### Error type:
image


cscheffauer (changes requested):

packages/components/src/components/badge/badge.stories.ts
```diff
@@ -1,17 +1,19 @@
...
     type="${args.type}"
-    icon-name="${args.iconName}"
-    size="${args.size}"
-    length-unit="${args.lengthUnit}"
+    icon-name="${args['icon-name']}"
     counter="${args.counter}"
-    maxCounter="${args.maxCounter}"
+    max-counter="${args['max-counter']}"
     variant="${args.variant}"
+    ?overlay=${args.overlay}
```
cscheffauer, line 16: issue: aria-label is not passed into here - no way to set it. Needs to be fixed.

cscheffauer (changes requested):

cscheffauer (approved):

PR #926, opened 2024-10-30
chore: update constants in text component

arjunkesava:
# Description

- Remove static strings of font type and valid text tag type.
- Move them into constants so that we can set them once and use them everywhere.

## Links

- https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-390


packages/components/src/components/text/text.component.ts
```diff
@@ -24,81 +23,28 @@ import { DEFAULTS } from './text.constants';
...
-   * - 'heading-large-regular'
-   * - 'heading-large-medium'
-   * - 'heading-large-bold'
-   * - 'heading-xlarge-regular'
-   * - 'heading-xlarge-medium'
-   * - 'heading-xlarge-bold'
-   * - 'headline-small-light'
-   * - 'headline-small-regular'
-   */
+  // Description for this property is moved to `./text.stories.ts` file
```
cscheffauer, line 26: issue: this description should stay here, cause tsdoc is also something consumers will see when they hover over a property when consuming in typescript. We always want this tsdoc on top of the property to be present.

packages/components/src/components/text/text.component.ts
```diff
@@ -24,81 +23,28 @@ import { DEFAULTS } from './text.constants';
...
-   * - 'h6'
-   * - 'p'
-   * - 'small'
-   * - 'span'
-   * - 'div'
-   *
-   * For instance, setting this attribute to `h2` will render the text element as an `h2` element.
-   * Note that the styling is determined by the `type` attribute.
-   */
+  // Description for this property is moved to `./text.stories.ts` file
```
cscheffauer, line 30: Same comment as above

cscheffauer (changes requested):
First review, have to test locally

PR #931, opened 2024-11-01
test(badge): e2e test for badge

Jason20020:
# Description

create e2e test for badge component

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-383


packages/components/src/components/badge/badge.component.ts
```diff
@@ -184,7 +188,8 @@ class Badge extends Component {
       case BADGE_TYPE.ERROR:
         return this.getBadgeIcon(ICON_NAMES_LIST.ERROR_ICON_NAME, overlay, ICON_STATE.ERROR);
       default:
-        return html``;
+        this.type = BADGE_TYPE.DOT;
+        return this.getBadgeDot(overlay);
```
arjunkesava, line 192: This is good. Instead of setting the `type` we can re-write this:
```typescript
  private getBadgeContentBasedOnType(): TemplateResult {
    const { counter, iconName, maxCounter, overlay, type, variant } = this;
    switch (type) {
 -
      case BADGE_TYPE.ICON:
        return this.getBadgeIcon(iconName || '', overlay, variant, type);
        ...
      case BADGE_TYPE.ERROR:
        return this.getBadgeIcon(ICON_NAMES_LIST.ERROR_ICON_NAME, overlay, ICON_STATE.ERROR);
+    case BADGE_TYPE.DOT:
+    default:
        return this.getBadgeDot(overlay);
    }
  }
 ```

File: packages/components/src/components/badge/badge.component.ts [hunk 2]
```diff
@@ -169,6 +170,9 @@ class Badge extends Component {
    * @returns the TemplateResult for the current badge type.
    */
   private getBadgeContentBasedOnType(): TemplateResult {
+    if (this.variant && !ALLOWED_ICON_VARIANT.includes(this.variant)) {
+      this.variant = ICON_VARIANT.PRIMARY;
+    }
```
arjunkesava, line 175: Please move this into `getBadgeIcon` method.
Currently, variant is used only on the icon type due to which changing it in the `getBadgeContentBasedOnType` is un-necessary.

packages/components/src/components/badge/badge.e2e-test.ts
```diff
@@ -1,22 +1,41 @@
...
   await componentsPage.mount({
     html: `
-        <mdc-badge />
+    <mdc-badge
+      ${restArgs.type ? `type="${restArgs.type}"` : ''}
+      ${restArgs.iconName ? `icon-name="${restArgs.iconName}"` : ''}
+      ${restArgs.variant ? `variant="${restArgs.variant}"` : ''}
+      ${restArgs.counter ? `counter="${restArgs.counter}"` : ''}
+      ${restArgs.maxCounter ? `max-counter="${restArgs.maxCounter}"` : ''}
+      ${restArgs.ariaLabel ? `ariaLabel="${restArgs.ariaLabel}"` : ''}
```
arjunkesava, line 25: HTML doesn't support camelCase attributes.
```suggestion
      ${restArgs.ariaLabel ? `aria-label="${restArgs.ariaLabel}"` : ''}
```

arjunkesava (changes requested):
Please rebase this branch and then resolve the comments.

cscheffauer (changes requested):

packages/components/src/components/badge/badge.component.ts
Jason20020: I tried to move this into getBadgeIcon, but it will need to check are the ICON_STATE include, also changing the function parameter will get me error from analyze prebuild. It makes the code more complicated.

arjunkesava (approved):
LGTM.

cscheffauer (approved):

PR #934, opened 2024-11-02
build: do not process png as svg

NotNestor:
# Description

In brand visuals svgs are being processed as part of the optimize svg build step.
Change so only svgs are processed in this stage and copy pngs to a new images folder
Update manifest to handle both locations

## Links

N/A

### DIFF
--- a/packages/assets/brand-visuals/config/momentum.json
+++ b/packages/assets/brand-visuals/config/momentum.json
@@ -5,7 +5,7 @@
       "flows": [
         {
           "id": "optimise-svgs",
-          "target": "./src/**/*.*",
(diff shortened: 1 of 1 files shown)


packages/assets/brand-visuals/config/momentum.json
```diff
@@ -30,9 +30,26 @@
             }
           }
         },
+        {
+          "id": "copy-images",
+          "target": "./src/**/*.png",
+          "destination": "./dist/images",
```
NotNestor, line 36: Currently only handling png as they are the only other file format this could be expanded to jpg/jpeg later if needed

cscheffauer (approved):

PR #939, opened 2024-11-05
style(presence): styling for presence commponent

Jason20020:
# Description

create presence component styling

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-395


2020Deeya (dismissed):
LGTM

packages/components/src/components/presence/presence.component.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+class Presence extends Component {
+  /**
+   * Type of the presence
+   */
+  @property({ type: String, reflect: true })
+  type = DEFAULTS.TYPE;
+
+  /**
+   * Size of the presence icon
```
cscheffauer, line 43: suggestion: as a consumer i will not be aware of what the different sizes are, we should document them here

packages/components/src/components/presence/presence.component.ts
```diff
@@ -0,0 +1,137 @@
...
+import { property } from 'lit/decorators.js';
+import { Component } from '../../models';
+import { DEFAULTS } from './presence.constants';
+import styles from './presence.styles';
+
+/**
+ * The `mdc-presence` component is a versatile UI element used to
+ * display the presence status of a user or entity within an avatar component.
+ *
+ * Supported presence types:
```
cscheffauer, line 12: this would be better suited in line 37

packages/components/src/components/presence/presence.constants.ts
```diff
@@ -0,0 +1,38 @@
...
+  SMALL: 'small',
+  MIDSIZE: 'midsize',
+  LARGE: 'large',
+  X_LARGE: 'x_large',
+  XX_LARGE: 'xx_large',
+};
+
+const DEFAULTS = {
+  TYPE: PRESENCE_TYPE.ACTIVE,
+  LENGTH_UNIT: 'rem',
```
cscheffauer, line 34: issue: length_unit is not used

packages/components/src/components/presence/presence.styles.ts
```diff
@@ -0,0 +1,106 @@
...
+      border-radius: 50%;
+      background-color: var(--mdc-presence-overlay-background-color);
+    }
+    .mdc-presence-icon__xx_small,
+    .mdc-presence-icon__x_small,
+    .mdc-presence-icon__small {
+      box-shadow: 0 0 0 0.09375rem var(--mdc-presence-overlay-background-color);
+    }
+    .mdc-presence-icon__midsize {
+      box-shadow: 0 0 0 0.129rem var(--mdc-presence-overlay-background-color);
```
cscheffauer, line 49: issue: i think these rem values (all of them) are not correct. Like i can see different values in Figma for this.

packages/components/src/components/presence/presence.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  private getIconClasses(): { [key: string]: boolean } {
+    const sizeClass = { [`mdc-presence-icon__${this.size}`]: true };
+    const backgroundClass = { [`mdc-presence-icon__${this.type}`]: true };
+    return {
+      ...backgroundClass,
+      ...sizeClass,
+    };
+  }
+
+  public override update(changedProperties: PropertyValues): void {
```
cscheffauer, line 119: suggestion: this doesn't need to be here

cscheffauer (changes requested):

packages/components/src/components/presence/presence.component.ts
```diff
@@ -0,0 +1,158 @@
...
+   * - `presenting`
+   * - `quiet`
+   * - `scheduled`
+   *
+   */
+  @property({ type: String, reflect: true })
+  type = DEFAULTS.TYPE;
+
+  /**
+   * Size of the presence icon in rem
```
cscheffauer, line 42: issue: a user will not pass in the size in rem, so the way this is written is confusing in a way that it could mean that a consumer should pass in rem - but its really the different values. We also don't need to provide that table indicating how much each value is in REM - just put the acceptable values there on their own

packages/components/src/components/presence/presence.component.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  /**
+   * Get the size of the presence based on the given size type
+   */
+  private get presenceSize() {
+    switch (this.size) {
+      case 'midsize':
+        return 1.15;
+      case 'large':
+        return 1.3;
```
cscheffauer, line 65: issue: these might indicate that those are rounded values - we would want to be more accurate on those

packages/components/src/components/presence/presence.component.ts
```diff
@@ -0,0 +1,158 @@
...
+   * - XX_LARGE: 2.25,
+   *
+   */
+  @property({ type: String, reflect: true })
+  size = DEFAULTS.SIZE;
+
+  /**
+   * Get the size of the presence based on the given size type
+   */
+  private get presenceSize() {
```
cscheffauer, line 60: issue: this function isn't used anymore, can be removed

cscheffauer (changes requested):

packages/components/src/components/presence/presence.component.ts
```diff
@@ -0,0 +1,156 @@
...
+      default:
+        return 'active-presence-small-filled';
+    }
+  }
+
+  /**
+   * This method generates the CSS classes for the presence.
+   * @returns - an object containing the CSS classes for the presence div.
+   */
+  private getPresenceClasses(): { [key: string]: boolean } {
```
cscheffauer, line 117: suggestion: this function can be avoided by putting this line directly in line 150, into the class map

packages/components/src/components/presence/presence.component.ts
```diff
@@ -0,0 +1,156 @@
...
+    return {
+      ...sizeClass,
+    };
+  }
+
+  /**
+   * This method generates the CSS classes for the presence icon.
+   * @returns - an object containing the CSS classes for the icon.
+   */
+  private getIconClasses(): { [key: string]: boolean } {
```
cscheffauer, line 128: suggestion: this function can be avoided by putting this line directly in line 142, into the class map

packages/components/src/components/presence/presence.component.ts
```diff
@@ -0,0 +1,156 @@
...
+    return {
+      ...backgroundClass,
+    };
+  }
+
+  /**
+   * Method to generate the presence icon.
+   * @returns the template result of the icon.
+   */
+  private getPresenceIcon(): TemplateResult {
```
cscheffauer, line 139: suggestion: i think we can just inline this into line 150, like there is nothing too complex about the icon here for why we need that separate method.

cscheffauer (changes requested):
some small things to cleanup, rest looks good

cscheffauer (approved):

PR #946, opened 2024-11-11
docs: pixel perfect values convention

2020Deeya:
# Description

In order to make sure sure correct Figma values are added in the codebase, follow the guidelines in pixel perfect convention guide. 

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-409


packages/components/conventions/pixel-perfect-values.md
```diff
@@ -0,0 +1,43 @@
+# Usage of Pixel Perfect values in momentum-design components
+
+## Purpose
+
+To ensure our components align with design specifications exactly as outlined in Figma, achieving pixel-perfect styling and consistency across the application.
+
+## Styling Guidelines
+
+### 1. Exact Heights, Sizes, and Spacing
+- Every component should match the exact height, width, padding, and margin values as shown in Figma.
```
cscheffauer, line 10: issue: sometimes the width and in rare cases also the height depends on the value which is passed in. We should only match the exact height if there are variants for it, like the Button has different size variants which are determining the height of the button - but not the width! the width depends on the content which is passed in.
I would recommend to refactor this line to accommodate for this. Our tech breakdowns should tell us if height, width should be fixed or not.

packages/components/conventions/pixel-perfect-values.md
```diff
@@ -0,0 +1,43 @@
...
+- Inspect designs carefully, paying attention to the smallest details, ensuring every layout and dimension reflects the Figma blueprint.
+
+### 2. Pixel-Perfect Approach
+- All components should replicate Figma specifications to a pixel-perfect level, ensuring a visually consistent user experience.
+- Perform cross-browser checks to maintain uniformity across various environments.
+
+### 3. Rounding of `rem` Values
+- **Rounding Strategy**: Round `rem` values to two decimal places to maintain accuracy and consistency, For example, if a calculated value is `1.876rem`, round it to `1.88rem`.
```
cscheffauer, line 18: issue: no, we should NOT round to two decimal places!

packages/components/conventions/pixel-perfect-values.md
```diff
@@ -0,0 +1,43 @@
...
+- **Conversion Convention**: Use `px` to `rem` conversions, following the px to rem unit conversion standards.
+- **Example**: If Figma specifies `30px`, use the base rem value (e.g., `30px / 16 = 1.875rem`) and round as per the strategy above.
+
+### 4. Consistent Sizing Tokens (Future Plan)
```
cscheffauer, line 22: suggestion: this point is not necessary for the convention

packages/components/conventions/pixel-perfect-values.md
```diff
@@ -0,0 +1,43 @@
...
+
+### 5. Detail-Oriented Execution
+- Attention to detail is critical. Confirm that component borders, shadows, line heights, font sizes, and other small but significant design elements reflect the design spec.
+- Regularly validate component styles with Figma to catch any discrepancies early.
+
+### 6. Documentation of Changes
+- Document any intentional deviations from Figma specifications, with the reason and approval details, to ensure transparency and track design choices.
+
+### 7. Consistency Across All Devices
+- Test components across different screen sizes and resolutions to ensure that pixel-perfect specifications hold up across all devices.
```
cscheffauer, line 34: issue: yes, that would be ideal, but we don't have our e2e testing setup for this yet. I would recommend to create a tech debt ticket for this instead and remove it from here.

packages/components/conventions/pixel-perfect-values.md
```diff
@@ -0,0 +1,43 @@
...
+
+---
+
+This convention ensures each component remains faithful to the design while allowing a smooth transition to tokenized layouts. Let's keep refining these practices as we integrate more components!
+
+### References
```
cscheffauer, line 40: suggestion: lets keep these links out from here, given that this is more internal information

cscheffauer (changes requested):

packages/components/conventions/pixel-perfect-values.md
2020Deeya: Tech Debt:
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-410.

cscheffauer (approved):

arjunkesava (approved):

PR #948, opened 2024-11-11
docs: button component along with storybook and documentation

supminn:
# Description
- Button component has 3 types:
  -  Pill button - Button that contains text value. It is commonly used for call to action, tags or filters.
  - Pill button with icons - Button containing icon either on the left or right side of the button. 
  - Icon button - Button is represented by just an icon without any text. 
- They have 3 variants - Primary, Secondary and Tertiary. 
- Different states a button can have are as follows: Active, Disabled, SoftDisabled (The button is currently disabled for user interaction; however, it remains focusable.)
- Colors:
  - Positive - Green button 
  - Negative - Red button 
  - Accent - Blue button 
  - Promotional - Purple button 
  - Default - Grey button 

We support the following sizes which vary based on the button type. As per our conventions guide, we shall use REM unit for button sizing.
- Pill button - 40, 32, 28, 24. 
- Icon button - 64, 52, 40, 32, 28 and 24.
    - Tertiary icon button can be of 20.
 
## Links
- [Tech Breakdown](https://confluence-eng-gpk2.cisco.com/conf/pages/viewpage.action?pageId=569871745)
- [Component and styling ticket](https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-386)


packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+import { getIconNameWithoutStyle, getIconSize, isValidIconSize, isValidPillSize } from './button.utils';
+
+/**
+ * button component, which ...
+ *
+ * @tagname mdc-button
+ *
+ * @slot default - This is a default/unnamed slot
+ *
+ * @cssprop --custom-property-name - Description of the CSS custom property
```
arjunkesava, line 15:
```suggestion
 * @cssproperty --custom-property-name - Description of the CSS custom property
```

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+ */
+class Button extends Component {
+  @property({ type: Boolean }) active: boolean = false;
+
+  @property({ type: Boolean }) disabled: boolean = false;
```
arjunkesava, line 20: Q: If we assign a value to a type when declaration, do we need to still need to specify the type.
```suggestion
  @property({ type: Boolean }) disabled = false;
```

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (softDisabled) {
+      element.setAttribute('soft-disabled', 'true');
+      element.setAttribute('aria-disabled', 'true');
+    } else {
+      element.removeAttribute('soft-disabled');
+      element.removeAttribute('aria-disabled');
+    }
+  }
+
+  private prevTabindex = 0;
```
arjunkesava, line 156: This should be moved to the top of this page.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  public override disconnectedCallback() {
+    super.disconnectedCallback();
+    this.removeEventListener('click', this.handleClick);
+    this.removeEventListener('keydown', this.handleKeyDown);
+    this.removeEventListener('keyup', this.handleKeyUp);
+    this.removeEventListener('focus', this.handleFocus);
+    this.removeEventListener('blur', this.handleBlur);
+  }
```
arjunkesava, line 59: This [article] suggests us to use javascript constructor for a single call events.
As the above events doesn't need to be re rendered / re attached again, we can follow this below approach:

```suggestion
  constructor() {
    super();
    this.role = 'button';
    this.handleButtonEvents();
  }

  private handleButtonEvents() {
    this.addEventListener('click', this.handleClick);
    this.addEventListener('keydown', this.handleKeyDown);
    this.addEventListener('keyup', this.handleKeyUp);
    this.addEventListener('focus', this.handleFocus);
    this.addEventListener('blur', this.handleBlur);
  }
```

Reg `removeEventListener`: we don't neeed to remove the events manually as JS will clean up when this component is no longer used.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+      this.type = BUTTON_TYPE.PILL_WITH_ICON;
+      Array.from(children).forEach((el) => {
+        if (el.nodeName === 'MDC-ICON') {
+          this.modifyIconSizeAndStyle(el);
+        }
+      });
+    }
+  }
+
+  private prevIconName = '';
```
arjunkesava, line 226: same here as well.

packages/components/src/components/button/button.utils.ts
```diff
@@ -0,0 +1,26 @@
+import { ICON_BUTTON_SIZES, PILL_BUTTON_SIZES } from './button.constants';
+
+const isValidIconSize = (
+  size: number,
+  variant: string,
+): number | boolean => Object.values(ICON_BUTTON_SIZES).includes(size)
+    && !(size === 20 && variant !== 'tertiary');
```
arjunkesava, line 7:
```suggestion
    && !(size === 20 && variant !== BUTTON_VARIANTS.TERTIARY);
```
Always use constants instead of strings, whenever wherever possible.

arjunkesava (changes requested):

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  @property({ type: Number }) override tabIndex = 0;
+
+  @state() private type = DEFAULTS.TYPE;
+
+  @state() private iconSize = 1;
+
+  constructor() {
+    super();
+    this.role = 'button';
```
cscheffauer, line 40: issue: we should make this a property, like consumers should be able to override the role - therefore we want to make it a property

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  public override connectedCallback() {
```
cscheffauer, line 43: issue: eventlisteners need to be moved to constructor, as discussed here:
https://hawkticehurst.com/2023/11/you-are-probably-using-connectedcallback-wrong/

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  private setColor(color: string) {
+    if (!Object.values(BUTTON_COLORS).includes(color) || this.variant === 'tertiary') {
+      this.setAttribute('color', `${DEFAULTS.COLOR}`);
+    } else {
+      this.setAttribute('color', color);
+    }
+  }
+
+  private setAriaPressed(element: HTMLElement, active: boolean) {
```
cscheffauer, line 130: issue: when a user passes in the attribute `aria-pressed` to the button (like controls it from the outside), this function will override it. Instead if a aria-pressed is passed in, we want that to take precidence.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (active) {
+      element.setAttribute('aria-pressed', 'true');
+    } else {
+      element.removeAttribute('aria-pressed');
+    }
+  }
+
+  private setAriaLabel(element: HTMLElement, ariaLabel: string) {
```
cscheffauer, line 138: issue: there is no point in having this function, since thats what the property declarator does automatically.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (!this.disabled && !this.softDisabled) {
+      if (this.onclick) {
+        this.onclick(event);
+      }
+    }
+  }
+
+  private handleFocus(event: FocusEvent) {
+    if (this.softDisabled) {
+      event.preventDefault();
```
cscheffauer, line 183: question: why is the default prevented here when softDisabled?

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+  private handleBlur(event: FocusEvent) {
+    if (this.softDisabled) {
+      event.preventDefault();
+    }
+    this.classList.remove('pressed');
+  }
+
+  private handleKeyDown(event: KeyboardEvent) {
+    if (!this.disabled && (event.key === 'Enter' || event.key === ' ')) {
+      event.preventDefault();
```
cscheffauer, line 196: issue: why is the default prevented here?

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+      this.classList.add('pressed');
```
cscheffauer, line 197: issue: pressing enter or space should fire onclick!

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+  private handleKeyUp(event: KeyboardEvent) {
+    if (!this.disabled && (event.key === 'Enter' || event.key === ' ')) {
+      this.handleClick(event as unknown as MouseEvent);
+      this.classList.remove('pressed');
+    }
+  }
+
+  private inferButtonType(children: HTMLCollection) {
+    if (children.length === 1) {
+      if (children[0].nodeName === 'MDC-ICON') {
```
cscheffauer, line 210: suggestion: hard coded values like MDC-ICON should always be set in constants. I think we can actually get the tagName constant from the icon component (import it from the constants file, thats important) and then we can do `toUppercase()` to make this string. This will make sure that in the future in case we change the tagName, we don't have to update this constant here.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+        this.type = BUTTON_TYPE.ICON;
+        this.modifyIconSizeAndStyle(children[0]);
+      } else {
+        this.type = BUTTON_TYPE.PILL;
+      }
+    } else if (Array.from(children).some((node) => node.nodeName === 'MDC-ICON')) {
```
cscheffauer, line 216: issue: what happens if there are 3 mdc-icons passed in? how will the button behave?

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  private prevIconName = '';
+
+  private modifyIconSizeAndStyle(node: ChildNode) {
+    const icon = node as HTMLElement;
+    icon.setAttribute('size', `${this.iconSize}`);
+    icon.setAttribute('length-unit', 'rem');
+    if (this.active) {
+      const name = icon.getAttribute('name')!;
+      this.prevIconName = name;
```
cscheffauer, line 234: issue: try to make a button active and then change the icon underneath, this will probably fail (haven't tried it though). My suggestion would be to just replace the suffix with `-filled` when active, otherwise just use the iconName of that node as it is.

packages/components/src/components/button/button.utils.ts
```diff
@@ -0,0 +1,26 @@
+import { ICON_BUTTON_SIZES, PILL_BUTTON_SIZES } from './button.constants';
+
+const isValidIconSize = (
```
cscheffauer, line 3: this is only used once, no reason to put this in a utils function.
Its also easier to read if we keep it in inline. Same for the isValidPillSize

cscheffauer (changes requested):
initial review, haven't tested locally

packages/components/src/components/button/button.component.ts
supminn: Yes, onclick is invoked within `handleKeyUp`. Adding it here would invoke the onclick method when the user is pressing and holding the enter/space key.

packages/components/src/components/button/button.component.ts
supminn: I have updated the docs section on https://github.com/supminn/momentum-design/pull/3 PR.

packages/components/src/components/button/button.component.ts
supminn: Do we need to create `handleButtonEvents`?
Can't we directly add them into the constructor code?

arjunkesava: It's always better to keep the constructor clean.
My suggestion is to have a separate method to register all event listeners at one place.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,388 @@
...
+   * Defaults to DEFAULTS.SIZE if invalid.
+   *
+   * @param size - The size to set.
+   */
+  private setSize(size: PillButtonSize | IconButtonSize) {
+    const isIconType = this.type === BUTTON_TYPE.ICON;
+    const isValidSize = isIconType
+      ? Object.values(ICON_BUTTON_SIZES).includes(size)
+      && !(size === ICON_BUTTON_SIZES[20] && this.variant !== BUTTON_VARIANTS.TERTIARY)
+      : Object.values(PILL_BUTTON_SIZES).includes(size as PillButtonSize);
```
arjunkesava, line 235: Suggestion: Please make this more readable. The logic is correct but due to multiple use of ternary operators, it becomes very tough to understand this snippet in one go.

arjunkesava (approved):
LGTM.

packages/components/src/components/button/button.stories.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export default meta;
+
+export const PillButton: StoryObj = {
+  args: {
+    children: 'Click Me',
+    active: false,
+    disabled: false,
+    'soft-disabled': false,
+    variant: BUTTON_VARIANTS.PRIMARY,
```
cscheffauer, line 86: issue: when i toggle between primary, secondary and tertiary the size of the button changes - every variant should be of the same size.

packages/components/src/components/button/button.stories.ts
```diff
@@ -0,0 +1,133 @@
+import type { Meta, StoryObj, Args } from '@storybook/web-components';
+import '.';
+import { html } from 'lit';
+import { BUTTON_COLORS, PILL_BUTTON_SIZES, BUTTON_VARIANTS, ICON_BUTTON_SIZES } from './button.constants';
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
+
+const render = (args: Args) => html`
```
cscheffauer, line 7: issue: tabIndex shows in storybook, but doesn't work. We either want it to work or we exclude it.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,399 @@
...
+  @property({ type: String }) color: ButtonColor = DEFAULTS.COLOR;
+
+  /**
+   * The tabindex of the button.
+   * @default 0
+   */
+  @property({ type: Number }) override tabIndex = 0;
+
+  /**
+   * The role of the button.
```
cscheffauer, line 117: suggestion: we should mention here when consumers should override the role. Its important that people are aware about this.

packages/components/src/components/button/button.stories.ts
```diff
@@ -0,0 +1,133 @@
...
+      description: 'The size of the button.',
+      control: 'select',
+      options: Object.values(PILL_BUTTON_SIZES),
+    },
+    color: {
+      description: 'The color of the button.',
+      control: 'select',
+      options: Object.values(BUTTON_COLORS),
+    },
+    ...classArgType,
```
cscheffauer, line 73: image
suggestion: i think if u add `@internal` in the tsdoc for all those private properties, they will not show up in storybook.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,399 @@
...
+   * @default button
+   */
+  @property({ type: String }) override role = 'button';
```
cscheffauer, line 120: issue: role button doesn't show up by default - this component is not accessible!
image

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,399 @@
...
+
+  /**
+   * Sets the variant attribute for the button component.
+   * If the provided variant is not included in the BUTTON_VARIANTS,
+   * it defaults to the value specified in DEFAULTS.VARIANT.
+   *
+   * @param variant - The variant to set.
+   */
+  private setVariant(variant: ButtonVariant) {
+    if (!Object.values(BUTTON_VARIANTS).includes(variant)) {
```
cscheffauer, line 219: suggestion: all these lines can be inlined too:
```javascript
this.setAttribute('variant', Object.values(BUTTON_VARIANTS).includes(variant) ? variant : DEFAULTS.VARIANT);
```

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,399 @@
...
+  /**
+   * Sets the soft-disabled attribute for the button.
+   * When soft-disabled, the button remains focusable but not clickable.
+   *
+   * @param element - The button element.
+   * @param softDisabled - The soft-disabled state.
+   */
+  private setSoftDisabled(element: HTMLElement, softDisabled: boolean) {
+    if (softDisabled) {
+      element.setAttribute('soft-disabled', 'true');
```
cscheffauer, line 285: issue: this line is not necessary, cause soft-disabled is already set to true. same with line 288

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,399 @@
...
+   * When disabled, the button is not focusable or clickable, and tabindex is set to -1.
+   * The previous tabindex is stored and restored when enabled.
+   * Also sets/removes aria-disabled attribute.
+   *
+   * @param element - The button element.
+   * @param disabled - The disabled state.
+   */
+  private setDisabled(element: HTMLElement, disabled: boolean) {
+    if (disabled) {
+      element.setAttribute('disabled', 'true');
```
cscheffauer, line 304: issue: same as above, you are setting the disabled attribute here again, it is already set. same for line 311

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,399 @@
...
+      element.setAttribute('aria-disabled', 'true');
+      this.prevTabindex = this.tabIndex;
+      this.tabIndex = -1;
+      element.setAttribute('tabindex', `${this.tabIndex}`);
```
cscheffauer, line 308: nitpick: `${}` is not necessary, you can do this.tabIndex directly

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,399 @@
...
+   * Handles the click event on the button.
+   * If the button is not disabled or soft-disabled, the onclick event is triggered.
+   * The onclick method is provided by the consumer.
+   *
+   * @param event - The mouse event.
+   */
+  private handleClick(event: MouseEvent) {
+    if (!this.disabled && !this.softDisabled) {
+      if (this.onclick) {
+        this.onclick(event);
```
cscheffauer, line 327: will the onClick event also be triggered cause of this or only the onclick method here?

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,399 @@
...
+      this.type = BUTTON_TYPE.PILL_WITH_ICON;
+    } else if (!slot && (this.prefixIcon || this.postfixIcon)) {
+      this.type = BUTTON_TYPE.ICON;
+    } else {
+      this.type = BUTTON_TYPE.PILL;
+    }
+  }
+
+  private handleSlotchange(event: Event) {
+    const slot = event.target as HTMLSlotElement;
```
cscheffauer, line 372: nitpick: this line is not necessary, we can pass event.target into the method directly

cscheffauer (changes requested):

packages/components/src/components/button/button.styles.ts
```diff
@@ -0,0 +1,198 @@
...
+    --mdc-button-primary-background-color: var(--mds-color-theme-common-button-promotion-normal);
+    --mdc-button-primary-hover-background-color: var(--mds-color-theme-common-button-promotion-hover);
+    --mdc-button-primary-pressed-background-color: var(--mds-color-theme-common-button-promotion-active);
+
+    --mdc-button-secondary-color: var(--mds-color-theme-text-primary-normal);
+    --mdc-button-secondary-border-color: var(--mds-color-theme-outline-promotion-normal);
+    --mdc-button-secondary-hover-background-color: var(--mds-color-theme-button-secondary-hover);
+    --mdc-button-secondary-pressed-background-color: var(--mds-color-theme-button-secondary-pressed);
+  }
+`, hostFocusRingStyles];
```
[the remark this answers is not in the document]
cscheffauer: actually was due to local setup, resolved

packages/components/src/components/button/button.utils.ts
```diff
@@ -0,0 +1,30 @@
+import { ICON_BUTTON_SIZES } from './button.constants';
+
+/**
+ * Returns the icon size multiplier based on the provided button size.
+ *
+ * @param size - The size of the button.
+ * @returns The multiplier for the icon size.
+ */
+const getIconSize = (size: number): number => {
```
2020Deeya, line 9: Suggestion:  Here, size is typed as number, but valid values are constrained by ICON_BUTTON_SIZES. A number type allows invalid input without compile-time error. We could use a union type if possible. 

type IconButtonSize = keyof typeof ICON_BUTTON_SIZES;
const getIconSize = (size: IconButtonSize): number => { ... }

packages/components/src/components/button/button.utils.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Returns the name of the icon without the style suffix.
+ *
+ * @param iconName - The name of the icon.
+ * @returns The name of the icon without the suffix.
+ */
+const getIconNameWithoutStyle = (iconName: string): string => {
+  const iconParts = iconName.split('-');
+  const variants = ['bold', 'filled', 'regular', 'light'];
```
2020Deeya, line 26: Suggestion: Move variants to a shared constant outside of this function.

packages/components/src/components/button/button.utils.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+/**
+ * Returns the name of the icon without the style suffix.
+ *
+ * @param iconName - The name of the icon.
+ * @returns The name of the icon without the suffix.
+ */
+const getIconNameWithoutStyle = (iconName: string): string => {
+  const iconParts = iconName.split('-');
```
2020Deeya, line 25: Question: Will there be any case when the icon name is empty or does not include a variant? If yes, we should handle it I think.

packages/components/src/components/button/button.styles.ts
```diff
@@ -0,0 +1,198 @@
...
+  :host([variant="tertiary"][soft-disabled]){
+    color: var(--mdc-button-tertiary-disabled-color);
+    background: var(--mdc-button-tertiary-disabled-background-color);
+    cursor: auto;
+  }
+  
+  :host([size="64"]){
+    height: var(--mdc-button-size-64);
+    padding: 1rem;
+    border-radius: 6.25rem;
```
2020Deeya, line 99: Suggestion: I see redundancy of border-radius property with same value. We could defined it on top and use that variable at all the places.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,399 @@
...
+ *
+ * @dependency mdc-icon
+ *
+ * @tagname mdc-button
+ *
+ * @slot default slot - Text label of the button.
+ */
+class Button extends Component {
+  /**
+   * Indicates whether the button is active.
```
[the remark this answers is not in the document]
supminn: Could you help me out with this one? How can i elaborate on explaining this property?

packages/components/src/components/button/button.component.ts
supminn: typescript was throwing an error to convert number to a string. Hence, I converted it into a template literal

packages/components/src/components/button/button.utils.ts
supminn: these are constants pertaining to the icon names and not button. Hence, I added it within the function. They are not scoped outside this function.

packages/components/src/components/button/button.utils.ts
supminn: As far as I understand, We support the list of icons from - https://momentum-design.github.io/momentum-design/en/tokens/icons/ 
I do not see any icon name without the variant.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,255 @@
...
+  private handleKeyDown(event: KeyboardEvent) {
+    if (!this.disabled && (event.key === 'Enter' || event.key === ' ')) {
+      event.preventDefault();
+      this.classList.add('pressed');
+    }
+  }
+
+  private handleKeyUp(event: KeyboardEvent) {
+    if (!this.disabled && (event.key === 'Enter' || event.key === ' ')) {
+      this.handleClick(event as unknown as MouseEvent);
```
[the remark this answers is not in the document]
supminn: thanks for this observation. It seems to be working in the same way across browsers. I have made necessary changes.

packages/components/src/components/button/button.component.ts
supminn: With the structural changes (removing named slots and adding icon attributes), we no longer consider mdc-icon as an icon button if passed within the slot.
We do support 3 mdc-icons, but it would be called as a "pill-with-icon button"

packages/components/src/components/button/button.stories.ts
```diff
@@ -0,0 +1,133 @@
...
+    disabled: false,
+    'soft-disabled': false,
+    variant: BUTTON_VARIANTS.PRIMARY,
+    size: PILL_BUTTON_SIZES[32],
+    color: BUTTON_COLORS.DEFAULT,
+  },
+};
+
+export const IconButton: StoryObj = {
+  argTypes: {
```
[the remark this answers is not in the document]
supminn: Yes, this is no longer an icon button. It is 'pill-with-icon' button. The empty string is also consider as text.

packages/components/src/components/button/button.component.ts
supminn: I'm not sure on how we would achieve that. This is the simplest form I could add the logic in. I have added parenthesis around the condition. Let me know if that helps.

packages/components/src/components/button/button.stories.ts
supminn: Not sure why it's behaving in this way. The changes are reflecting when I update it directly on the tagname.
I have excluded it from storybook.

packages/components/src/components/button/button.stories.ts
```diff
@@ -0,0 +1,119 @@
...
+export const PillButton: StoryObj = {
+  args: {
+    children: 'Click Me',
+    active: false,
+    disabled: false,
+    'soft-disabled': false,
+    variant: BUTTON_VARIANTS.PRIMARY,
+    size: PILL_BUTTON_SIZES[32],
+    color: BUTTON_COLORS.DEFAULT,
+  },
```
arjunkesava, line 75: Please set the default values on storybook, it becomes very tough to set a value as storybook will use string to object.

image

arjunkesava (approved):
LGTM.

One minor non-blocking comment.

arjunkesava (approved):

packages/components/src/components/button/button.styles.ts
```diff
@@ -0,0 +1,197 @@
...
+    --mdc-button-height-size-40: 2.5rem;
+    --mdc-button-height-size-32: 2rem;
+    --mdc-button-height-size-28: 1.75rem;
+    --mdc-button-height-size-24: 1.5rem;
+    --mdc-button-height-size-20: 1.25rem;
+
+    --mdc-button-line-height-size-40: var(--mds-font-lineheight-body-large);
+    --mdc-button-line-height-size-32: var(--mds-font-lineheight-body-large);
+    --mdc-button-line-height-size-28: var(--mds-font-lineheight-body-midsize);
+    --mdc-button-line-height-size-24:  var(--mds-font-lineheight-body-small);
```
2020Deeya, line 44: issue: Minor spacing issue, not a blocker one

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,443 @@
...
+   *
+   * @param element - The button element.
+   * @param disabled - The disabled state.
+   */
+  private setDisabled(element: HTMLElement, disabled: boolean) {
+    if (disabled) {
+      element.setAttribute('aria-disabled', 'true');
+      this.prevTabindex = this.tabIndex;
+      this.tabIndex = -1;
+      element.setAttribute('tabindex', `${this.tabIndex}`);
```
cscheffauer, line 352: suggestion: this can be made easier by using the reflect:true on the property, cause then u can just delete line 352 & 356.

packages/components/src/components/button/button.stories.ts
```diff
@@ -0,0 +1,128 @@
...
+  <mdc-button 
+  ?active="${args.active}"
+  ?disabled="${args.disabled}"
+  ?soft-disabled="${args['soft-disabled']}"
+  variant="${args.variant}"
+  size="${args.size}"
+  color="${args.color}"
+  prefix-icon="${args['prefix-icon']}"
+  postfix-icon="${args['postfix-icon']}"
+  type="${args.type}"
```
cscheffauer, line 17: issue: role is not passed here.

packages/components/src/components/button/button.component.ts
```diff
@@ -0,0 +1,443 @@
...
+ * - **Pill button**: A button that contains text value. Commonly used for call to action, tags, or filters.
+ * - **Pill button with icons**: A button containing an icon either on the left or right side of the button.
+ * - **Icon button**: A button represented by just an icon without any text.
+ * The type of button is inferred based on the presence of slot and/or prefix and postfix icons.
+ *
+ * @dependency mdc-icon
+ *
+ * @tagname mdc-button
+ *
+ * @slot default slot - Text label of the button.
```
cscheffauer, line 54: issue: the default slot is wrong here, like this makes it actually a named slot from a docs perspective.
it should be `@slot - Text label ...`

cscheffauer (changes requested):

cscheffauer (approved):

PR #950, opened 2024-11-12
chore: update strings with consts in prescence component

arjunkesava:
# Description

- Update strings in the component with constants.
- This will help reduce the risk of updating the fixed values of both type and size properties.

## Links

- NA

### DIFF
--- a/packages/components/src/components/presence/presence.component.ts
+++ b/packages/components/src/components/presence/presence.component.ts
@@ -1,7 +1,7 @@
 import { CSSResult, html } from 'lit';
 import { property } from 'lit/decorators.js';
 import { Component } from '../../models';
-import { DEFAULTS } from './presence.constants';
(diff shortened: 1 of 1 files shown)


Jason20020 (dismissed):

packages/components/src/components/presence/presence.component.ts
```diff
@@ -56,19 +56,18 @@ class Presence extends Component {
...
+      case PRESENCE_SIZE.XX_LARGE:
         return 2.25;
-      case 'xx_small':
-      case 'x_small':
-      case 'small':
+      case PRESENCE_SIZE.XX_SMALL:
+      case PRESENCE_SIZE.X_SMALL:
+      case PRESENCE_SIZE.SMALL:
       default:
-        this.size = DEFAULTS.SIZE;
```
Jason20020, line 71: This should not be remove, the value will back to default if invalid value passed in

Jason20020 (approved):

PR #951, opened 2024-11-12
fix: add defaults in tsdoc

2020Deeya:
# Description

Add `@default` property in jsdoc to maintain readability across Storybook for consumers. 

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-407


packages/components/conventions/component-tsdoc.md
```diff
@@ -40,6 +40,11 @@ Usage: can be multiple
 Describes an event that can be emitted by the component, detailing the specific actions or occurrences that can trigger it.
 Usage: can be multiple
 
+### @default
+
+Indicates the initial value a property or parameter takes if none is provided.
+Usage: can be multiple
```
supminn, line 46: issue: we cannot have multiple default values.

packages/components/src/components/iconprovider/iconprovider.component.ts
```diff
@@ -41,23 +41,26 @@ class IconProvider extends Provider<IconProviderContext> {
...
+   * File extension of icons
+   * @default svg
    */
   @property({ type: String, attribute: 'file-extension', reflect: true })
   fileExtension?: string = DEFAULTS.FILE_EXTENSION;
 
   /**
-   * Length unit used for sizing of icons, default: 'em'
+   * Length unit used for sizing of icons
+   * @default em
```
supminn, line 52: question (not related to docs): why do we have the default value as em and not rem?

packages/components/src/components/iconprovider/iconprovider.constants.ts
```diff
@@ -13,6 +13,7 @@ const DEFAULTS = {
     em: 1,
     rem: 1,
   } as Record<string, number>,
+  DEFAULT_SIZE: 1,
```
supminn, line 16: question: if we are adding a new property to DEFAULTS, do we still need to have `LENGTH_UNIT_SIZE` within the DEFAULTS object?

packages/components/src/tsdoc.json
```diff
@@ -30,6 +30,11 @@
       "tagName": "@dependency",
       "syntaxKind": "modifier",
       "allowMultiple": true
+    },
+    {
+      "tagName": "@default",
+      "syntaxKind": "modifier",
+      "allowMultiple": true
```
supminn, line 37: issue: we cannot have multiple default values.

supminn (changes requested):

packages/components/conventions/component-tsdoc.md
2020Deeya: You're right, it got overlooked.

packages/components/src/components/iconprovider/iconprovider.constants.ts
2020Deeya: I think, we should still keep it, because it defines default values for the units.

supminn (dismissed):

supminn (approved):

PR #956, opened 2024-11-14
fix: update text component styling

2020Deeya:
# Description

Update text styles to make sure consumers can override the styles in the text component.

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-416


packages/components/src/components/text/fonts.styles.ts
```diff
@@ -1,231 +1,244 @@
...
+  :host([tagname='h2']),
+  :host([tagname='h3']),
+  :host([tagname='h4']),
+  :host([tagname='h5']),
+  :host([tagname='h6']),
+  :host([tagname='p']),
+  :host([tagname='small']),
+  :host([tagname='span']),
+  :host([tagname='div']) {
+    all: unset; /* This removes default styles applied by the browser */
```
cscheffauer, line 14: issue: we should not override the browser defaults here if possible, if the browser has a default we should just keep it.

cscheffauer (changes requested):
general issue: the expectation is that this PR doesn't change the screenshots at all

packages/components/src/components/text/fonts.styles.ts
2020Deeya: yes, Screenshots changes are temporary. They will be reverted back. It was to pushed to check something.

2020Deeya:
Fresh PR is created instead of this. Hence, closing this one.

PR #965, opened 2024-11-20
docs: convention-no-relative-workspace-imports

cscheffauer:
# Description

- Adding new convention to avoid relative workspace imports and instead use direct package imports
- No change to output


Jason20020 (approved):

packages/components/conventions/no-relative-imports-workspace.md
supminn, line 1: suggestion: I believe this is a good place to include - while we import other modules into our workspace, please keep in mind that we are not causing circular dependencies.

cscheffauer: i think it doesn't fit in this convention, just tried to add it and it doesn't really make sense.
I have a feeling that circular dependencies can be catched by E2E tests or other tooling (ESLint or so), so I created a tech debt issue to improve that.

PR #967, opened 2024-11-21
test: create utility function for creating sticker sheet and reduce screenshots

Jason20020:
# Description

- Create method that allow to create stickersheets on the page in e2e tests
- Reduce the amount of screenshots by using stickerSheet utility
- Address previous PR import type issue
## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-424
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-425


packages/components/src/components/icon/icon.types.ts
```diff
@@ -0,0 +1,3 @@
+import type IconNames from '@momentum-design/icons/dist/types/types';
```
Jason20020, line 1: Using import type to prevent circular dependencies

packages/components/config/playwright/setup/utils/visual-regression.ts
```diff
@@ -1,5 +1,6 @@
 import { Page, expect } from '@playwright/test';
 import type { ScreenShotOptions } from '../types';
+import { ComponentsPage } from '../../setup';
```
supminn, line 3: By importing `ComponentsPage` within this file, we are creating a cyclic link between this file and `Components.page.ts`
We would need to use this in a different way.

supminn (changes requested):
initial review - I pulled in the branch locally and found many linting issues.

packages/components/config/playwright/setup/utils/visual-regression.ts
```diff
@@ -47,6 +48,90 @@ class VisualRegression {
...
+    .map(
+      (combination) =>
+        `<${componentTag} ${primaryKey}="${primaryValue}" ${combination}>${children}</${componentTag}>`,
+    )
+    .join('')}
+          </div>`;
+        }
+
+        markup += rowMarkup;
+      }
```
supminn, line 116:
```suggestion
 for (const primaryValue of Object.values(primaryValues)) {
        const combinations = otherAttributes.reduce<string[]>(
          (acc, [key, values]) => {
            const valueArray = Object.values(values);
            return acc.flatMap((prev) =>
              valueArray.map((currVal) => `${prev} ${key}="${currVal}"`));
          },
          [''],
        );

        const rowMarkup = `<div class="componentRowWrapper">
          ${combinations.map((combination) => `<${componentTag} 
        ${primaryKey}="${primaryValue}"${combination}>${children}</${componentTag}>`).join('')}
        </div>`;

        markup += rowMarkup;
      }
```

suggestion: i think we could simply this for-loop as suggested.

packages/components/config/playwright/setup/utils/visual-regression.ts
```diff
@@ -35,7 +36,7 @@ class VisualRegression {
         threshold: CONSTANTS.VISUAL_REGRESSION.THRESHOLD,
         maxDiffPixelRatio: CONSTANTS.VISUAL_REGRESSION.MAX_DIFF_PIXELS_RATIO,
       });
-      
+
       // LTR Screenshot
       await this.page.evaluate(() => {
```
2020Deeya, line 41: Suggestion: Avoid Repetitive document.documentElement.setAttribute Calls and abstract this into a helper function.

packages/components/config/playwright/setup/utils/visual-regression.ts
```diff
@@ -47,6 +48,90 @@ class VisualRegression {
...
+   * @param attributes - Attributes to apply to the components, with key-value pairs representing attribute
+   *                     names and their possible values. The values are defined as an object of key-value pairs.
+   *
+   * @returns Locator for the component list containing all generated components.
+   */
+  async createStickerSheet(
+    componentsPage: ComponentsPage,
+    componentTag: string,
+    attributes: Record<string, Record<string, string>>,
+    children = '',
```
supminn, line 66: issue: Some of the components (ex: badge, icon) do not have `children`. Some other components might need to add a few attributes by default. Hence, I'd suggest you to make the following changes.
```
async createStickerSheet(
    componentsPage: ComponentsPage,
    componentTag: string,
    attributes: Record<string, Record<string, string>>,
    defaultAttributes?: Record<string, Record<string, string>>
)
```
Here, `defaultAttributes` could be defined as follows (i'm taking an example of the button component)
```
defaultAttributes:{
children: 'Label',
'prefix-icon': 'placeholder-bold'
}
```

supminn (changes requested):

supminn (changes requested):
minor changes requested. Functionality looks good.

packages/components/config/playwright/setup/utils/visual-regression.ts
supminn: handled on https://github.com/momentum-design/momentum-design/pull/971

supminn (approved):
LGTM

PR #969, opened 2024-11-25
fix: handle hight contrast issues

2020Deeya:
# Description

Handled hight contrast default behaviour in dot badge and focus ring

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-430


packages/components/src/components/badge/badge.styles.ts
```diff
@@ -65,6 +65,13 @@ const styles = [
       background-color: var(--mdc-badge-secondary-background-color);
       color: var(--mdc-badge-secondary-foreground-color);
     }
+
+    /* High Contrast Mode */
+    @media (forced-colors: active) {
+    .mdc-badge-dot, .mdc-badge-icon, .mdc-badge-text {
```
cscheffauer, line 71: q: why putting the outline on all of them? Isn't this looking weird when all of the child elements also have an outline? I haven't tested it, but just from this line it feels of to me. Like i would expect that the outline is on the badge itself (host)

packages/components/src/components/badge/badge.styles.ts
```diff
@@ -65,6 +65,13 @@ const styles = [
...
+      outline: 0.125rem solid;
```
cscheffauer, line 72: q: does this work correctly? Like normal you would need "value solid color", in your case the outline has no color

2020Deeya: Yes, it works. The reason color is not provided here is, it doesn't matter if the color is present or not, a default color is always going to be set by the browser.

supminn (approved):

PR #971, opened 2024-11-26
test: create Visual Regression Snapshots in High Contrast Mode

2020Deeya:
# Description

Additionally to Right to Left language support, added additional screenshot has been added  for High Contrast Mode.

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-431


packages/components/config/playwright/setup/utils/visual-regression.ts
```diff
@@ -14,6 +14,28 @@ class VisualRegression {
...
+    await this.page.evaluate((dir) => {
+      document.documentElement.setAttribute('dir', dir);
+    }, direction);
+  }
+
+   /**
+   * Enables or disables high contrast mode by emulating the `forcedColors` setting in the browser.
+   * @param enable - Boolean flag to enable or disable high contrast mode.
+   */
+  private async toggleHighContrastMode(enable: boolean): Promise<void> {
```
2020Deeya, line 31: The forcedColors emulation currently works only in Chromium-based browsers like Chrome and Edge and doesn’t apply to Firefox, Safari, or WebKit-based browsers, because those browsers do not yet fully support this emulation feature in Playwright or the forced-colors media query so we need an alternative solution.

packages/components/config/playwright/setup/utils/visual-regression.ts
```diff
@@ -24,27 +42,30 @@ class VisualRegression {
...
+        await this.setDocumentDirection(direction);
+        expect(await elementToTakeScreenShotFrom.screenshot(options)).toMatchSnapshot({
+          name: `${name}-${direction}.${CONSTANTS.VISUAL_REGRESSION.FILE_EXTENSION}`,
+          threshold: CONSTANTS.VISUAL_REGRESSION.THRESHOLD,
+          maxDiffPixelRatio: CONSTANTS.VISUAL_REGRESSION.MAX_DIFF_PIXELS_RATIO,
+        });
+      }
+
+      // High contrast screenshot only for LTR and supported browsers
+      if (['chromium', 'msedge'].includes(browserName)) {
```
supminn, line 59: question: Unless https://github.com/microsoft/playwright/issues/33765 has been resolved, we'd be testing only for chromium and edge is it?

2020Deeya: Yes, Since there is a limitation for other browsers, we would stick to Chrome and edge for now.

Jason20020 (approved):

PR #972, opened 2024-11-28
chore: avatar component and storybook

arjunkesava:
# Description

- Build Avatar component based on Tech breakdown guidelines.
- This PR covers the Component and Storybook part of Avatar.

## Screenshots
image


packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,431 @@
...
+    return html`
+      ${this.getPhotoPlaceHolderContent(type)}
+      ${this.getTemplateBasedOnType(type)}
+      ${this.getPresenceTemplateBasedOnType(type)}
+      ${this.getLoadingContent()}
+    `;
+  }
+
+  public override render(): TemplateResult {
+    const dimensions = this.generateAvatarDimensions();
```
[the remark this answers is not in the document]
arjunkesava: Removed, moved this to CSS.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,431 @@
...
+   * If the photo is not yet loaded, and the avatar type is PHOTO with initials provided,
+   * it generates and returns the initials as a placeholder text.
+   * If the photo is already loaded, it returns an empty template.
+   *
+   * @param type - The type of the avatar.
+   * @returns The template result containing the placeholder content or an empty template.
+   */
+  private getPhotoPlaceHolderContent(type: AvatarType): TemplateResult | typeof nothing {
+    // if photo is already loaded then no need to show placeholder
+    if (this.isPhotoLoaded) {
```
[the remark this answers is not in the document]
arjunkesava: I would prefer to have this separate, as this is more readable.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,431 @@
...
+      case AVATAR_TYPE.COUNTER:
+        return this.generateTextContent(type);
+      case AVATAR_TYPE.ICON:
+      default:
+        return this.iconTemplate();
+    }
+  }
+
+  private generateAvatarDimensions(): DirectiveResult {
+    const rems = [getAvatarSize(this.size), 'rem'].join('');
```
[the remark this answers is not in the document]
arjunkesava: done. moved it to css.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,431 @@
...
+   * to `false` and throws an error.
+   *
+   * @returns The template result containing the avatar photo.
+   */
+  private photoTemplate(): TemplateResult {
     return html`
       <img
+        class="photo"
         src="${ifDefined(this.src)}"
         alt="${ifDefined(this.alt)}"
```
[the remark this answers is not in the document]
arjunkesava: done. removed it.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,431 @@
...
-    }
+  /**
+   * @internal
+   * Handles errors that occur during the image loading process.
+   * Sets `isPhotoLoaded` to `false` to indicate the failure and throws an error message.
+   * The error message suggests checking the `src` attribute.
+   */
+  private handleOnError(): void {
+    this.isPhotoLoaded = false;
+    throw new Error('There was a problem while fetching the <img/>. Please check the src attribute and try again.');
```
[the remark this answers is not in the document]
arjunkesava: done, replaced it with `console.error` and added eslint disable for this line.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,431 @@
...
+
+  /**
+   * The avatar presence will be hidden if the user is typing or if the avatar type is counter.
+   * If the presence is set, it will be rendered as a child of the avatar.
+   *
+   * @param type - The type of the avatar.
+   * @returns The presence template or an empty template.
+   */
+  private getPresenceTemplateBasedOnType(type: AvatarType): TemplateResult | typeof nothing {
+    // while typing the loading spinner will be displayed and presence will be hidden
```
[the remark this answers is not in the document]
arjunkesava: Removed this line, from now, presence will be shown when typing.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,431 @@
...
+    if (!this.isTyping) {
+      return nothing;
     }
-    return html`${content}`;
+    const loadStyle = styleMap({ transform: getAvatarLoadingScaleSize(this.size) });
+    return html`
+      <div class="loading__container" aria-hidden="true" style="${loadStyle}">
+        <div class="loading__wrapper">
+          <!-- Load 3 small filled icons -->
+          ${repeat(Array(3), () => html`
```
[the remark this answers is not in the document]
arjunkesava: I've tried the mentioned css dots link above, the problem which I've faced is with the scaling.
the dots are not scalable for different size of avatars. Let me take another look at them.

packages/components/src/components/avatar/avatar.component.ts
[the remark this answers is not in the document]
cscheffauer: suggestion: as based on the latest convention, i think this one needs to be updated to have the this.onError handling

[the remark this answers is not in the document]
arjunkesava: done. via this commit: https://github.com/momentum-design/momentum-design/pull/972/commits/33da1baad45e350853e2b9328237ebb6b764ced2

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,431 @@
...
+   * Determines whether the avatar is clickable or not.
+   * @default false
    */
-  private updateSize() {
-    if (this.size) {
-      const value = `${this.size}${LENGTH_UNIT}`;
-      this.style.width = value;
-      this.style.height = value;
+  @property({ type: Boolean, attribute: 'is-clickable' })
+  isClickable = false;
```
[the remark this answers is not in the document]
arjunkesava: Done. removed it via: https://github.com/momentum-design/momentum-design/pull/972/commits/3394d712de2aa36c63bb6af74371664405d83fc3

packages/components/src/components/avatar/avatar.utils.ts
```diff
@@ -0,0 +1,68 @@
...
+    case AVATAR_SIZE.SMALL:
+      return FONT_TYPE.HEADING_SMALL_MEDIUM;
+    case AVATAR_SIZE.XX_SMALL: return FONT_TYPE.BODY_SMALL_MEDIUM;
+    case AVATAR_SIZE.X_SMALL:
+    default:
+      return FONT_TYPE.BODY_MIDSIZE_MEDIUM;
+  }
+};
+
+const getAvatarLoadingScaleSize = (size: AvatarSize): string => {
```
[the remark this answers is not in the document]
arjunkesava: Done. via https://github.com/momentum-design/momentum-design/pull/972/commits/6279a48657a2511ba0eae9f303952790e48af725

packages/components/src/components/avatar/avatar.component.ts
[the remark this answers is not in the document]
arjunkesava: Done via https://github.com/momentum-design/momentum-design/pull/972/commits/6279a48657a2511ba0eae9f303952790e48af725

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,385 @@
...
+   * Generates the counter text by converting the given number to a string.
+   * If the counter exceeds the maximum limit of 99, it will return the maximum limit as a string
+   * followed by a '+' character.
+   *
+   * @param counter - the number to be converted to a string
+   * @returns the counter text
+   */
+  private generateCounterText(counter: number): string {
+    if (counter > MAX_COUNTER) {
+      return [MAX_COUNTER, '+'].join('');
```
supminn, line 239: nitpick: why not `${MAX_COUNTER}+` instead?

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,385 @@
...
+      case AVATAR_TYPE.COUNTER:
+        return this.generateTextContent(type);
+      case AVATAR_TYPE.ICON:
+      default:
+        return this.iconTemplate();
+    }
+  }
+
+  /**
+   * Generates the loading spinner content for the avatar when typing.
```
supminn, line 320: suggestion: can we use the term 'loading indicator' rather than 'spinner'. As we show 3 dots, it doesn't justify to be spinner.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,385 @@
...
+   */
+  private getPhotoPlaceHolderContent(type: AvatarType): TemplateResult | typeof nothing {
+    // if photo is already loaded then no need to show placeholder
+    if (this.isPhotoLoaded) {
+      return nothing;
+    }
+    if (this.initials && type === AVATAR_TYPE.PHOTO) {
+      return this.textTemplate(this.generateInitialsText(this.initials));
+    }
+    return nothing;
```
supminn, line 350: question: should we render nothing if photo is still loading and initials are not provided? 
suggestion: Would it be good to show the 'user-regular' icon instead? (as it is the default icon)

packages/components/src/components/avatar/avatar.stories.ts
```diff
@@ -1,29 +1,114 @@
...
+        arg: 'counter',
+        exists: false,
+      },
+      control: 'select',
+      options: Object.values(PRESENCE_TYPE),
+    },
+    size: {
+      control: 'select',
+      options: Object.values(PRESENCE_SIZE),
+    },
```
supminn, line 42: issue: we need to define all the other args within `meta.argTypes`. Currently it is missing `src`, `counter`, `initial`, `icon-name`.
If we do not add them, any changes made to these args on storybook does not reflect on the component example.

packages/components/src/components/avatar/avatar.stories.ts
```diff
@@ -1,29 +1,114 @@
...
+    'icon-name': 'placeholder-bold',
+    'aria-label': 'An Avatar with Photo',
+  },
+};
+
+export const Default: StoryObj = {
+  args: {
+    size: 'x_small',
+  },
+};
```
supminn, line 62: question: if we are referring to this as the default avatar, I'd suggest that we use the same for fallback when the image is loading and initials are not provided 
default > initials > image

supminn (changes requested):

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,385 @@
...
- * @slot - This is a default/unnamed slot
+ * The `mdc-avatar` component is used to represent a person or a space.
+ * An avatar can be an icon, initials and photo.
  *
- * @summary This is MyElement
+ * To set the photo of an avatar,
+ * you need to set "src" attribute.
+ *
+ * While the avatar image is loading, as a placeholder,
+ * we can show the initials text.
```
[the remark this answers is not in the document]
arjunkesava: Initials are not mandatory. The default will be a `user-regular` icon.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,385 @@
...
+
+  /**
+   * Aria-label attribute to be set for accessibility
+   * @default null
+   */
+  @property({ type: String, attribute: 'aria-label' })
+  override ariaLabel: string | null = null;
+
+  /**
+   * The avatar presence will be hidden if the user is typing or if the avatar type is counter.
```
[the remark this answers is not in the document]
arjunkesava: The avatar presence will be hidden for counter type only.
For all other type of attributes, presence will be shown.

packages/components/src/components/avatar/avatar.component.ts
arjunkesava: Both are valid, it's just that I prefer this approach. That's it.

cscheffauer: issue: doing string concatenation as you mentioned Arjun is actually less performant than using ES6 Literals:
https://stackoverflow.com/a/29083467
You are correct in general that both are valid, but join is slower than ES6 Literals, so please change this to Supriya's suggestion and also use this moving forward.
This is a JS best practice, i think we don't need a convention for that.

packages/components/src/components/avatar/avatar.component.ts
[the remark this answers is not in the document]
supminn: sure. Could you please update the documentation. it says "The avatar presence will be hidden if the user is typing or if the avatar type is counter." currently.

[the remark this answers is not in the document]
arjunkesava: Thanks for noticing it. I've updated it via https://github.com/momentum-design/momentum-design/pull/972/commits/70c0f69bee67b82856b6077fec58e87770f8c683

packages/components/src/components/avatar/avatar.component.ts
arjunkesava: Done. I've updated it to template strings via https://github.com/momentum-design/momentum-design/pull/972/commits/70c0f69bee67b82856b6077fec58e87770f8c683

packages/components/src/components/avatar/avatar.stories.ts
```diff
@@ -1,29 +1,133 @@
...
-    type: DEFAULTS.TYPE,
+  },
+};
+
+export const Initials: StoryObj = {
+  args: {
+    initials: 'MD',
+    size: 'xx_large',
+  },
+};
```
supminn, line 88:
```suggestion
export const Initials: StoryObj = {
 argTypes: {
    ...disableControls(['counter', 'icon-name']),
  },
  args: {
    initials: 'MD',
    size: 'xx_large',
  },
};
```

suggestion: can we restrict the attributes for specific avatar type examples. When I am testing them individual, it is causing confusion.
For example: On the 'Initial' story, we have an option to set counter. However, if I set the counter value, the avatar does not change to a counter type (it still remains as an text avatar).

supminn: The same restriction can be applied conditionally for each story component

supminn (changes requested):

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,390 @@
...
+      ${this.getPresenceTemplateBasedOnType(type)}
+      ${this.getLoadingContent()}
+    `;
+  }
+
+  public override render(): TemplateResult {
+    const renderedContent = this.renderedContent();
+    const typingBackgroundStyle = styleMap({ opacity: this.isTyping ? '.7' : '1' });
+
+    if (this.onclick) {
```
supminn, line 372: issue: While the onclick handler functions correctly when the button is clicked with a mouse, keyboard interactions using keydown and keyup events are not being recognized.

For instance, I expect the onclick handler defined on the avatar to be triggered both when the avatar is clicked with a mouse and when it is navigated to via keyboard. Currently, the keyboard interactions do not seem to be functioning as intended.

supminn (changes requested):

packages/components/src/components/avatar/avatar.stories.ts
```diff
@@ -1,29 +1,168 @@
...
+          ?is-typing="${args['is-typing']}"
+          size="${size}"
+        ></mdc-avatar>
+      `)}
+    </div>
+  `,
+  args: {
+    src: 'https://picsum.photos/id/63/256',
+    presence: 'active',
+    'is-typing': false,
```
cscheffauer, line 166: nitpick: is typing is by default false, so no need to set it here

packages/components/src/components/avatar/avatar.stories.ts
```diff
@@ -1,29 +1,168 @@
...
+    },
+    ...classArgType,
+    ...styleArgType,
+  },
 };
 
 export default meta;
 
-export const Primary: StoryObj = {
+export const Image: StoryObj = {
```
cscheffauer, line 66: if u create a "Example" story as mentioned and I as a user can set the src on my own, this story is not needed anymore.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,390 @@
...
    */
-  private updateSize() {
-    if (this.size) {
-      const value = `${this.size}${LENGTH_UNIT}`;
-      this.style.width = value;
-      this.style.height = value;
+  @state() private isPhotoLoaded = false;
+
+  /**
+   * Aria-label attribute to be set for accessibility
```
cscheffauer, line 122: issue: this aria-label will only be set for the Button (onclick is set) - we have to mention this here, otherwise consumers might be confused.

packages/components/src/components/avatar/avatar.styles.ts
```diff
@@ -2,19 +2,100 @@ import { css } from 'lit';
 
 const styles = css`
   :host {
+    --mdc-avatar-default-background-color: var(--mds-color-theme-avatar-default);
```
cscheffauer, line 5: issue: the avatar is currently 612px wide (like the page) - we have to make sure that the element itself is just as big as the content. For that we have a utility called `hostFitContentStyles` - see Button component as an example.
image

packages/components/src/components/avatar/avatar.styles.ts
```diff
@@ -2,19 +2,100 @@ import { css } from 'lit';
...
+  }
+  .container {
+    padding: unset;
+    margin: unset;
+    border-radius: 10%;
+    background-color: unset;
+    color: var(--mdc-avatar-default-foreground-color);
+  }
+  .content {
+    cursor: pointer;
```
cscheffauer, line 63: issue: this makes the whole Avatar to look like it can be clicked, but it can't - if i do not pass in onclick it still shows like this, which will be a User Experience issue cause users are expecting that something happens, but it doesn't.

packages/components/src/components/avatar/avatar.styles.ts
```diff
@@ -2,19 +2,100 @@ import { css } from 'lit';
...
+  .photo {
+    border-radius: 50%;
+    height: 100%;
+    width: 100%;
     object-fit: cover;
   }
+  .icon {
+    height: 100%;
+    width: 100%;
+    color: white;
```
cscheffauer, line 79: suggestion: this is not correct and doesn't match the Figma spec. We always have to use our Theme tokens.
Use the same token for the loading animation then as well if possible.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,390 @@
...
+      return AVATAR_TYPE.ICON;
+    }
+    if (this.counter) {
+      return AVATAR_TYPE.COUNTER;
+    }
+    return AVATAR_TYPE.ICON;
+  }
+
+  /**
+   * @internal
```
cscheffauer, line 300: nitpick: this method has `@internal` and the other private methods don't.
We want to be consistent, so please adjust.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,390 @@
...
+      ${this.getPhotoPlaceHolderContent(type)}
+      ${this.getTemplateBasedOnType(type)}
+      ${this.getPresenceTemplateBasedOnType(type)}
+      ${this.getLoadingContent()}
+    `;
+  }
+
+  public override render(): TemplateResult {
+    const renderedContent = this.renderedContent();
+    const typingBackgroundStyle = styleMap({ opacity: this.isTyping ? '.7' : '1' });
```
cscheffauer, line 370: suggestion: as mentioned already, we should not need javascript to style if its based on a attribute - please move this into the stylesheet.

packages/components/src/components/avatar/avatar.stories.ts
```diff
@@ -1,29 +1,168 @@
...
 import '.';
 import { html } from 'lit';
+import { TYPE as PRESENCE_TYPE, SIZE as PRESENCE_SIZE } from '../presence/presence.constants';
 import { DEFAULTS } from './avatar.constants';
+import { disableControls } from '../../../config/storybook/utils';
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
 
 const render = (args: Args) => html`
-  <mdc-avatar src="${args.src}" size="${args.size}" type="${args.type}"></mdc-avatar>
+  <mdc-avatar
```
cscheffauer, line 11: issue(a11y): if i put a aria-label onto a clickable button and use the screenreader, it will read `aria-label, aria-label, group`, which is ultimately wrong. This Avatar is not a group, but supposed to be a button.
We have to fix this.

packages/components/src/components/avatar/avatar.stories.ts
```diff
@@ -1,29 +1,168 @@
...
+    alt="${args.alt}"
+    aria-label="${args['aria-label']}"
+    counter="${args.counter}"
+    icon-name="${args['icon-name']}"
+    initials="${args.initials}"
+    presence="${args.presence}"
+    size="${args.size}"
+    src="${args.src}"
+    ?is-typing="${args['is-typing']}"
```
cscheffauer, line 20: issue: i did pass in an `id` attribute and connected a label to this avatar (`<label for="avatar">test</label>`) - normally if a label is connected to a button and you click on the label, it should click the button. In this case it doesn't, since i think that the `label` can't find the button. I also tried passing the `id` to the mdc-button, its also not working - i think that has something to do with how shadow dom works, like it can't pierce through the shadow dom.
This is where u did the investigation on at the very beginning for the avatar - host-aria vs delegate aria. In this case i have a feeling we need to refactor the component to be host-aria, as suggested earlier.

cscheffauer (changes requested):

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,380 @@
...
+    return nothing;
+  }
+
+  /**
+   * @internal
+   * Assembles and returns the complete content for the avatar component.
+   *
+   * @returns The template result containing the fully assembled avatar content.
+   */
+  private renderedContent(): TemplateResult {
```
cscheffauer, line 360: suggestion: this function is not necessary, can be put into the render directly. this will avoid one function call and reduces the code lines in this file. also readability will not suffer, since its very good structured.

packages/components/src/components/avatar/avatar.stories.ts
cscheffauer: this is not addressed, now there are both Example and Image story - also make sure that the `Example` is always the first one - check this PR for the exact convention:
https://github.com/momentum-design/momentum-design/pull/987/files#diff-39335bf37c23b757ff30a6782f5c4f7d2aff9421c9847b6167252124194b396dR3

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,380 @@
...
+   * The photo is hidden until it is fully loaded, at which point
+   * `isPhotoLoaded` is set to `true` by the `handleOnLoad` method.
+   * If an error occurs during loading, `handleOnError` sets `isPhotoLoaded`
+   * to `false` and throws an error.
+   *
+   * @returns The template result containing the avatar photo.
+   */
+  private photoTemplate(): TemplateResult {
     return html`
       <img
```
cscheffauer, line 173: issue: currently when i do turn Chrome Network Emulation to "Slow 4G" and i paste the src and remove it again (see below), it removes the initials and then a while after shows the image. This will be a flicker for the user. We have to fix this.

https://github.com/user-attachments/assets/6d44912b-7889-487d-a651-3569b0b92252

cscheffauer: this also happens if there is a different fallback:

https://github.com/user-attachments/assets/237c6ed3-3790-4b8d-949c-619e49fb59aa

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,380 @@
...
-      content = html``;
+  /**
+   * @internal
+   * Generates the icon template for the avatar component.
+   * Utilizes the `mdc-icon` component to display an icon.
+   * If the `iconName` property is not provided, it defaults to the `DEFAULTS.ICON_NAME`.
+   *
+   * @returns The template result containing the avatar icon.
+   */
+  private iconTemplate(): TemplateResult {
```
cscheffauer, line 192: issue: this is not only an issue on the img, but also when the Icon Template is loaded or when we fall back to the "user-regular" icon. To fix this, we probably want a custom event in the Icon to indicate that its loaded (or a onload method similar to the img), which we can then hook into and have another internal state (same as img) which we can then use to only render the icon once its actually downloaded.

https://github.com/user-attachments/assets/d9db24ce-e9b6-4f39-a490-5de70ba1d9a2

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,380 @@
...
+
+  /**
+   * @internal
+   * Generates the text template for the avatar component.
+   * Utilizes the `mdc-text` component to display text.
+   *
+   * @param content - the text content to be displayed
+   * @returns The template result containing the avatar text.
+   */
+  private textTemplate(content: string): TemplateResult {
```
cscheffauer, line 211: suggestion: general note, we discussed that we are building the ButtonAvatar component after this component, but the code here is not prepared to allow for that. Like once u build the ButtonAvatar component my expectation is that u reuse the existing code from here (which is currently not possible). So we have to abstract our styling which we share into separate styles and also our templating which is happening here and probably logic.
For that i think we should look into making use of Mixins or Controllers (see lit.dev)

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,380 @@
...
 
   /**
-   * Scale of the avatar
+   * The counter is the number which can be displayed for the avatar.
+   * The maximum number is 99 and if the give number is greater than 99,
+   * then the avatar will be displayed as `99+`.
    */
   @property({ type: Number })
-  size?: number = DEFAULTS.SIZE;
+  counter?: number;
```
cscheffauer, line 104: issue: when passing in `0` (zero) into the counter, it doesn't render it:
image

cscheffauer (changes requested):

packages/components/src/components/avatar/avatar.utils.ts
```diff
@@ -0,0 +1,38 @@
...
+    case AVATAR_SIZE.LARGE: return FONT_TYPE.HEADING_MIDSIZE_MEDIUM;
+    case AVATAR_SIZE.MIDSIZE:
+    case AVATAR_SIZE.SMALL:
+      return FONT_TYPE.HEADING_SMALL_MEDIUM;
+    case AVATAR_SIZE.XX_SMALL: return FONT_TYPE.BODY_SMALL_MEDIUM;
+    case AVATAR_SIZE.X_SMALL:
+    default:
+      return FONT_TYPE.BODY_MIDSIZE_MEDIUM;
+  }
+};
```
2020Deeya, line 33: suggestion: The switch cases in both functions have repetitive logic. These can be replaced by a mapping object for better maintainability.

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,384 @@
...
+ * The avatar component is non clickable and non interactive/focusable component.
+ * If the avatar is typing, then the loading indicator will be displayed and presence will be hidden.
  *
- * @summary This is MyElement
+ * @dependency mdc-icon
+ * @dependency mdc-presence
+ * @dependency mdc-text
  *
  * @tagname mdc-avatar
  */
```
2020Deeya, line 35: question: I see, CSS variables have been used in styling, should we not mentioned them over here?

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,384 @@
...
+   * @internal
+   * The avatar presence will be hidden if the avatar type is COUNTER.
+   * If the presence is set, it will be rendered as a child of the avatar.
+   *
+   * @param type - The type of the avatar.
+   * @returns The presence template or an empty template.
+   */
+  private getPresenceTemplateBasedOnType(type: AvatarType): TemplateResult | typeof nothing {
+    // avatar type of counter should not have presence
+    if (type === AVATAR_TYPE.COUNTER && this.counter) {
```
2020Deeya, line 131: issue: Here, this.counter can also have 0 value where it'll return nothing, which is not expected. Logic should be refactored w.r.t using this.counter.

packages/components/src/components/avatar/avatar.component.ts
cscheffauer: this is not resolved

packages/components/src/components/avatar/avatar.component.ts
cscheffauer: agreed, presence gets rendered when 0 is set:
image

packages/components/src/components/avatar/avatar.component.ts
```diff
@@ -1,70 +1,384 @@
...
+    return nothing;
+  }
+
+  public override update(changedProperties: any): void {
+    super.update(changedProperties);
+
+    if (changedProperties.has('src')) {
+      // Reset the state of photo loaded, so that
+      // the avatar photo is displayed only when it is loaded.
+      this.isPhotoLoaded = false;
```
cscheffauer, line 368: issue: this is causing this behavior now:
When you change the src, it will go back to the initials, then jump to the new image. The expected behavior is that when the src changes that the existing src image stays until the new src is loaded.
Momentum React v2 does this as well, maybe check it there.

https://github.com/user-attachments/assets/07bfaa58-4396-40a0-aefa-7285d8898f82

cscheffauer (changes requested):

packages/components/src/components/avatar/avatar.component.ts
arjunkesava: This issue is fixed via https://github.com/momentum-design/momentum-design/pull/972/commits/9efd8986b729366e4beb474d582a6bf6514da29a,
We will now reset the photo loaded state when the src attribute is empty.

packages/components/src/components/avatar/avatar.component.ts
arjunkesava: Fixed.

packages/components/src/components/avatar/avatar.component.ts
arjunkesava: Fixed.

packages/components/src/components/avatar/avatar.utils.ts
arjunkesava: Fixed.

packages/components/src/components/avatar/avatar.component.ts
arjunkesava: created a new ticket to solve this issue: https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-454

packages/components/src/components/avatar/avatar.component.ts
2020Deeya: I still don't see CSS variables defined here in component file.

cscheffauer (approved):

supminn (commented):
documentation update required.

supminn (approved):

2020Deeya (approved):
LGTM

PR #973, opened 2024-11-29
refactor: updating stickersheet util

supminn:
# Description

We created a stickersheet util to capture combination of various attributes of an element on a single snapshot.
`generateComponentMarkup` has been moved out as a separate method within VisualRegression class. We need this method while creating a single snapshot for multiple button types (to minimize the no. of snapshots).


Jason20020 (dismissed):
LGTM

packages/components/config/playwright/setup/utils/visual-regression.ts
```diff
@@ -59,66 +61,76 @@ class VisualRegression {
...
+    componentMarkup: string,
+  ) {
     await componentsPage.mount({
       html: `
       <div class="componentWrapper">
-        ${generateComponentMarkup()}
+        ${componentMarkup}
       </div>
       `,
       clearDocument: true,
```
cscheffauer, line 136: can't add a comment at that line, but line 140 with the networkidle one should not be used, like we don't need to wait for that - instead i would recommend to actually wait for the componentList of line 139 - so 
`await componentList.waitFor();`

cscheffauer (approved):

PR #976, opened 2024-11-29
docs: Implement Marker component 

sandgupta23:
Implemented a new Marker Component.

# Description

- Added new Marker Component
- Added E2E Tests

## Links

[Confluence](https://confluence-eng-gpk2.cisco.com/conf/pages/viewpage.action?pageId=591008598)

[Figma](https://www.figma.com/design/9qhjTsC97Q3W8ZsbnTiZho/%F0%9F%A7%A9-%5BWIP%5D-Momentum-Web-Library?node-id=19902-33835&node-type=instance&m=dev)


packages/components/src/components/marker/marker.component.ts
```diff
@@ -0,0 +1,45 @@
+import { CSSResult } from 'lit';
+import styles from './marker.styles';
+import { Component } from '../../models';
+import { property } from 'lit/decorators.js';
+import type { MarkerVariant } from './marker.constants';
+import { MARKER_VARIANTS, MARKER_VARIANTS_CLASS_MAPPING} from './marker.constants';
+
+type MarkerChangedProperties = Map<keyof Marker, Marker[keyof Marker]>;
```
cscheffauer, line 8: suggestion: types should be moved into the types file (following our convention here: https://github.com/momentum-design/momentum-design/blob/main/packages/components/conventions/component-code-structure.md)

packages/components/src/components/marker/marker.component.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * `mdc-marker` which renders the default form which is a vertical line (4px wide)
+ * Marker Variants:
+ * - **Solid**: Solid marker.
+ * - **Striped**: Striped marker.
+ * 
+ * @tagname mdc-marker
+ *
+ * @slot default - This is a default/unnamed slot
```
cscheffauer, line 18: issue: this is not applicable (same as the cssprop below), please clean up!

packages/components/src/components/marker/marker.component.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @cssprop --custom-property-name - Description of the CSS custom property
+ */
+class Marker extends Component {
+  @property({ type: String }) variant: MarkerVariant = MARKER_VARIANTS.SOLID;
```
cscheffauer, line 23: issue: according to our conventions we should document every public property (adding TSDoc) - see here:
https://github.com/momentum-design/momentum-design/blob/main/packages/components/conventions/component-tsdoc.md#attributes--properties--methods-block

packages/components/src/components/marker/marker.component.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  override connectedCallback() {
+    super.connectedCallback();
+    this.classList.add(MARKER_VARIANTS_CLASS_MAPPING[this.variant]);
```
cscheffauer, line 27: issue: using javascript to change the styling here is an issue, like causes an performance issue. You can remove lines 25 to 41 completely and do the styling in the css file directly.

packages/components/src/components/marker/marker.constants.ts
```diff
@@ -0,0 +1,23 @@
+import utils from '../../utils/tag-name';
+const TAG_NAME = utils.constructTagName('marker');
+
+const MARKER_VARIANTS = {
+    "SOLID": "SOLID",
+    "STRIPED": "STRIPED",
+} as const;
+
+type MarkerVariant = keyof typeof MARKER_VARIANTS;
```
cscheffauer, line 9: suggestion: we should not mix constants and types, put the types in the types file and the constants in here.

packages/components/src/components/marker/marker.constants.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const MARKER_CLASSES = {
+    "SOLID": "mdc-marker--solid",
```
cscheffauer, line 12: issue: your local setup is not correct, like i see a ton of eslint & typescript issues in all your files. Please make sure to install the right VS Code extensions (ESLint) - will create a section in our guides for that.

packages/components/src/components/marker/marker.constants.ts
```diff
@@ -0,0 +1,23 @@
...
+    "STRIPED": "mdc-marker--striped",
+} as const;
+
+const MARKER_VARIANTS_CLASS_MAPPING = {
```
cscheffauer, line 16: suggestion: this is not necessary if u follow the CSS styling approach (see other comments)

packages/components/src/components/marker/marker.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+  },
+  argTypes: {
+    ...classArgType,
+    ...styleArgType
+  },
+};
+
+export default meta;
+
+export const solid: StoryObj = {
```
cscheffauer, line 29: suggestion: those Story names should be Uppercase, see other components.

packages/components/src/components/marker/marker.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+  args: {
+    variant: MARKER_VARIANTS.SOLID,
+  },
+};
+
+export const striped: StoryObj = {
```
cscheffauer, line 35: i think we might not need multiple stories, but just have a single one "Example" and there we just put the variant as a dropdown (see other components on how to do this)

packages/components/src/components/marker/marker.styles.ts
```diff
@@ -0,0 +1,29 @@
+import { css, unsafeCSS } from 'lit';
+import { MARKER_CLASSES } from './marker.constants';
+
+const SOLID_CLASS_NAME_BYPASS = unsafeCSS(MARKER_CLASSES.SOLID);
```
cscheffauer, line 4: issue: this is not necessary, please check other components like this one:
https://github.com/momentum-design/momentum-design/blob/main/packages/components/src/components/button/button.styles.ts#L52
you can directly do `:host([variant="striped"]){` - no need to create separate classes or such.

packages/components/src/components/marker/marker.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+const STRIPED_CLASS_NAME_BYPASS = unsafeCSS(MARKER_CLASSES.STRIPED);
+
+const styles = css`
+  :host {
+    width: 0.25rem;
+    height: 100%;
+    display: inline-block;
```
cscheffauer, line 11: question: is inline-block necessary or can it also work with a normal block?

packages/components/src/components/marker/marker.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  :host(.${SOLID_CLASS_NAME_BYPASS}) {
+    background: var(--mds-color-theme-outline-secondary-normal, rgba(0, 0, 0, 0.20));
```
cscheffauer, line 15: suggestion: i would recommend to abstract the color here out to a component css variable, see:
https://github.com/momentum-design/momentum-design/blob/main/packages/components/conventions/component-styles-variables.md
the same variable can then be used in line 21,22

cscheffauer (changes requested):

packages/components/src/components/marker/marker.component.ts
sandgupta23: Fixed.

packages/components/src/components/marker/marker.component.ts
sandgupta23: Fixed.

packages/components/src/components/marker/marker.component.ts
sandgupta23: Fixed.

packages/components/src/components/marker/marker.stories.ts
sandgupta23: Fixed.

packages/components/src/components/marker/marker.stories.ts
sandgupta23: Fixed

packages/components/src/components/marker/marker.styles.ts
sandgupta23: By default lit comp are display: inline. In order for height to work we need display: block/inline-block/flex etc.

packages/components/src/components/marker/marker.styles.ts
sandgupta23: fixed

packages/components/src/components/marker/marker.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  :host([variant='solid']) {
+    background: var(--mdc-marker-solid-background-color, rgba(0, 0, 0, 0.20));
+  }
+
+  :host([variant='striped']) {
+    background: repeating-linear-gradient(
+      135deg,
+      var(--mdc-marker-striped-color, rgba(0, 0, 0, 0.20)),
```
2020Deeya, line 21: Issue: We should not provide any fallback values because color can vary depending upon the curremt theme.

packages/components/src/components/marker/marker.e2e-test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const stickerSheet = new StickerSheet(componentsPage, 'mdc-marker');
+    await stickerSheet.mountComponents({ variants: MARKER_VARIANTS });
+    const container = stickerSheet.getWrapperContainer();
+    await test.step('matches screenshot of different variants', async () => {
+      await componentsPage.visualRegression.takeScreenshot('mdc-marker-variants', { element: container });
+    });
+  });
+};
+
+test.describe.parallel('mdc-Marker', () => {
```
cscheffauer, line 62: suggestion: this is not necessary at all, like there is only one test

packages/components/src/components/marker/marker.e2e-test.ts
```diff
@@ -0,0 +1,66 @@
...
+    await componentsPage.accessibility.checkForA11yViolations('marker-default');
+  });
+
+  /**
+   * VISUAL REGRESSION
+   */
+  await test.step('visual-regression', async () => {
+    const stickerSheet = new StickerSheet(componentsPage, 'mdc-marker');
+    await stickerSheet.mountComponents({ variants: MARKER_VARIANTS });
+    const container = stickerSheet.getWrapperContainer();
```
cscheffauer, line 55: issue: something is broken, cause all the screenshots you created are completely white, like nothing is shown.
Please verify the screenshots, not just create them without looking at them.

cscheffauer (changes requested):

packages/components/src/components/marker/marker.e2e-test.ts
supminn, line 55:
```suggestion
 await stickerSheet.mountComponents({ variant: MARKER_VARIANTS });
    const container = stickerSheet.getWrapperContainer();
    await container.evaluate((el) => {
      const wrapper = el.querySelector('.componentRowWrapper') as HTMLElement;
      wrapper.style.height = '100px';
    });
    
   These suggested changes will fix the snapshots. As the height of the marker depends on the container, it is necessary to set a fixed height for the wrapper container.
    
    
```

packages/components/src/components/marker/marker.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const styles = css`
+  :host {
+    width: 0.25rem;
+    height: 100%;
+    display: block;
+
+    --mdc-marker-solid-background-color: var(--mds-color-theme-outline-secondary-normal);
+    --mdc-marker-striped-color: var(--mds-color-theme-outline-secondary-normal);
+    --mdc-marker-striped-background-color: transparent;
```
2020Deeya, line 11: nitpick: We could keep CSS custom variables on top for better readability purpose.

packages/components/src/components/marker/marker.styles.ts
```diff
@@ -0,0 +1,29 @@
+import { css } from 'lit';
+
+const styles = css`
+  :host {
+    width: 0.25rem;
```
2020Deeya, line 5: issue: We should allow consumers to override the Marker's width if they wish to. To achieve that, we should create a CSS variable for Marker's width.

supminn: @2020Deeya Do we allow width to be modified? This is not mentioned on the tech breakdown.

cscheffauer (approved):

packages/components/src/components/marker/marker.styles.ts
2020Deeya: @supminn , Yes, Width can be overridden, it's mentioned in the tech breakdown.

cscheffauer (approved):

PR #978, opened 2024-12-03
docs: Implement Divider Component

2020Deeya:
# Description

Create Divider Component and Storybook docs

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-432
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-433


packages/components/src/components/divider/divider.component.ts
```diff
@@ -0,0 +1,262 @@
...
+   * @returns {void}
+   */
+  private setGrabberButton(): void {
+    const buttonElement = this.querySelector('mdc-button');
+    if (!buttonElement) return;
+
+    const iconType = this.getArrowIcon();
+    buttonElement.setAttribute('variant', 'secondary');
+    buttonElement.setAttribute('aria-label', this.ariaLabel || '');
+    buttonElement.setAttribute('aria-expanded', this.ariaExpanded || '');
```
supminn, line 146: issue: we should not set it to an empty string. The default value would automatically be taken when user hasn't set it right?

packages/components/src/components/divider/divider.component.ts
```diff
@@ -0,0 +1,262 @@
...
+  private setClassBasedOnType(type: string) {
+    this.classList.remove('mdc-primary-divider', 'mdc-text-divider', 'mdc-grabber-divider');
+    if (type === DIVIDER_TYPE_INTERNAL.PRIMARY) {
+      this.classList.add('mdc-primary-divider');
+    } else if (type === DIVIDER_TYPE_INTERNAL.TEXT) {
+      this.classList.add('mdc-text-divider');
+    } else if (type === DIVIDER_TYPE_INTERNAL.GRABBER_BUTTON) {
+      this.classList.add('mdc-grabber-divider');
+    }
+  }
```
supminn, line 223: issue: since we are setting class names on the component, it is treated similar to having classes on an HTML element. The consumers have an option to remove these classes. Thus, we shouldn't depend on class names for our component logic.
suggestion: We can make use of data attributes. Setting data attributes on the element itself can be helpful. We can further use this attributes for styling.

```suggestion
  private setClassBasedOnType(type: string) {
    if (type === DIVIDER_TYPE_INTERNAL.PRIMARY) {
      this.setAttribute('data-type','mdc-primary-divider');
    } else if (type === DIVIDER_TYPE_INTERNAL.TEXT) {
        this.setAttribute('data-type','mdc-text-divider');
    } else if (type === DIVIDER_TYPE_INTERNAL.GRABBER_BUTTON) {
      this.setAttribute('data-type','mdc-grabber-divider');
    }
  }
```

On the styles file, you can use them just like any other attribute
```
:host([data-type='mdc-text-divider']){
// styles
}
```

packages/components/src/components/divider/divider.component.ts
```diff
@@ -0,0 +1,262 @@
...
+   */
+  private inferDividerType() {
+    this.dividerTypeInternal = DIVIDER_TYPE_INTERNAL.PRIMARY;
+
+    const slot = this.shadowRoot?.querySelector('slot');
+    const assignedElements = slot?.assignedElements({ flatten: true }) || [];
+    if (assignedElements.length > 1) return;
+
+    const hasTextChild = assignedElements.some((el) => el.tagName === 'MDC-TEXT');
+    const hasButtonChild = assignedElements.some((el) => el.tagName === 'MDC-BUTTON');
```
supminn, line 237: suggestion: we can make use of the `TAGNAME` constants from button and text instead of hard coding these values.

packages/components/src/components/divider/divider.component.ts
```diff
@@ -0,0 +1,262 @@
...
+      this.setTextDivider();
+    } else if (!hasTextChild && hasButtonChild) {
+      this.dividerTypeInternal = DIVIDER_TYPE_INTERNAL.GRABBER_BUTTON;
+      this.setGrabberButton();
+    }
+  }
+
+  protected override render() {
+    this.setOrientation(this.orientation);
+    this.setVariant(this.variant);
```
supminn, line 250: question: why do we need to invoke them here?

supminn (changes requested):

packages/components/src/components/divider/divider.component.ts
```diff
@@ -0,0 +1,236 @@
...
+import { DEFAULTS, DIRECTIONS, DIVIDER_ORIENTATION, DIVIDER_TYPE_INTERNAL, DIVIDER_VARIANT } from './divider.constants';
+import { DividerOrientation, DividerTypeInternal, DividerVariant } from './divider.types';
+
+/**
+ * `mdc-divider` is basically a line that can also have a button/text positioned centrally.
+ * It is used to separate and organize the content and allow users to interact with the layout.
+ *
+ * Divider Orientation:
+ * - **Horizontal**: A thin, horizontal line with 0.063rem width.
+ * - **Vertical**: A thin, vertical line with 0.063rem width.
```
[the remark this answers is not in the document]
2020Deeya: ya, it should be 0.0625rem. Correcting it.
I think, it should be mentioned over here as well.

packages/components/src/components/divider/divider.component.ts
2020Deeya: Sure, we can. I left it to get suggestions from everyone if it's actually required.

packages/components/src/components/divider/divider.component.ts
2020Deeya: Do you mean, the default value of button? 
If that's the case, we can remove the empty string.

packages/components/src/components/divider/divider.component.ts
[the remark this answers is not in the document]
supminn: It's mentioned on the docs above the property. So, i'm not sure if it adds value here.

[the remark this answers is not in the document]
2020Deeya: Fair enough.

packages/components/src/components/divider/divider.stories.ts
```diff
@@ -0,0 +1,153 @@
...
+    ...meta.argTypes,
+    typeOfChildren: {
+      control: 'radio',
+      options: ['text'],
+      description: 'Text type is selected',
+    },
+    orientation: {
+      control: 'radio',
+      options: [DIVIDER_ORIENTATION.HORIZONTAL],
+      description: 'Only horizontal orientation with 0.0625rem width is allowed',
```
supminn, line 128: suggestion:

>  Note: We do not support "Vertical Text Divider" as of now.

This note is shown on all the other stories except this one. I'd suggest you to add this note here as well.

packages/components/src/components/divider/divider.component.ts
```diff
@@ -0,0 +1,297 @@
...
+   * Sets the buttonPosition and arrowDirection attribute for the divider component.
+   * If the provided buttonPosition and arrowDirection are not included in the DIRECTIONS,
+   * it defaults to the value specified in DIRECTIONS based on the ORIENTATION.
+   *
+   * @param buttonPosition - The buttonPosition to set.
+   * @param arrowDirection - The arrowDirection to set.
+   */
+  private ensureValidDirections() {
+    /* eslint-disable max-len */
+    const defaultDirection = this.orientation === DIVIDER_ORIENTATION.HORIZONTAL ? DIRECTIONS.NEGATIVE : DIRECTIONS.POSITIVE;
```
supminn, line 174:
```suggestion
    const defaultDirection = this.orientation === DIVIDER_ORIENTATION.HORIZONTAL
      ? DIRECTIONS.NEGATIVE
      : DIRECTIONS.POSITIVE;
```
By doing this, we can avoid using eslint-disabled.

supminn (changes requested):

packages/components/src/components/divider/divider.component.ts
2020Deeya: Done

packages/components/src/components/divider/divider.component.ts
```diff
@@ -0,0 +1,298 @@
...
+  @state()
+  private observer: MutationObserver;
+
+  /** @internal */
+  @state()
+  private windowDirection: string = 'ltr';
+
+  constructor() {
+    super();
+    this.observer = new MutationObserver(() => this.handleWindowDirectionChange());
```
supminn, line 114: issue: this shouldn't be used. MutationObserver is an expensive call. We should avoid using this and make changes using css styling instead.

packages/components/src/components/divider/divider.component.ts
```diff
@@ -0,0 +1,298 @@
...
+   * - **positive**
+   * - **negative**
+   *
+   * Note: Positive and Negative directions are defined based on Cartesian plane.
+   * @default 'negative'
+   */
+  @property({ type: String, attribute: 'button-position', reflect: true })
+  buttonPosition: string = DEFAULTS.BUTTON_DIRECTION;
+
+  /** @internal */
```
supminn, line 100: suggestion: We can omit using this internal state as we are capturing `data-type` to infer the type of divider.

supminn (changes requested):

supminn (approved):
LGTM

packages/components/src/components/divider/divider.component.ts
2020Deeya: Fixed.

PR #980, opened 2024-12-04
test: e2e test cases for button component

supminn:
# Description
This PR covers the end to end test cases for button component along with accessibility checks as well as visual regression snapshot testing.

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-388


packages/components/src/components/button/button.e2e-test.ts
```diff
@@ -1,66 +1,516 @@
...
+
+    await test.step('matches screenshot of icon-button element', async () => {
+      await componentsPage.visualRegression.takeScreenshot(
+        'mdc-button-icon',
+        { element: buttonSheet.getWrapperContainer() },
+      );
+    });
+  });
+
+  test('mdc-button key pressed and focused events', async ({ componentsPage }) => {
```
cscheffauer, line 425: issue: in none of these tests the "onclick" method on the button is verified. We have to test that as well.
I would recommend to just do the same assertion in the already existing tests, lets not create another one.

packages/components/src/components/button/button.e2e-test.ts
```diff
@@ -1,66 +1,516 @@
...
+
+      await componentsPage.page.keyboard.press('Tab');
+      await expect(button).not.toBeFocused();
     });
+  });
 
-    await test.step('keyboard', async () => {
-      await test.step('component should fire callback x when pressing y', async () => {
-        // TODO: add test here
+  test('mdc-button click event', async ({ componentsPage }) => {
```
cscheffauer, line 455: suggestion: you are using `test` very often - what that does is that everytime you do `test`, it opens a new browser window - thats very resource heavy - i would recommend to combine tests together, use the test.step more often.
With that u don't need to the mount all the time, but just keep one button there, change the attributes of the button and assert again. With that the overall runtime of this test file should be reduced.

packages/components/src/components/button/button.styles.ts
```diff
@@ -146,6 +145,7 @@ const styles = [hostFitContentStyles, css`
   }
   :host([size="20"]){
     height: var(--mdc-button-height-size-20);
+    padding: 0.0625rem;
```
cscheffauer, line 148: q: was that missed? or why are we adding this now? this is in the actual component, just want to confirm.

packages/components/src/components/icon/icon.component.ts
```diff
@@ -141,8 +150,10 @@ class Icon extends Component {
     if (changedProperties.has('name')) {
       // fetch icon data if name changes:
       this.getIconData().catch((err) => {
+        if (err.name !== 'AbortError') {
         // eslint-disable-next-line no-console
-        console.error(err);
+          console.error('Fetch error:', err);
```
cscheffauer, line 155: suggestion: since you are making changes here, i would recommend to not console.error, but do the onerror callback, as mentioned in our newly created error handling convention.

packages/components/src/components/icon/icon.utils.ts
```diff
@@ -10,8 +10,13 @@
  * @returns The valid icon element
  * @throws Error if the response is not ok
  */
-const dynamicSVGImport = async (url: string, name: string, fileExtension: string): Promise<Element> => {
-  const response = await fetch(`${url}/${name}.${fileExtension}`);
+const dynamicSVGImport = async (
```
cscheffauer, line 13: i think it would be neat to document how the signal works (or at least link to the lit.dev page) in this TSDoc.

packages/components/src/components/icon/icon.component.ts
```diff
@@ -89,7 +96,9 @@ class Icon extends Component {
     if (this.iconProviderContext.value) {
       const { fileExtension, url } = this.iconProviderContext.value;
       if (url && fileExtension && this.name) {
-        const iconHtml = await dynamicSVGImport(url, this.name, fileExtension);
+        this.abortController.abort();
+        this.abortController = new AbortController();
```
cscheffauer, line 100: q: why do we need to create another instance of the abortController? i think this is not necessary and might also cause issues in terms of constantly initialising another controller.

packages/components/src/components/icon/icon.component.ts
```diff
@@ -81,6 +81,13 @@ class Icon extends Component {
 
   private readonly iconProviderContext = providerUtils.consume({ host: this, context: IconProvider.Context });
 
+  @state() private abortController: AbortController;
+
+  constructor() {
+    super();
+    this.abortController = new AbortController(); // Initialize AbortController
```
cscheffauer, line 88: praise: for adding this, this is very neat!

cscheffauer (changes requested):

packages/components/src/components/button/button.component.ts
```diff
@@ -411,6 +411,10 @@ class Button extends Component {
       this.typeInternal = BUTTON_TYPE_INTERNAL.PILL_WITH_ICON;
     } else if (!slot && (this.prefixIcon || this.postfixIcon)) {
       this.typeInternal = BUTTON_TYPE_INTERNAL.ICON;
+      this.setAttribute(
+        'aria-label',
+        this.getAttribute('aria-label') || this.prefixIcon || this.postfixIcon || BUTTON_TYPE_INTERNAL.ICON,
```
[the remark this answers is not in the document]
supminn: I agree. I had initially not added this attribute. However, the a11y test case failed due to its absence. Do I need to pass aria-label from my test case instance?

packages/components/src/components/button/button.e2e-test.ts
supminn: Thank you for sharing this valuable information. I'll make the necessary changes.

packages/components/src/components/button/button.styles.ts
supminn: yes, we seem to have missed it during the component development. I figured it out while having a discussion with Jason over the popover close button.

packages/components/src/components/icon/icon.component.ts
supminn: yes, this is necessary. If we reuse the same AbortController instance and call abort() after starting a new fetch, it will terminate the previous fetch as well as the new one.

packages/components/src/components/icon/icon.component.ts
supminn: Thank you!

packages/components/src/components/icon/icon.utils.ts
supminn: Sure, I've got this added. I shall share about this during our learning session too.

packages/components/src/components/icon/icon.component.ts
cscheffauer: just checked, it seems to be fine in terms of cleanup - i was thinking that its actually an issue, since we don't cleanup - it seems `.abort()` is somehow already the cleanup.

packages/components/src/components/button/button.component.ts
```diff
@@ -254,18 +253,6 @@ class Button extends Component {
     }
   }
 
-  /**
-   * Sets the class of 'icon' for icon buttons.
-   * @param type - The type of the button.
-   */
-  private setClassBasedOnType(type: string) {
```
2020Deeya, line 261: question: Could you explain what was the issue which made us to switch our class logic to setting attribute?

supminn:
Closing this PR has it has bloated the number of commits. Created https://github.com/momentum-design/momentum-design/pull/994 with necessary changes.

PR #983, opened 2024-12-06
fix: stickersheet created as a class

supminn:
# Description

- Converting the sticker sheet functionality to a separate class.
- It gives us the flexibility to set default attributes, provide the combination of attributes to iterate through and create a sticker sheet for the given component.


packages/components/config/playwright/setup/utils/Stickersheet.ts
```diff
@@ -0,0 +1,177 @@
...
+    this.attributes = attributes;
+  }
+
+  /**
+   * Adds a new component to the sheet with specified attributes.
+   */
+  private async addComponentToSheet() {
+    const childrenEl = this.children
+      ? `<${this.tagname} id='${this.tagname}-${this.rowId}'>${this.children}</${this.tagname}>`
+      : `<${this.tagname} id='${this.tagname}-${this.rowId}'></${this.tagname}>`;
```
supminn, line 60: this has to be done in the above mentioned way and not as the following. `children = ''` would still create a value within the slot, and we don't want that for icon button
```suggestion
    const childrenEl = `<${this.tagname} id='${this.tagname}-${this.rowId}'>${this.children || ''}</${this.tagname}>`;
```

packages/components/config/playwright/setup/utils/Stickersheet.ts
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Adds a new component to the sheet with specified attributes.
+   */
+  private async addComponentToSheet() {
+    const attributesString = this.getAttributesAsString();
+    const childrenEl = this.children
+      ? `<${this.tagname} id='${this.tagname}-${this.rowId}' ${attributesString}>${this.children}</${this.tagname}>`
+      : `<${this.tagname} id='${this.tagname}-${this.rowId}' ${attributesString}></${this.tagname}>`;
+
+    console.log('Turbo 🚀  ~ addComponentToSheet ~ childrenEl:', childrenEl);
```
cscheffauer, line 70: issue: leftover console log

cscheffauer (dismissed):

cscheffauer (approved):

PR #985, opened 2024-12-09
fix: added part to icon css

cscheffauer:
# Description

- Added part to the Icon CSS to be able to style it and assert better during testing.


Jason20020 (approved):
LGTM

PR #987, opened 2024-12-10
docs: refactored conventions & added new

cscheffauer:
# Description

- Refactored the existing conventions to be less (merged styling related into the component-styling.md)
- Added new conventions:
  - Example story per Component
  - Tagging & Releasing of Components
  - Dynamic Styling
  - CSS Animations
- Applied Example Story to all existing stories by modifying them.

This has no impact to any output, only internal changes.


packages/components/conventions/component-styling.md
```diff
@@ -0,0 +1,157 @@
...
+    .example {
+         font-size: 1.5rem; /* 1.5 * 16px = 24px */
+         padding: 1rem;    /* 1 * 16px = 16px */
+         margin: 2rem;     /* 2 * 16px = 32px */
+    }
+    ```
+
+3. **Avoid Using `px` directly**:
+    - We should refrain from using `px` unless absolutely necessary. Any instances where `px` might be required should be discussed prior to implementation.
+
```
supminn, line 57: question: don't we want to maintain the reference links?

supminn (approved):

packages/components/conventions/component-styling.md
cscheffauer: i removed them to make the document less big. I can add them to the bottom of the page again - in general i think it would be easy to google as well, since its not too much of a niche topic.

PR #988, opened 2024-12-11
style(popover): popover styling and focus trap mixin

Jason20020:
### Description

Create Popover component styling
Create Focus Trap for Popover


packages/components/src/components/popover/popover.styles.ts
```diff
@@ -0,0 +1,109 @@
+import { css } from 'lit';
+
+const styles = css`
```
cscheffauer, line 3: issue: it seems like that the arrow position is off slightly:
image

packages/components/src/components/popover/popover.stories.ts
```diff
@@ -0,0 +1,364 @@
...
+import { html, TemplateResult } from 'lit';
+import { POPOVER_PLACEMENT, DEFAULTS } from './popover.constants';
+import { COLOR, ROLE } from '../modalcontainer/modalcontainer.constants';
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
+
+const createPopover = (id: string, triggerID: string, args: Args, content: TemplateResult) => html`
+<mdc-popover
+  id="${id}"
+  triggerID="${triggerID}"
+  trigger="${args.trigger}"
```
cscheffauer, line 12: issue: i just tried changing the trigger in storybook to "mouseenter", it is still click

packages/components/src/components/popover/popover.component.ts
```diff
@@ -0,0 +1,767 @@
...
+  /**
+   * Determines whether the backdrop is .
+   * @default false
+   */
+  @property({ type: Boolean, reflect: true })
+  backdrop: boolean = DEFAULTS.BACKDROP;
+
+  /**
+   * Changes the placement of popover to keep it in view.
+   * @default false
```
cscheffauer, line 165: suggestion: this is actually very useful to always be set, lets set the default to true for flip.

packages/components/src/components/popover/popover.stories.ts
```diff
@@ -0,0 +1,364 @@
...
+  placement="${args.placement}"
+  delay="${args.delay}"
```
cscheffauer, line 14: issue: changing delay in storybook will not apply it in the canvas

packages/components/src/components/popover/popover.styles.ts
```diff
@@ -0,0 +1,109 @@
+import { css } from 'lit';
```
cscheffauer, line 1: issue: closebutton should have a transparent background:
image

packages/components/src/components/popover/popover.stories.ts
```diff
@@ -0,0 +1,364 @@
...
+  ?set-index="${args['set-index']}"
+  ?visible="${args.visible}"
+  .offset="${args.offset ?? 0}"
+  ?interactive=${args.interactive}
+  ?focus-trap=${args['focus-trap']}
```
cscheffauer, line 19: issue: if i set this to true and interactive to true, it does not trap the focus inside the popover

packages/components/src/components/popover/popover.component.ts
```diff
@@ -0,0 +1,767 @@
...
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'hide-on-blur' })
+  hideOnBlur: boolean = DEFAULTS.HIDE_ON_BLUR;
+
+  /**
+   * Hide on outside click of the popover.
+   * @default false
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'hide-on-outside-click' })
+  hideOnOutsideClick: boolean = DEFAULTS.HIDE_ON_CLICK_OUTSIDE;
```
cscheffauer, line 147: issue: setting this to `true` and also setting the `backdrop` to `true` does not work - like if the backdrop is true, hide on outside click does not work. Is this expected?

cscheffauer (changes requested):
first initial review

packages/components/src/components/popover/popover.component.ts
```diff
@@ -0,0 +1,380 @@
...
+  ariaLabelledBy: string | null = null;
+
+  @property({ type: String, attribute: 'aria-modal' })
+  override ariaModal: string | null = null;
+
+  @property({ type: String })
+  override role: string | null = null;
+
+  @property({ type: String, attribute: 'aria-describeby' })
+  ariaDecribeBy: string | null = null;
```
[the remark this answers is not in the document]
cscheffauer: this is not fixed - its `aria-describedby` (with a d). We want both the attribute and the property to be spelled correctly.

packages/components/src/components/popover/popover.styles.ts
cscheffauer: this has not been fixed

packages/components/src/components/popover/popover.component.ts
cscheffauer: this is not fixed

packages/components/src/components/popover/popover.component.ts
```diff
@@ -0,0 +1,800 @@
...
+      this.onShowPopover();
+    }, this.openDelay);
+    popoverStack.push(this);
+  };
+
+  /**
+   * Hides the popover.
+   */
+  public hidePopover = () => {
+    if (popoverStack.peek() === this) {
```
cscheffauer, line 578: issue: when i press escape, the backdrop is still there and does not get removed.

packages/components/src/components/popover/popover.component.ts
```diff
@@ -0,0 +1,800 @@
...
+   */
+  @property({ type: Boolean, reflect: true })
+  size: boolean = DEFAULTS.SIZE;
+
+  /**
+   * The z-index of the popover.
+   * @default 1000
+   */
+  @property({ type: Number, reflect: true, attribute: 'set-index' })
+  setIndex: number = 1000;
```
cscheffauer, line 197: issue: this is not a `z-index` - please name it it `z-index`, both in the attribute name and the property name

cscheffauer (changes requested):

cscheffauer (dismissed):
YES! awesome stuff! 💯

cscheffauer (approved):

PR #994, opened 2024-12-13
test: e2e test cases for button component

supminn:
# Description

This PR covers the end to end test cases for button component along with accessibility checks as well as visual regression snapshot testing.

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-388


packages/components/src/components/button/button.e2e-test.ts
```diff
@@ -1,66 +1,554 @@
...
+          ${restArgs.color ? `color="${restArgs.color}"` : ''}
+          ${restArgs.prefixIcon ? `prefix-icon="${restArgs.prefixIcon}"` : ''}
+          ${restArgs.postfixIcon ? `postfix-icon="${restArgs.postfixIcon}"` : ''}
+          ${restArgs.ariaLabel ? `aria-label="${restArgs.ariaLabel}"` : ''}
+        ></mdc-button>
+      ${restArgs.secondButtonForFocus ? '<mdc-button>Second Button</mdc-button></div>' : ''}
+        `,
+      clearDocument: true,
+    });
+  }
```
arjunkesava, line 64: [suggestion]: This can reduced like this
```suggestion

  const htmlButtonContent = `
    ${restArgs.secondButtonForFocus ? '<div id="wrapper">' : ''}
      <mdc-button
        ${restArgs.active ? 'active' : ''}
        ${restArgs.disabled ? 'disabled' : ''}
        ${restArgs.softDisabled ? 'soft-disabled' : ''}
        ${restArgs.variant ? `variant="${restArgs.variant}"` : ''}
        ${restArgs.size ? `size="${restArgs.size}"` : ''}
        ${restArgs.color ? `color="${restArgs.color}"` : ''}
        ${restArgs.prefixIcon ? `prefix-icon="${restArgs.prefixIcon}"` : ''}
        ${restArgs.postfixIcon ? `postfix-icon="${restArgs.postfixIcon}"` : ''}
        ${restArgs.ariaLabel ? `aria-label="${restArgs.ariaLabel}"` : ''}
      >
        ${restArgs.children ? restArgs.children : ''}
      </mdc-button>
    ${restArgs.secondButtonForFocus ? '<mdc-button>Second Button</mdc-button></div>' : ''}
  `;
  await componentsPage.mount({
    html: htmlButtonContent,
    clearDocument: true,
  });
```

packages/components/src/components/button/button.e2e-test.ts
```diff
@@ -1,66 +1,554 @@
...
+
+  const element = restArgs.secondButtonForFocus
+    ? componentsPage.page.locator('div#wrapper')
+    : componentsPage.page.locator('mdc-button');
+  await element.waitFor();
+
+  // always return the first button:
+  const allButtons = await componentsPage.page.locator('mdc-button').all();
+  return allButtons[0];
```
arjunkesava, line 73: When we written the first button, we can use `first` instead of `all`.
```suggestion
  const firstButton = await componentsPage.page.locator('mdc-button').first();
  return firstButton;
```

packages/components/src/components/button/button.e2e-test.ts
```diff
@@ -1,66 +1,554 @@
...
+      await componentsPage.removeAttribute(button, 'active');
+    });
 
-  // initial check for the button be visible on the screen:
-  await button.waitFor();
+    // Active Disabled button
+    await test.step(`attribute active and disabled should be present on ${buttonType} button`, async () => {
+      await componentsPage.setAttributes(button, {
+        disabled: '',
+        active: '',
```
arjunkesava, line 144: [question] If we check, disabled and active both, then do we need to check them individually on top (above this case).

packages/components/src/components/button/button.e2e-test.ts
```diff
@@ -1,66 +1,554 @@
...
     });
+
+    if (buttonType === 'icon') {
+      for (const size of Object.values(ICON_BUTTON_SIZES)) {
+        if (size !== ICON_BUTTON_SIZES[20]) {
+          await test.step(`attribute size="${size}" should be present on ${buttonType} button`, async () => {
+            await componentsPage.setAttributes(button, { size: `${size}`, 'aria-label': 'icon-button-20' });
+            await expect(button).toHaveAttribute('size', `${size}`);
+          });
+        }
```
arjunkesava, line 181: [suggestion]: instead of an if condition, which checks for each iteration, we can filter the size 20 before we run the loop.
```suggestion
      const iconSizesToTest = Object.values(ICON_BUTTON_SIZES).filter(size => size !== ICON_BUTTON_SIZES[20]);
      for (const size of iconSizesToTest) {
        await test.step(`attribute size="${size}" should be present on ${buttonType} button`, async () => {
          await componentsPage.setAttributes(button, { size: `${size}`, 'aria-label': 'icon-button-20' });
          await expect(button).toHaveAttribute('size', `${size}`);
        });
      }
```
Note: this is just a suggestion, feel free to ignore.

packages/components/src/components/button/button.e2e-test.ts
supminn: Initially, I had created it in this way. However, `${restArgs.children ? restArgs.children : ''}` adds an empty value between opening and closing tag of the button, which is considered in the slot value. Thus, I had to make it verbose.

We are determining the type of the button based on the children values that are passed into the slot. For icon buttons (rounded in shape) we shouldn't have any children/slot values.

packages/components/config/playwright/setup/utils/Stickersheet.ts
```diff
@@ -184,6 +186,92 @@ class StickerSheet {
...
+        this.setAttributes({ ...this.attributes, ...combination });
+        childrenEl += this.addComponentToSheetV2();
+      }
+    }
+    this.createComponentsMarkupHTML(childrenEl, true);
+  }
+
+  public async createMarkupWithCombination(combinations: Record<string, Record<string, any>>, createNewRow = false) {
+    if (!this.tagname) {
+      throw new Error('tagname is required');
```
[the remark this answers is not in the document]
supminn: Here, since its pure javascript that we deal with and we are using it only on test cases, it's better to throw an Error. As its not a part of the component or lit, it should be fine.

cscheffauer (approved):
lgtm

arjunkesava (approved):

PR #995, opened 2024-12-13
fix: conditionally create row wrapper

supminn:
# Description

This PR updates the stickersheet's mountComponent function to support an additional attribute to create a new wrapper div.


arjunkesava (dismissed):

packages/components/config/playwright/setup/utils/Stickersheet.ts
```diff
@@ -133,18 +133,23 @@ class StickerSheet {
    * Mounts components onto the page using specified combinations of attributes.
    * @param combinations - An object where keys are attribute names and
    * values are objects containing possible attribute configurations.
+   * @param createNewRow - A boolean (optional) indicating whether to create a new row for each combination.
+   * Its default value is false.
    * @throws Will throw an error if tagname is not defined.
    */
 
-  public async mountComponents(combinations: Record<string, Record<string, any>>) {
+  public async mountComponents(combinations: Record<string, Record<string, any>>, createNewRow: boolean = false) {
```
arjunkesava, line 141: Do we need to set type when we assign the value at declaration ?
```suggestion
  public async mountComponents(combinations: Record<string, Record<string, any>>, createNewRow = false) {
```

cscheffauer (dismissed):

arjunkesava (dismissed):

arjunkesava (approved):

PR #999, opened 2024-12-17
chore: build avatar button component

arjunkesava:
# Description

- Build an clickable avatar with button as a wrapper.
- This button avatar component uses the same attributes which are used in avatar component.
- All of the business logic will be the same for Avatar and this component.
- Create a new mixin: AvatarComponentMixin to inherit common props in between both Avatar and ButtonAvatar components.

image


packages/components/src/components/button/button.component.ts
```diff
@@ -20,6 +21,7 @@ import type {
 } from './button.types';
 import { getIconNameWithoutStyle, getIconSize } from './button.utils';
 import type { IconNames } from '../icon/icon.types';
+import { AvatarSize } from '../avatar/avatar.types';
```
cscheffauer, line 24: suggestion: follow our convention of making this a type only import

packages/components/src/components/buttonavatar/buttonavatar.component.ts
```diff
@@ -0,0 +1,70 @@
...
+import { AvatarSize } from '../avatar/avatar.types';
+import { DEFAULTS } from '../avatar/avatar.constants';
+
+/**
+ * The `mdc-buttonavatar` component is an interactable version of the `mdc-avatar` component.
+ *
+ * This component is made using the `mdc-button` component.
+ * The button component acts as a wrapper for the avatar component.
+ *
+ * @dependecy mdc-avatar
```
cscheffauer, line 16: typo

packages/components/src/components/buttonavatar/buttonavatar.component.ts
```diff
@@ -0,0 +1,70 @@
+import { CSSResult, html } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { property } from 'lit/decorators.js';
+import styles from './buttonavatar.styles';
+import Button from '../button/button.component';
+import { AvatarComponentMixin } from '../../utils/mixins/AvatarComponentMixin';
+import { AvatarSize } from '../avatar/avatar.types';
```
cscheffauer, line 7: suggestion: same as above, follow our convention of type only import

packages/components/src/components/buttonavatar/buttonavatar.component.ts
```diff
@@ -0,0 +1,70 @@
...
+   */
+  @property({ type: String, reflect: true })
+  override size: AvatarSize = DEFAULTS.SIZE;
+
+  constructor() {
+    super();
+    this.unsetButtonUnusedProps();
+  }
+
+  private unsetButtonUnusedProps() {
```
cscheffauer, line 41: suggestion: there is no reason to have this function, its only called in one place. Move the code in the constructor.

packages/components/src/components/buttonavatar/buttonavatar.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    this.color = undefined;
+    this.disabled = undefined;
+    this.postfixIcon = undefined;
+    this.prefixIcon = undefined;
+    this.softDisabled = undefined;
+    this.variant = undefined;
+  }
+
+  public override render() {
+    const { src, initials, size, presence, iconName, counter, isTyping } = this;
```
cscheffauer, line 52: suggestion: no need for this destructuring, use this.attributename directly in the render.

packages/components/src/components/buttonavatar/buttonavatar.stories.ts
```diff
@@ -0,0 +1,73 @@
...
+      control: 'boolean',
+    },
+    'icon-name': {
+      control: 'text',
+    },
+    counter: {
+      control: 'number',
+    },
+    ...disableControls([
+      'disabled',
```
cscheffauer, line 54: issue: those should not be in the storybook at all, use the hideControls utility

packages/components/src/components/buttonavatar/buttonavatar.styles.ts
```diff
@@ -0,0 +1,11 @@
+import { css } from 'lit';
+import { hostFitContentStyles, hostFocusRingStyles } from '../../utils/styles';
+
+const styles = [hostFitContentStyles, css`
+  :host {
```
cscheffauer, line 5: issue: border radius when focused doesn't match Figma - in Figma its 4px, in yours its 20px

packages/components/src/components/buttonavatar/buttonavatar.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  private unsetButtonUnusedProps() {
+    this.active = undefined;
+    this.color = undefined;
+    this.disabled = undefined;
+    this.postfixIcon = undefined;
+    this.prefixIcon = undefined;
+    this.softDisabled = undefined;
+    this.variant = undefined;
```
cscheffauer, line 48: issue: both `role` and `type` should be hardcoded, since we don't want consumers to override the ButtonAvatar role and type for now.

packages/components/src/components/buttonavatar/buttonavatar.stories.ts
```diff
@@ -0,0 +1,73 @@
...
+      'active',
+      'soft-disabled',
```
cscheffauer, line 56: issue: `role` & `type` are visible in storybook, should not be overriddable by the user - so we have to remove them.

packages/components/src/components/buttonavatar/buttonavatar.stories.ts
```diff
@@ -0,0 +1,73 @@
...
+    ]),
+    ...classArgType,
+    ...styleArgType,
+  },
+};
```
cscheffauer, line 61: issue: all of those button attributes are all visible in storybook still - we should remove all of them.
image

packages/components/src/components/buttonavatar/index.ts
```diff
@@ -0,0 +1,13 @@
+import ButtonAvatar from './buttonavatar.component';
+import { TAG_NAME } from './buttonavatar.constants';
+import '../avatar';
```
cscheffauer, line 3: issue: button needs to be registered here as well

packages/components/src/components/button/button.component.ts
```diff
@@ -271,15 +275,18 @@ class Button extends Component {
    *
    * @param size - The size to set.
    */
-  private setSize(size: PillButtonSize | IconButtonSize) {
+  private setSize(size: PillButtonSize | IconButtonSize | AvatarSize) {
+    if (Object.values(AVATAR_SIZES).includes(size as AvatarSize)) {
```
cscheffauer, line 279: issue: with this code we are adding logic from the avatar to the button, which is causing dependencies between them which we don't want!
solution would be: change the `setSize` method here from private to protected, keep the original implementation and then in the avatar itself do an override with super.setSize(); to call the rest.

packages/components/src/components/button/button.component.ts
```diff
@@ -10,6 +10,7 @@ import {
   ICON_BUTTON_SIZES,
   PILL_BUTTON_SIZES,
 } from './button.constants';
+import { SIZE as AVATAR_SIZES } from '../presence/presence.constants';
```
cscheffauer, line 13: issue: this doesn't need to be imported and used if the way the setSize is written will be changed - see other comment

packages/components/src/components/button/button.component.ts
```diff
@@ -109,13 +111,13 @@ class Button extends Component {
    * - Tertiary icon button cam also be 20.
    * @default 32
    */
-  @property({ type: Number }) size: PillButtonSize | IconButtonSize = DEFAULTS.SIZE;
+  @property({ type: Number }) size: PillButtonSize | IconButtonSize | AvatarSize = DEFAULTS.SIZE;
```
cscheffauer, line 114: issue: setting the AvatarSize here will be an issue for consumers. When consumers use the Button Component only with Typescript, they will see that you can also pass in a "AvatarSize", which is not correct. We will mislead the consumer in the way they can use the button component.
I would need to look closer at how to fix the issue of typing here - can do that separately or maybe u can have a look?

packages/components/src/components/button/button.component.ts
```diff
@@ -288,11 +295,13 @@ class Button extends Component {
    *
    * @param color - The color to set.
    */
-  private setColor(color: ButtonColor) {
-    if (!Object.values(BUTTON_COLORS).includes(color) || this.variant === BUTTON_VARIANTS.TERTIARY) {
-      this.setAttribute('color', `${DEFAULTS.COLOR}`);
-    } else {
-      this.setAttribute('color', color);
+  private setColor(color?: ButtonColor) {
+    if (color) {
```
cscheffauer, line 299: issue: this introduces an issue in the button now. If first the color is passed in and then the color changes to undefined, it will never be reset with the way this is written now. I know this has been done to fix Typescript issues - try to fix them directly in line 300 or so or just do `as string` etc.

packages/components/src/components/button/button.component.ts
```diff
@@ -260,8 +262,10 @@ class Button extends Component {
    *
    * @param variant - The variant to set.
    */
-  private setVariant(variant: ButtonVariant) {
-    this.setAttribute('variant', Object.values(BUTTON_VARIANTS).includes(variant) ? variant : DEFAULTS.VARIANT);
+  private setVariant(variant?: ButtonVariant) {
+    if (variant) {
```
cscheffauer, line 266: issue: same comment as below (color) - with this code the variant attribute never gets reset to the DEFAULTS.VARIANT if no variant is passed in, after a variant has been set before.

packages/components/src/components/buttonavatar/buttonavatar.component.ts
```diff
@@ -0,0 +1,70 @@
...
+        presence="${ifDefined(presence)}"
+        size="${ifDefined(size)}"
+        icon-name="${ifDefined(iconName)}"
+        counter="${ifDefined(counter)}"
+        ?is-typing="${isTyping}"
+      ></mdc-avatar>
+    `;
+  }
+
+  public static override styles: Array<CSSResult> = [...Button.styles, ...styles];
```
cscheffauer, line 67: issue: since all the button styles are applied, all the `--mdc-button-` CSS Properties are also now set. None of them are necessary:
image

My suggestion to fix it is to remove the Button.styles from here and just add the focus ring styles (same as in the button) - since i think the focus ring is the only style we need.

cscheffauer (changes requested):
first round of review, multiple issues

packages/components/src/utils/mixins/AvatarComponentMixin.ts
```diff
@@ -0,0 +1,83 @@
...
+
+type Constructor<T = {}> = new (...args: any[]) => T;
+
+export interface AvatarComponentMixinInterface {
+  src?: string;
+  initials?: string;
+  presence?: PresenceType;
+  iconName?: IconNames;
+  counter?: number;
+  isTyping: boolean;
```
arjunkesava, line 14: I've removed size from this interface.

packages/components/src/index.ts
```diff
@@ -20,6 +21,7 @@ export {
   Text,
   Button,
   Bullet,
+  AvatarButton,
```
arjunkesava, line 24: Renamed from 'ButtonAvatar' to 'AvatarButton'.

packages/components/src/components/avatarbutton/avatarbutton.component.ts
```diff
@@ -0,0 +1,79 @@
...
+  constructor() {
+    super();
+
+    this.active = undefined as unknown as boolean;
+    this.color = undefined as unknown as ButtonColor;
+    this.disabled = undefined as unknown as boolean;
+    this.postfixIcon = undefined;
+    this.prefixIcon = undefined;
+    this.softDisabled = undefined as unknown as boolean;
+    this.variant = undefined as unknown as ButtonVariant;
```
arjunkesava, line 48: I've removed the optional check of these properties in button component and set them to undefined using `unknown` type.

packages/components/src/components/button/button.component.ts
arjunkesava: I've renamed the size attribute to avatarSize internally via this commit: https://github.com/momentum-design/momentum-design/pull/999/commits/690da3fdbe01dc00afe86996e84ebe18ff7f32bd

packages/components/src/components/avatarbutton/avatarbutton.component.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    this.role = 'button';
+    this.type = BUTTON_DEFAULTS.TYPE;
```
cscheffauer, line 33: issue: when you just open storybook for the first time, it will set a lot of default values from the button:
image

also see that the variant and color are not undefined, like they are still there.

packages/components/src/components/buttonavatar/buttonavatar.styles.ts
cscheffauer: issue: it has now the different border radius, but the focus ring styling is broken cause of it:
image

this is how it should look like:
image

packages/components/src/components/button/button.component.ts
cscheffauer: this is solved just half way - you change the private to protected, but i can't see any overrides of the setSize method in the avatarbutton nor avatar. Is this logic now removed?
if we are not intending to override the setSize method, you can keep the function private.

packages/components/src/components/avatarbutton/avatarbutton.stories.ts
```diff
@@ -0,0 +1,81 @@
...
+      'disabled',
+      'soft-disabled',
+      'postfix-icon',
+      'prefix-icon',
+      'role',
+      'softDisabled',
+      'tabIndex',
+      'type',
+      'variant',
+    ]),
```
cscheffauer, line 65: there is now a difference between attribute and property:
image

i think we should actually rename the attribute as well, so in your component make sure the attribute and property are both "avatarSize" / "avatar-size"

cscheffauer (changes requested):

packages/components/src/utils/mixins/AvatarComponentMixin.ts
```diff
@@ -0,0 +1,101 @@
+import { property } from 'lit/decorators.js';
+import { DEFAULTS as AVATAR_DEFAULTS } from '../../components/avatar/avatar.constants';
+import type { AvatarSize } from '../../components/avatar/avatar.types';
+import type { IconNames } from '../../components/icon/icon.types';
+import type { PresenceType } from '../../components/presence/presence.types';
+import type { Component } from '../../models';
+
+type Constructor<T = {}> = new (...args: any[]) => T;
```
2020Deeya, line 8: Suggestion: Let's extract this 8th line from this mixin (and others including DisabledMixin and TabIndexMixin) and add it inside types.ts file.

packages/components/src/components/avatarbutton/avatarbutton.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  protected override setSize() {
+    // The avatar sizes and the button sizes are different.
+    // If the given size is a valid avatar size, then we will not call the parent setSize() method.
+    if (Object.values(AVATAR_SIZE).includes(this.avatarSize)) {
```
2020Deeya, line 40: issue: The setSize method checks if the size is a valid avatar size before deferring to the parent. However, there's a potential issue if this.avatarSize is undefined or null. This could lead to unintended behavior.

Updated one:
```
if (this.avatarSize && Object.values(AVATAR_SIZE).includes(this.avatarSize)) {
  return;
}
```

packages/components/src/components/avatarbutton/avatarbutton.component.ts
```diff
@@ -0,0 +1,70 @@
...
+import type { ButtonColor, ButtonVariant } from '../button/button.types';
+import { SIZE as AVATAR_SIZE } from '../presence/presence.constants';
+import styles from './avatarbutton.styles';
+
+/**
+ * The `mdc-avatarbutton` component is an interactable version of the `mdc-avatar` component.
+ *
+ * This component is made using the `mdc-button` component.
+ * The button component acts as a wrapper for the avatar component.
+ *
```
2020Deeya, line 15: Suggestion: Document attributes' fallback behavior. For instance: If presence or iconName are invalid or undefined is undefined, how does the avatarbutton component handle it?

supminn:
This PR will be marked closed after https://github.com/momentum-design/momentum-design/pull/1008 merges.
https://github.com/momentum-design/momentum-design/pull/1009 is the new PR reflecting the same changes.

arjunkesava:
Closing this PR as these changes were merged via https://github.com/momentum-design/momentum-design/pull/1009

PR #1005, opened 2024-12-20
fix: added input draft

cscheffauer:
# Description

_Please replace this line with a description of the changes made in this Pull Request._

## Links

_Please replace this line with relevant links to the changes made in this Pull Request._

### DIFF
--- a/packages/components/src/components/input/input.component.ts
+++ b/packages/components/src/components/input/input.component.ts
@@ -1,0 +1,102 @@
+import { CSSResult, html, nothing } from 'lit';
+import { property } from 'lit/decorators.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import styles from './input.styles';
+import { Component } from '../../models';
+
+/**
+ * input component, which ...
+ *
+ * @tagname mdc-input
+ *
+ * @slot default - This is a default/unnamed slot
+ *
+ * @cssprop --custom-property-name - Description of the CSS custom property
+ */
+class Input extends Component {
+  @property()
+  label = undefined;
+
+  /**
+   * @beta this attribute is in beta, and is subject to change (given the toggletip component is not ready yet)
+   */
+  @property()
+  labelInfoText = undefined;
+
+  @property()
+  helpText = undefined;
+
+  @property()
+  prefixText = undefined;
+
+  @property()
+  helpTextType?: 'error' | 'default' = 'default';
+
+  @property()
+  value = undefined;
+
+  protected renderLabel() {
+    if (!this.label) {
+      return nothing;
+    }
+    return html`<label for="${this.id}">${this.label}</label>`;
+  }
+
+  protected renderLabelInfoToggleTip() {
+    if (!this.labelInfoText) {
+      return nothing;
+    }
+    return html`<mdc-icon name=""></mdc-icon>`;
+  }
+
+  protected renderPrefixText() {
+    if (!this.prefixText) {
+      return nothing;
+    }
+    return html`<mdc-text>${this.prefixText}</mdc-text>`;
+  }
+
+  protected renderHelpTextIcon() {
+    if (!this.helpText) {
+      return nothing;
+    }
+    switch (this.helpTextType) {
+      case 'error':
+        return html`<mdc-icon name="error-regular"></mdc-icon>`;
+
+      default:
+        return nothing;
+    }
+  }
+
+  protected renderHelpText() {
+    if (!this.helpText) {
+      return nothing;
+    }
+    return html` <mdc-text>${this.helpText}</mdc-text> `;
+  }
+
+  public override render() {
+    return html`
+      <div class="input-header">
+        <slot name="label">${this.renderLabel()}</slot>
+        <slot name="label-icon">${this.renderLabelInfoToggleTip()}</slot>
+      </div>
+      <div class="input-container" part="input-container">
+        <slot name="input-prefix-text">${this.renderPrefixText()}</slot>
+        <slot name="input">
+          <input id="${this.id}" value="${ifDefined(this.value)}" />
+        </slot>
+        <slot name="trailing-button"></slot>
+      </div>
+      <div class="input-footer" part="input-footer">
+        <slot name="help-icon">${this.renderHelpTextIcon()}</slot>
+        <slot name="help-text">${this.renderHelpText()}</slot>
+      </div>
+    `;
+  }
+
+  public static override styles: Array<CSSResult> = [...Component.styles, ...styles];
+}
+
+export default Input;
--- a/packages/components/src/components/input/input.constants.ts
+++ b/packages/components/src/components/input/input.constants.ts
@@ -1,0 +1,17 @@
+import utils from '../../utils/tag-name';
+
+const TAG_NAME = utils.constructTagName('input');
+
+const VALIDATION = {
+  ERROR: 'error',
+  ERROR: 'error',
+  DEFAULT: 'default',
+};
+
+const DEFAULTS = {
+  NAME: undefined,
+  INFO_ICON_NAME: 'info-filled',
+  SIZE: 1,
+} as const;
+
+export { TAG_NAME, DEFAULTS, VALIDATION };
--- a/packages/components/src/components/input/input.stories.ts
+++ b/packages/components/src/components/input/input.stories.ts
@@ -1,0 +1,37 @@
+import type { Meta, StoryObj, Args } from '@storybook/web-components';
+import '.';
+import { html } from 'lit';
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
+
+const render = (args: Args) =>
+  html` <mdc-input
+    label="${args.label}"
+    labelInfoText="${args.labelInfoText}"
+    helpText="${args.helpText}"
+    validationType="${args.validationType}"
+    class="${args.class}"
+    style="${args.style}"
+  ></mdc-input>`;
+
+const meta: Meta = {
+  title: 'Work In Progress/input',
+  tags: ['autodocs'],
+  component: 'mdc-input',
+  render,
+  parameters: {
+    badges: ['wip'],
+  },
+  argTypes: {
+    ...classArgType,
+    ...styleArgType,
+  },
+};
+
+export default meta;
+
+export const Example: StoryObj = {
+  args: {
+    class: 'custom-classname',
+    style: 'margin-top: 20px;',
+  },
+};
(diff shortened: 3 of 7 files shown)


packages/components/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,193 @@
...
+  ></mdc-input>
+  <mdc-input 
+  id="7"
+  label="Label"
+  labelInfoText="Label info text"
+  helpText="Helper text"
+  helpTextType="default"
+  readonly placeholder="Placeholder"
+  value="This is readonly"
+  ></mdc-input>
```
2020Deeya, line 163: issue: It is a read-only text, but it is getting cleared. To avoid that, the clear button should not be present.

packages/components/src/components/input/input.styles.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+  :host([helpTextType="success"]) .input {
+    border-color: var(--mds-color-theme-text-success-normal);
+  }
+  :host([helpTextType="priority"]) .input {
+    border-color: var(--mds-color-theme-text-accent-normal);
+  }
+
+  .hidden{
+    opacity: 0;
```
2020Deeya, line 113: issue: Clear button is hidden, but not completely gone from the dom. To prevent interaction with hidden elements, consider adding `pointer-events: none`.

packages/components/src/components/input/input.styles.ts
```diff
@@ -0,0 +1,118 @@
...
+    color: var(--mds-color-theme-inverted-text-primary-normal);
+  }
+
+
+  .prefix-text{
+    color: var(--mds-color-theme-text-secondary-normal);
+  }
+
+  .input-container:hover{
+    background-color: var(--mds-color-theme-background-primary-hover);
```
2020Deeya, line 75: question: do we require transition property?

packages/components/src/components/input/input.styles.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  .input-container:active, .input-container:focus{
```
2020Deeya, line 78: question: do we require transition property?

packages/components/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,193 @@
...
+    prefixText: '',
+    leadingIcon: '',
+  },
+};
+
+export const AllVariants: StoryObj = {
+  argTypes: {
+    ...disableControls(['label', 'labelInfoText', 'helpText', 'required', 'placeholder', 'value', 'helpTextType']),
+  },
+  render: () => html`
```
2020Deeya, line 100: suggestion: Many inputs in the `AllVariants` story repeat the same props. Consider using args for shared values to reduce redundancy.

packages/components/src/components/input/input.constants.ts
```diff
@@ -0,0 +1,26 @@
...
+const DEFAULTS = {
+  NAME: undefined,
+  INFO_ICON_NAME: 'info-circle-filled',
+  SIZE: 1,
+  VALIDATION: VALIDATION.DEFAULT,
+} as const;
+
+const PREFIX_TEXT_OPTIONS = {
+  MAX_LENGTH: 10,
+  HELPERTEXT: 'Prefix text should not exceed 10 characters',
```
2020Deeya, line 22: nitpick: Replace `HELPERTEXT` with `HELPER_TEXT`

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * @dependency mdc-icon
+ * @dependency mdc-text
+ * @dependency mdc-button
+ */
+class Input extends LabelMixin(DisabledMixin(Component)) {
+  /**
+   * The unique id of the input field. It is used to link the input field with the label.
+   */
+  @property({ type: String }) override id = '';
```
2020Deeya, line 36: question: what happens if user doesn't provide any id? In that case, are we proving any default id?

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,222 @@
...
+    if (!this.helpText) {
+      return nothing;
+    }
+    return html` <mdc-text tagname='span' type='body-midsize-regular'>${this.helpText}</mdc-text> `;
+  }
+
+  private clearInputText() {
+    this.value = '';
+    // focus the input field after clearing the text
+    this.shadowRoot?.querySelector('input')?.focus();
```
2020Deeya, line 167: suggestion: Leverage @query decorator to directly reference the input element.

```
@query('input') private inputElement!: HTMLInputElement;

  private clearInputText() {
    this.value = '';
    this.inputElement?.focus();
  }
```

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * The type of help text. It can be 'default', 'error', 'warning', 'success', 'priority'.
+   */
+  @property({ reflect: true }) helpTextType: ValidationType = DEFAULTS.VALIDATION;
+
+  /**
+   * The help text that is displayed below the input field.
+   */
+  @property({ type: String, reflect: true }) helpText = '';
```
2020Deeya, line 77: question:  Do we really require reflect property here?

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,222 @@
...
+        <slot name="input">
+          <input 
+            class='input' 
+            id="${this.id}"
+            .value="${this.value}"
+            ?disabled="${this.disabled}"
+            ?readonly="${this.readonly}"
+            placeholder=${ifDefined(this.placeholder)}
+            aria-required=${this.required}
+            aria-invalid=${this.helpTextType === 'error'}
```
2020Deeya, line 203: issue/question: ARIA attributes like aria-required and aria-invalid are conditionally set, but there’s no fallback for invalid or undefined values.

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  protected renderLeadingIcon() {
+    if (!this.leadingIcon) {
+      return nothing;
+    }
+    return html`
+      <mdc-icon 
+        class="leading-icon" 
+        name=${this.leadingIcon as IconNames} 
+        size="1"
```
2020Deeya, line 140: question: We are providing hardcoded value (size='1') directly in here. Shouldn't we allow consumers to override icon sizes via properties or CSS custom properties?

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,222 @@
...
+    this.shadowRoot?.querySelector('input')?.focus();
+  }
+
+  protected renderClearButton() {
+    return html`
+      <mdc-button 
+        class='clear-button ${!this.value ? 'hidden' : ''}'
+        prefix-icon='cancel-regular'
+        variant='tertiary'
+        size="20"
```
2020Deeya, line 176: question: We are providing hardcoded value (size='20') directly in here. Shouldn't we allow consumers to override button size via properties or CSS custom properties?

packages/components/src/components/input/input.component.ts
arjunkesava: I think we shouldn't provide the option to change the icon size, 
if the consumer want's to change the size, they can change the entire input component, but not a specific icon, which looks dope.

packages/components/src/components/input/input.component.ts
arjunkesava: I think we shouldn't provide the option to change the clear button size, 
if the consumer want's to change the size, they can change the entire input component, but not a specific button, which looks dope.

packages/components/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,193 @@
...
+  label="Label"
+  labelInfoText="Label info text"
+  helpText="Help text"
+  helpTextType="default"
+  placeholder="Placeholder"
+  leadingIcon="placeholder-bold"
+  ></mdc-input>
+  </div>
+  `,
+};
```
arjunkesava, line 193: In figma, when the helper text is in error mode, then the input container color is set to red and when the consumer focus it, you can see both red and blue (focus) color.
image

This is missing.

arjunkesava (changes requested):

packages/components/src/components/input/input.stories.ts
arjunkesava: There is a character limit text to be displayed on the right bottom of the input component in Figma, this is also missing.
image

packages/components/src/components/input/input.styles.ts
supminn: yes, they are required to change the background color on hovering upon the input field.

packages/components/src/components/input/input.component.ts
supminn: The value for helpText changes to show validation message (when prefix text reaches max length)
To reflect this change on the element, it has been added.

packages/components/src/components/input/input.component.ts
supminn: Yes, they need to be hard coded. The button is of fixed size on the input element.

packages/components/src/components/input/input.component.ts
supminn: Yes, they need to be hard coded. The icon is of fixed size on the input element.

packages/components/src/components/input/input.stories.ts
supminn: The border color is set to red. On focus ring is having its default color.

This is the figma spec.
image

packages/components/src/components/input/input.component.ts
supminn: I read the MDN docs and understood these are not required. (since we use the native input element, a11y is handled)
Getting these removed.

packages/components/src/components/input/input.component.ts
cscheffauer: imho opinion clear button size should just be depending on the inputs size - no need to hard code anything, but working with `height: 100%` will make this much easier.

packages/components/src/components/input/input.component.ts
cscheffauer: same here, let the input / container determine the size of the icon, instead of hard coding it.

packages/components/src/components/input/input.styles.ts
cscheffauer: issue: it might still be focusable! u have to consider that its not focusable if the clear button is hidden.
imho we shouldn't hide it, but just conditionally render it.

supminn:
closing this PR, refer to https://github.com/momentum-design/momentum-design/pull/1031

PR #1006, opened 2024-12-20
chore: add focus trap mixin for popover

Jason20020:
### Description

- Create Focus Trap Mixin

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-446


packages/components/src/utils/mixins/FocusTrapMixin.ts
```diff
@@ -0,0 +1,318 @@
+import { property } from 'lit/decorators.js';
+import { FocusTrapClass } from './abstract/AbstractFocusTrap';
+
+type Constructor<T = {}> = new (...args: any[]) => T;
```
supminn, line 4: Its no longer required to define this line. We have moved it into `mixins/index.types.ts`
You can just import them as type.

packages/components/src/utils/mixins/FocusTrapMixin.ts
```diff
@@ -0,0 +1,318 @@
...
+    enablePreventScroll: boolean;
+}
+
+export const FocusTrapMixin = <T extends Constructor<FocusTrapClass>>(
+  superClass: T,
+) : T & Constructor<FocusTrapClass & FocusTrapClassInterface> => {
+  class FocusTrap extends superClass {
+    /**
+     * Determines whether the focus trap is enabled.
+     * If true, focus will be restricted to the content within this component.
```
supminn, line 17: `@default false` is missing from ts-docs.

packages/components/src/utils/mixins/FocusTrapMixin.ts
```diff
@@ -0,0 +1,318 @@
...
+     * Checks if the element is interactive.
+     */
+    private isInteractiveElement(element: HTMLElement): boolean {
+      const interactiveTags = new Set([
+        'BUTTON',
+        'DETAILS',
+        'EMBED',
+        'IFRAME',
+        'SELECT',
+        'TEXTAREA',
```
supminn, line 115: question: are these the only set of interactive tags?
I see Textarea is mentioned but not Input.

packages/components/src/utils/mixins/FocusTrapMixin.ts
```diff
@@ -0,0 +1,318 @@
...
+    override connectedCallback() {
+      super.connectedCallback();
+      this.addEventListener('keydown', this.handleKeydown);
+    }
+
+    override disconnectedCallback() {
+      super.disconnectedCallback();
+      this.removeEventListener('keydown', this.handleKeydown);
+    }
+  }
```
supminn, line 316:
```suggestion
override constructor(){
  super();
  this.addEventListener('keydown', this.handleKeydown)
  }
}
```

This will take care of registering as well as removing the listener for cleanup.

https://hawkticehurst.com/2023/11/you-are-probably-using-connectedcallback-wrong/

packages/components/src/utils/mixins/abstract/AbstractFocusTrap.ts
supminn, line 1: question: How is this useful? I don't see any implementation added for these functions.

supminn (changes requested):

PR #1007, opened 2024-12-20
style: create modal container

Jason20020:
### Description

- Create Modal Container
- Added Storybook docs

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-466


packages/components/src/components/modalcontainer/modalcontainer.component.ts
```diff
@@ -0,0 +1,75 @@
...
+class Modalcontainer extends Component {
+  /**
+   * Color of the modalcontainer
+   * @default tonal
+   */
+  @property({ type: String, reflect: true })
+  color: ModalContainerColor = DEFAULTS.COLOR;
+
+  /**
+   * Elevation of the modalcontainer
```
supminn, line 30: suggestion: a brief explanation and what each value of elevation would do and the different values available.

packages/components/src/components/modalcontainer/modalcontainer.component.ts
```diff
@@ -0,0 +1,75 @@
...
+   * @default true
+   */
+  @property({ type: String, attribute: 'aria-modal' })
+  override ariaModal: string = 'true';
+
+  /**
+   * Aria-label attribute to be set for accessibility
+   */
+  @property({ type: String, attribute: 'aria-label' })
+  override ariaLabel: string | null = null;
```
supminn, line 54: question: why is it set to `null`? since aria-label is a string, we could set the default value to an empty string ''

packages/components/src/components/modalcontainer/modalcontainer.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  public override render() {
+    return html`
+      <div 
+        id='modal-container' 
+        class='modal-container'
```
supminn, line 60:
```suggestion
        id='mdc-modal-container' 
        class='mdc-modal-container'  
```

packages/components/src/components/modalcontainer/modalcontainer.stories.ts
```diff
@@ -0,0 +1,61 @@
...
+    },
+    role: {
+      control: 'text',
+    },
+    'aria-modal': {
+      control: 'text',
+    },
+    'aria-label': {
+      control: 'text',
+      description: 'Provides an accessible name for the modal.',
```
supminn, line 44: suggestion: This description can be moved into tsdocs of the component. This instance can be omitted.

packages/components/src/components/modalcontainer/modalcontainer.styles.ts
```diff
@@ -0,0 +1,47 @@
...
+    .modal-container {
+      padding: 0.75rem;
+      background-color: var(--mdc-modalcontainer-primary-background-color);
+      border-radius: 0.5rem;
+      border: 0.0625rem solid var(--mdc-modalcontainer-border-color);
+    }
+
+    .modal-container[contrast] {
+      background-color: var(--mdc-modalcontainer-inverted-background-color);
+      border: 0.0625rem solid var(--mdc-modalcontainer-inverted-border-color);
```
supminn, line 28: nitpick: since the border width and style is the same as above(line 23), we could change only the color as 
`border-color: var(--mdc-modalcontainer-inverted-border-color)`

packages/components/src/components/modalcontainer/modalcontainer.e2e-test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const setup = async (args: SetupOptions) => {
+  const { componentsPage, ...restArgs } = args;
+  await componentsPage.mount({
+    html: `
+       <mdc-modalcontainer
+        ${restArgs.color ? `color="${restArgs.color}"` : ''}
+        ${restArgs.elevation ? `elevation="${restArgs.elevation}"` : ''}
+        role="dialog"
+        aria-label="avatar button"
```
supminn, line 24: issue: aria-label value is incorrect

packages/components/src/components/modalcontainer/modalcontainer.e2e-test.ts
```diff
@@ -0,0 +1,95 @@
...
+   */
+  test('visual-regression', async ({ componentsPage }) => {
+    const modalcontainerStickerSheet = new StickerSheet(componentsPage, 'mdc-modalcontainer');
+    const text = '<mdc-text>Lorem ipsum dolor sit amet.</mdc-text>';
+    await test.step('Elevation and Color', async () => {
+      modalcontainerStickerSheet.setChildren(text);
+      await modalcontainerStickerSheet.createMarkupWithCombination({ color: COLOR, elevation: ELEVATION });
+      await modalcontainerStickerSheet.mountStickerSheet();
+      const container = modalcontainerStickerSheet.getWrapperContainer();
+
```
supminn, line 89: suggestion: You could replace `test` with `test.steps` as each `test` block pertains to create a new instance of the components page which leads to performance drop

packages/components/src/components/modalcontainer/modalcontainer.e2e-test.ts
```diff
@@ -0,0 +1,95 @@
...
+    await test.step('attributes', async () => {
+      await attributeTestCases(componentsPage);
+    });
+  });
+
+  /**
+   * ACCESSIBILITY
+   */
+  test('accessibility', async ({ componentsPage }) => {
+    await componentsPage.accessibility.checkForA11yViolations('modalcontainer-default');
```
supminn, line 75: suggestion: You could run A11y on the stickersheet too. It would test for all the variants of the component at once.

supminn (changes requested):

packages/components/src/components/modalcontainer/modalcontainer.component.ts
cscheffauer: i would recommend to be careful with this - aria-label should either be set or not on the element at all, we should avoid a `<mdc-modalcontainer aria-label ../>` at all costs, cause automated accessibility testing might consider this to "pass", where the aria-label is actually not correctly set. please check whats the right way to remove the aria-label completely in the DOM tree.

packages/components/src/components/modalcontainer/modalcontainer.component.ts
```diff
@@ -0,0 +1,77 @@
...
+   */
+  @property({ type: Number, reflect: true })
+  elevation: ModalContainerElevation = DEFAULTS.ELEVATION;
+
+  /**
+   * Role of the modalcontainer
+   * @default dialog
+   */
+  @property({ type: String })
+  override role: string = DEFAULTS.ROLE;
```
supminn, line 49: suggestion: we need to restrict the role to `dialog` and `alertdialog`. We can create a TS type to support the 2 roles.
Currently, it is of the type string, hence we get the following issue.

image

packages/components/src/components/modalcontainer/modalcontainer.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  override ariaModal: string = 'true';
+
+  public override render() {
+    return html`
+      <div 
+        id='mdc-modal-container' 
+        class='mdc-modal-container'  
+        ?contrast="${this.color === 'contrast'}"
+        role="${this.role}"
+        aria-modal=${this.ariaModal}
```
supminn, line 65:
```suggestion

  public override render() {
    return html`
      <div 
        id='mdc-modal-container' 
        class='mdc-modal-container'  
        ?contrast="${this.color === 'contrast'}"
        role="${this.role}"
        aria-modal="true"
```

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-modal

aria-modal is relevant only on dialog and alertdialog containers. We could force set them to true since the modal container could have either of these 2 roles.

packages/components/src/components/modalcontainer/modalcontainer.e2e-test.ts
```diff
@@ -0,0 +1,96 @@
...
+const setup = async (args: SetupOptions) => {
+  const { componentsPage, ...restArgs } = args;
+  await componentsPage.mount({
+    html: `
+       <mdc-modalcontainer
+        ${restArgs.color ? `color="${restArgs.color}"` : ''}
+        ${restArgs.elevation ? `elevation="${restArgs.elevation}"` : ''}
+        role="dialog"
+        aria-label="modal container"
+        aria-model="dialog"
```
supminn, line 25: This is no longer required with the above mentioned changes.

packages/components/src/components/modalcontainer/modalcontainer.stories.ts
```diff
@@ -0,0 +1,69 @@
...
+import { COLOR, DEFAULTS, ELEVATION } from './modalcontainer.constants';
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
+import { disableControls } from '../../../config/storybook/utils';
+
+const render = (args: Args) => html`
+<mdc-modalcontainer
+  color="${args.color}"
+  elevation="${args.elevation}"
+  role="${args.role}"
+  aria-modal="${args['aria-modal']}"
```
supminn, line 13: same as above. This is not required.

packages/components/src/components/modalcontainer/modalcontainer.component.ts
```diff
@@ -0,0 +1,77 @@
...
+        aria-label=${ifDefined(this.ariaLabel)}
```
supminn, line 66:
```suggestion
        aria-label=${ifDefined(this.ariaLabel) as string}
```

Adding a type assertion to fix the issue.

supminn (changes requested):

packages/components/src/components/modalcontainer/modalcontainer.stories.ts
```diff
@@ -0,0 +1,64 @@
...
+import { disableControls } from '../../../config/storybook/utils';
+
+const render = (args: Args) => html`
+<mdc-modalcontainer
+  color="${args.color}"
+  elevation="${args.elevation}"
+  role="${args.role}"
+  aria-label="${args['aria-label']}"
+>
+  <mdc-text>Lorem ipsum dolor sit amet.</mdc-text>
```
supminn, line 15: question: We should set this as a part of children right?

packages/components/src/components/modalcontainer/modalcontainer.component.ts
```diff
@@ -0,0 +1,70 @@
...
+import { ifDefined } from 'lit/directives/if-defined.js';
+import styles from './modalcontainer.styles';
+import { Component } from '../../models';
+import { DEFAULTS } from './modalcontainer.constants';
+import type { ModalContainerColor, ModalContainerElevation, ModalContainerRole } from './modalcontainer.types';
+
+/**
+ * The `mdc-modalcontainer` component is an element used to
+ * display a modal container that can further be used in popover.
+ *
```
supminn, line 12: suggestion: Can we add tsdoc info about `children` here.
The contents of modal container is considered with the default slot.

packages/components/src/components/modalcontainer/modalcontainer.stories.ts
```diff
@@ -0,0 +1,64 @@
...
+      control: 'select',
+      options: Object.values(COLOR),
+    },
+    elevation: {
+      control: 'select',
+      options: Object.values(ELEVATION),
+    },
+    role: {
+      control: 'text',
+    },
```
supminn, line 38: issue: update this with the available role options.
suggestion: you could use a radio option here.

packages/components/src/components/modalcontainer/modalcontainer.e2e-test.ts
```diff
@@ -0,0 +1,95 @@
...
+import type { ModalContainerColor, ModalContainerElevation } from './modalcontainer.types';
+import StickerSheet from '../../../config/playwright/setup/utils/Stickersheet';
+import { COLOR, DEFAULTS, ELEVATION } from './modalcontainer.constants';
+
+type SetupOptions = {
+  componentsPage: ComponentsPage;
+  color?: ModalContainerColor;
+  elevation?: ModalContainerElevation;
+  children?: string;
+};
```
supminn, line 14:
```suggestion
type SetupOptions = {
  componentsPage: ComponentsPage;
  color?: ModalContainerColor;
  elevation?: ModalContainerElevation;
  role?: ModalContainerRole;
  children?: string;
};
```

packages/components/src/components/modalcontainer/modalcontainer.e2e-test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const setup = async (args: SetupOptions) => {
+  const { componentsPage, ...restArgs } = args;
+  await componentsPage.mount({
+    html: `
+       <mdc-modalcontainer
+        ${restArgs.color ? `color="${restArgs.color}"` : ''}
+        ${restArgs.elevation ? `elevation="${restArgs.elevation}"` : ''}
+        role="dialog"
```
supminn, line 23:
```suggestion
       ${restArgs.role ? `role="${restArgs.role}"` : ''}
```

packages/components/src/components/modalcontainer/modalcontainer.e2e-test.ts
```diff
@@ -0,0 +1,95 @@
...
+        await componentsPage.setAttributes(modalcontainer, { color });
+        await expect(modalcontainer).toHaveAttribute('color', color);
+      });
+    }
+    for (const elevation of Object.values(ELEVATION)) {
+      await test.step(`attribut elevation ${elevation} should be present as expected`, async () => {
+        await componentsPage.setAttributes(modalcontainer, { elevation: elevation.toString() });
+        await expect(modalcontainer).toHaveAttribute('elevation', elevation.toString());
+      });
+    }
```
supminn, line 54: suggestion: Test step for checking the role of the modal container could be added here.

supminn (changes requested):

supminn (approved):

PR #1017, opened 2025-01-08
fix: add activeType to update the class only after the icon loads

shakti97:
# Description
1. Added activeType state to retain the previous type/class until the new icon associated with the updated type is successfully loaded.
2. Implemented handlers for the load and error events to manage the icon loading process.
3. Upon the load event, the activeType state is updated, which subsequently updates the corresponding CSS class.

https://github.com/user-attachments/assets/fa19f06e-fa96-4c1e-b489-4aed6defd019

## Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-441

### DIFF
--- a/packages/components/src/components/presence/presence.component.ts
+++ b/packages/components/src/components/presence/presence.component.ts
@@ -40,12 +40,6 @@ class Presence extends Component {
   type: PresenceType = DEFAULTS.TYPE;
 
   /**
-   * State to track the active type (previous type until the new icon is loaded)
-   */
-  @state()
-  private activeType: PresenceType = DEFAULTS.TYPE;
-
-  /**
    * Acceptable values include:
    * - XX_SMALL
    * - X_SMALL
@@ -61,6 +55,13 @@ class Presence extends Component {
    */
   @property({ type: String, reflect: true })
   size: PresenceSize = DEFAULTS.SIZE;
+
+  /**
+   * @internal
+   * State to track the active type (previous type until the new icon is loaded)
+   */
+  @state()
+  private currentIconType: PresenceType = DEFAULTS.TYPE;
 
   /**
    * Get the size of the presence icon based on the given size type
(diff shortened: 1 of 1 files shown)


packages/components/src/components/presence/presence.component.ts
```diff
@@ -89,13 +95,30 @@ class Presence extends Component {
...
+  */
+  private handleOnLoad(): void {
+    this.activeType = this.type;
+  }
+
+  /**
+   * Handles an error that occurs when loading an icon.
+  */
+  private handleOnError(error: Event): void {
+    console.error(`Failed to load icon: ${this.icon} - ${error}`);
```
supminn, line 110: suggestion: avoid the use of console.error. We have [convention](https://github.com/momentum-design/momentum-design/blob/main/packages/components/conventions/clean-code-guidelines.md). You can go through this convention on [error handling](https://github.com/momentum-design/momentum-design/blob/main/packages/components/conventions/component-error-handling.md)

supminn (changes requested):

packages/components/src/components/presence/presence.component.ts
```diff
@@ -39,6 +39,12 @@ class Presence extends Component {
   @property({ type: String, reflect: true })
   type: PresenceType = DEFAULTS.TYPE;
 
+  /**
+   * State to track the active type (previous type until the new icon is loaded)
+   */
+  @state()
+  private activeType: PresenceType = DEFAULTS.TYPE;
+
```
arjunkesava, line 47: - Please add `@internal` decorator to this state property. Absence of `internal` keyword will let this prop visible in the storybook.
- It would be better to move this state down, (i.e. place this after the size property below).

packages/components/src/components/presence/presence.component.ts
```diff
@@ -89,13 +95,30 @@ class Presence extends Component {
...
+  private handleOnError(error: Event): void {
+    console.error(`Failed to load icon: ${this.icon} - ${error}`);
+  }
+
   public override render() {
     return html`
       <div class="mdc-presence mdc-presence__${this.size}">
         <mdc-icon
-          class="mdc-presence-icon mdc-presence-icon__${this.type}"
+          class="mdc-presence-icon mdc-presence-icon__${this.activeType}"
```
arjunkesava, line 117: The variable `activeType` doesn't make any meaningful sense.
Active is one of the icon names.
We set the icon color based on the icon type (which is the icon name).

Can we set it to something like `iconColorByIconName` ?

arjunkesava (changes requested):

packages/components/src/components/presence/presence.component.ts
shakti97: Added.

packages/components/src/components/presence/presence.component.ts
shakti97: Added

packages/components/src/components/presence/presence.component.ts
shakti97: Changed it to currentIconType as it stores the value of type.

packages/components/src/components/presence/presence.component.ts
```diff
@@ -56,6 +56,13 @@ class Presence extends Component {
   @property({ type: String, reflect: true })
   size: PresenceSize = DEFAULTS.SIZE;
 
+  /**
+   * @internal
+   * State to track the active type (previous type until the new icon is loaded)
```
arjunkesava, line 61:
```suggestion
   * State to track the current icon type (previous type until the new icon is loaded)
```

arjunkesava (approved):
LGTM.

supminn (approved):

PR #1022, opened 2025-01-10
chore: update focus ring styles for specific id

arjunkesava:
# Description

### Problem:
- The current implementation of focus ring focus the entire element.
- In the input type components such as input, checkbox, radio, we need to focus only on a specific section of the element instead of the entire element.

### Solution:
- We can now add a `className` to the focus group, and the focus ring will get attached only to that specific class instead of the entire element.

>  in `XXX.component.ts` file
```
<input  class="mdc-focus-ring"/>
```
> in `XXX.styles.ts` file,
```
...hostFocusRingStyles(true);
```
- If we want the focus ring for the entire element, then we don't need to specify any classname and it will be applied to the entire element.
```
...hostFocusRingStyles();
```

https://github.com/user-attachments/assets/2dc3698f-eff9-4b8c-adac-2b4abfe24c56

## Links

- https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-476

### DIFF
--- a/packages/components/src/utils/styles/index.ts
+++ b/packages/components/src/utils/styles/index.ts
@@ -1,4 +1,4 @@
-import { css, unsafeCSS } from 'lit';
+import { css } from 'lit';
 
 const hostFitContentStyles = css`
   :host {
@@ -10,7 +10,7 @@ const hostFitContentStyles = css`
   }
 `;
 
-const hostFocusRingStyles = (className?: string) => {
+const hostFocusRingStyles = (applyFocusRingOnClass = false) => {
   const baseHostStyleVariables = css`
     :host {
       --mdc-focus-ring-inner-color: var(--mds-color-theme-focus-default-0);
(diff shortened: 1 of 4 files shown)


packages/components/src/utils/styles/index.ts
```diff
@@ -10,36 +10,66 @@ const hostFitContentStyles = css`
...
-const hostFocusRingStyles = css`
-  :host {
-    --mdc-focus-ring-inner-color: var(--mds-color-theme-focus-default-0);
-    --mdc-focus-ring-middle-color: var(--mds-color-theme-focus-default-1);
-    --mdc-focus-ring-outer-color: var(--mds-color-theme-focus-default-2); 
-
-    --mdc-focus-ring-inner-width: 0.125rem;
-    --mdc-focus-ring-middle-width: calc(2 * var(--mdc-focus-ring-inner-width));
-    --mdc-focus-ring-outer-width: calc(0.0625rem + var(--mdc-focus-ring-middle-width));
+const hostFocusRingStyles = (className?: string) => {
```
supminn, line 13:
```suggestion
const hostFocusRingStyles = (applyFocusRingOnClass = false) => {
```

packages/components/src/utils/styles/index.ts
```diff
@@ -10,36 +10,66 @@ const hostFitContentStyles = css`
...
+        }
+        /* High Contrast Mode */
+        @media (forced-colors: active) {
+          .${unsafeCSS(className)}:focus-within {
+            outline: 0.125rem solid var(--mds-color-theme-focus-default-0);
+          }
+        }
+      `,
+    ];
   }
```
supminn, line 54:
```suggestion
  if (applyFocusRingOnClass) {
    return [
      baseHostStyleVariables,
      css`
      .mdc-focus-ring:focus-visible {
          outline: none;
        }
        :host([disabled]).mdc-focus-ring:focus {
          box-shadow: none;
        }
        .mdc-focus-ring:focus-within {
          position: relative;
          box-shadow: ${boxShadow};
        }
        /* High Contrast Mode */
        @media (forced-colors: active) {
          .mdc-focus-ring:focus-within {
            outline: 0.125rem solid var(--mds-color-theme-focus-default-0);
          }
        }
      `,
    ];
  }
```

In this way, we could avoid the usage of `unsafeCSS`. We could restrict attaching a fixed class name `mdc-focus-ring` on our component accordingly.

supminn (changes requested):

supminn (approved):

PR #1027, opened 2025-01-13
fix: Multiple illustration updates

morgan1bathe:
All illustration colors have been updated. No visual changes.


cscheffauer (changes requested):
@morgan1bathe all the default illustrations have been changed from black to white. i think this is not necessary, since the default illustrations to my knowledge should be theme agnostic, which means a consumer anyways would need to set the color (like icons). So my suggestion would be:
All default illustrations should actually NOT be colored, like the core icons - and then a consumer can override them. Maybe we can also organise the default illustrations in a way (separate folder or so), such that those are getting the color removed from the svg and they will pick up the text color next to it (like a icon does).

Let me know what u think.

PR #1031, opened 2025-01-15
refactor: input component extending internal label and helper

supminn:
### Description

Extending the internal class that contains helper text, validation along with label and its information to build the input component.

image

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-468


packages/components/src/index.ts
```diff
@@ -11,6 +11,8 @@ import Marker from './components/marker';
 import Divider from './components/divider';
 import Buttonsimple from './components/buttonsimple';
 import Avatarbutton from './components/avatarbutton';
+import Helperlabel from './components/helperlabel';
```
cscheffauer, line 14: issue: its an internal component, so it should **not** be exported here.

cscheffauer (changes requested):

packages/components/src/components/helperlabel/helperlabel.component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * helperlabel is an internal component that contains the label and helper/validation text
+ *  that can be configured in various ways to suit different use cases (most of the input related components).
+ * It is used as an internal component and is not intended to be used directly by consumers.
+ *
+ * @tagname mdc-helperlabel
+ *
+ */
+class Helperlabel extends Component {
```
cscheffauer, line 18: the naming is weird and confusing - i would recommend to look at other component libs, like Lion, Shoelace, Material how they name internal components and in particular this base input.

packages/components/src/components/helperlabel/helperlabel.component.ts
```diff
@@ -0,0 +1,139 @@
...
+  <slot name="label-icon">${this.renderLabelInfoToggleTip()}</slot>
+</div>`;
+}
+
+/**
+ * renders the footer container that contains the helpertext icon and helpertext.
+ * @returns void
+ */
+protected renderFooterHelperText() {
+  return html`<div class="footer" part="footer">
```
cscheffauer, line 122: suggestion: i would advise to not use header & footer, since those are words which might be used in applications for page header & footer. lets just stick to helper & label text

packages/components/src/index.ts
supminn: Although ButtonSimple is an internal component, we have exported it on line 12.
On another note, if we do not import/export them here, the component isn't registered and loaded properly within the E2E test cases.

cscheffauer: thats definitely an issue - i would recommend to NOT export ButtonSimple and not export the HelperLabel.
For E2E tests, just import them in the html file directly - we should *never* make our output of the package different just cause it doesn't work in our e2e tests.

kpankaj3 (commented):
issue: in disabled state, prefix text color is different than designs.

issue: in disabled state, clear icon shall not be present.

suggestion: in disabled state, shall we use ellipsis.

image

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  protected renderTrailingButton() {
+    if (!this.trailingButton) {
+      return nothing;
+    }
+    return html`
+      <mdc-button 
+        class='clear-button ${!this.value ? 'hidden' : ''}'
+        prefix-icon='cancel-regular'
```
kpankaj3, line 310: issue: the cancel icon shall be `cancel-bold`

image

packages/components/src/components/input/index.ts
```diff
@@ -0,0 +1,13 @@
+import Input from './input.component';
+import { TAG_NAME } from './input.constants';
+import '../button';
```
arjunkesava, line 3: Please add these two component imports as well.
```suggestion
import '../button';
import '../icon';
import '../text';
```

arjunkesava (commented):
Will review more tmrw.

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,360 @@
...
+        variant='tertiary'
+        size="20"
+        aria-label="Clear"
```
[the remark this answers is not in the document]
supminn: good catch. We are not supposed to do this. Any suggestions on how to handle this? Should we add a `clear-button-aria-label`?

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+    this.setValidityFromInput();
+  }
+
+  private onChange(event: Event) {
+    this.updateValue();
+    this.setValidityFromInput();
+    this.dispatchEvent(new Event('change', event));
+  }
+
+  protected renderLeadingIcon() {
```
2020Deeya, line 265: issue: When user provides wrong Icon name or an icon which doesn't exist at all, then this extra space should just be avoided. Please take a look at this snapshot. 

Screenshot 2025-02-05 at 1 25 42 PM

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+    if (!this.leadingIcon) {
+      return nothing;
+    }
+    return html`
```
2020Deeya, line 269: issue: When the user provides a correct icon name but then updates it with an incorrect name by removing extra letters from the end, the old icon still remains, which is not expected.

Screenshot 2025-02-05 at 1 31 14 PM

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+  /**
+   * The minimum number of characters that the input field can accept.
+   */
+  @property({ type: Number }) minlength?: number;
+
+  /**
+   * The autocapitalize attribute of the input field.
+   * @default 'off'
+   */
+  @property({ type: String }) override autocapitalize: AutoCapitalizeType = AUTO_CAPITALIZE.OFF;
```
2020Deeya, line 93: issue: When autocapitalize is on, it's not changing the input. 
Screenshot 2025-02-05 at 1 36 59 PM

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+        if (this.onerror) {
+          this.onerror(error);
+        }
+      });
+    }
+  }
+
+  private setValidityFromInput() {
+    if (this.inputElement) {
+      this.internals.setValidity(
```
2020Deeya, line 241: question: When the user adds extra spaces at the beginning, is it acceptable?
Screenshot 2025-02-05 at 1 41 09 PM

packages/components/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,241 @@
...
+  html` <mdc-input
+    @change="${action('onchange')}"
+    @focus="${action('onfocus')}"
+    @blur="${action('onblur')}"
+    id="${args.id}"
+    label="${args.label}"
+    help-text-type="${args['help-text-type']}"
+    help-text="${args['help-text']}"
+    placeholder="${args.placeholder}"
+    value="${args.value}"
```
2020Deeya, line 21: issue: In Storybook, when the maximum character length is set to 3, the value field still accepts more than 3 characters.
Screenshot 2025-02-05 at 1 45 08 PM

Screenshot 2025-02-05 at 1 47 33 PM

packages/components/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,241 @@
...
+import '.';
+import { html } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { action } from '@storybook/addon-actions';
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
+import { VALIDATION } from '../formfieldwrapper/formfieldwrapper.constants';
+import { disableControls } from '../../../config/storybook/utils';
+import { AUTO_CAPITALIZE } from './input.constants';
+
+const render = (args: Args) =>
```
2020Deeya, line 11: issue: There's an accessibility violation. 
image

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+    }
+    if (this.prevHelperText !== '' && this.helpText === PREFIX_TEXT_OPTIONS.HELPERTEXT) {
+      this.helpText = this.prevHelperText;
+      this.helpTextType = this.prevHelperTextType;
+      this.prevHelperText = '';
+    }
+    return html`<mdc-text class="prefix-text" tagname='span' type='body-midsize-regular'>${this.prefixText}</mdc-text>`;
+  }
+
+  private clearInputText() {
```
2020Deeya, line 301: question/issue: When clearInputText() sets this.value = '', no input event is dispatched, so forms and event listeners won't detect the change. Should we manually dispatch an input event after clearing the value ?

packages/components/src/components/input/input.component.ts
2020Deeya, line 269: issue: Here, this.leadingIcon is a string, but it is cast to IconNames without validation. We should validate this.leadingIcon before casting. 
`const iconName = this.leadingIcon as IconNames;`

If we don't do this, it will accept number also and create an extra space for the icon. (Refer snapshot)
Screenshot 2025-02-05 at 2 00 06 PM
Screenshot 2025-02-05 at 2 00 35 PM

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+    this.inputElement?.focus();
+  }
+
+  protected renderTrailingButton() {
+    if (!this.trailingButton) {
+      return nothing;
+    }
+    return html`
+      <mdc-button 
+        class='clear-button ${!this.value ? 'hidden' : ''}'
```
arjunkesava, line 313: - This class `clear-button` is not used in the input styles.
- The trailing icon has a padding of 2px on all around.

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+      this.helpText = PREFIX_TEXT_OPTIONS.HELPERTEXT;
+      this.helpTextType = PREFIX_TEXT_OPTIONS.VALIDATION;
+      return nothing;
+    }
+    if (this.prevHelperText !== '' && this.helpText === PREFIX_TEXT_OPTIONS.HELPERTEXT) {
+      this.helpText = this.prevHelperText;
+      this.helpTextType = this.prevHelperTextType;
+      this.prevHelperText = '';
+    }
+    return html`<mdc-text class="prefix-text" tagname='span' type='body-midsize-regular'>${this.prefixText}</mdc-text>`;
```
arjunkesava, line 298:
```suggestion
    return html`<mdc-text class="prefix-text" tagname=${TAGNAME.SPAN} type=${TYPE.BODY_MIDSIZE_REGULAR}>${this.prefixText}</mdc-text>`;
```

packages/components/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,241 @@
...
+    },
+    pattern: {
+      control: 'text',
+    },
+    list: {
+      control: 'text',
+    },
+    size: {
+      control: 'number',
+    },
```
arjunkesava, line 117:
```suggestion
    },
    ...hideControls(['internalValue', 'internals']),
```

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+ * @cssproperty --mdc-input-hover-background-color - Background color for the input field when hovered
+ * @cssproperty --mdc-input-focused-background-color - Background color for the input field when focused
+ * @cssproperty --mdc-input-focused-border-color - Border color for the input container when focused
+ * @cssproperty --mdc-input-error-border-color - Border color for the input container when error
+ * @cssproperty --mdc-input-warning-border-color - Border color for the input container when warning
+ * @cssproperty --mdc-input-success-border-color - Border color for the input container when success
+ * @cssproperty --mdc-input-primary-border-color - Border color for the input container when primary
+ *
+ */
+class Input extends NameMixin(FormfieldWrapper) {
```
arjunkesava, line 46: - The external id can be used to change styles or do a lot more.
```html
  <mdc-input id="inputId"></mdc-input>
```
- Do we allow the internal `id` to be changed by the consumer?
```html
  <mdc-input>
      <label for="internal-uuid"></label>
      <input type="text" id="internal-uuid" />
  </mdc-input>
```

Now if we allow the consumer to change the internal id, then that same id will be present in 3 places. Is that ok.
```html
<mdc-input id="user-given-id">
      <label for="user-given-id"></label>
      <input type="text" id="user-given-id" />
  </mdc-input>
  ```

File: packages/components/src/components/input/input.component.ts [hunk 18]
```diff
@@ -0,0 +1,364 @@
...
+   * The autocapitalize attribute of the input field.
+   * @default 'off'
+   */
+  @property({ type: String }) override autocapitalize: AutoCapitalizeType = AUTO_CAPITALIZE.OFF;
+
+  /**
+   * The autocomplete attribute of the input field.
+   * @default 'off'
+   */
+  @property({ type: String }) autocomplete: any = 'off';
```
arjunkesava, line 99: I guess this can be set to type of `string` instead of any.

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+   * Specifies a regular expression that the input value must match for validation purposes.
+   */
+  @property({ type: String }) pattern?: string;
+
+  /**
+   * The list attribute of the input field.
+   * Identifies a list of pre-defined options to suggest to the user.
+   * @default ''
+   */
+  @property({ type: String }) list = '';
```
arjunkesava, line 123:
```suggestion
   * @default undefined
   */
  @property({ type: String }) list?: string;
```

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+
+  /**
+   * The prefix text that is displayed before the input field. It has a max length of 10 characters.
+   */
+  @property({ type: String, attribute: 'prefix-text' }) prefixText = '';
+
+  /**
+   * The leading icon that is displayed before the input field.
+   */
+  @property({ type: String, attribute: 'leading-icon' }) leadingIcon = '';
```
arjunkesava, line 71: As these two props are optional, can we set the undefined by default and don't assign any empty value?

arjunkesava (changes requested):
- Please add return type and generate tsdoc for all methods in input.component.ts file?
-

packages/components/src/components/input/input.component.ts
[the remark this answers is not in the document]
supminn: For autocomplete, we have purposely added `any` as it supports a lot of options. Setting it to `string` was  giving TS error.

packages/components/src/components/input/input.stories.ts
supminn: This is for the disabled state. As per the designers, we do not consider A11y violation for disabled state.

packages/components/src/components/input/input.component.ts
supminn: yes, this will be considered as valid if `pattern` is not defined.
If regex pattern is provided, it gets validated and shows a error message to the user.

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+
+  @state() internalValue = '';
+
+  @property()
+  set value(val: string) {
+    this.internalValue = val;
+    this.internals.setFormValue(val);
+
+    this.updateComplete.then(() => {
+      this.setValidityFromInput();
```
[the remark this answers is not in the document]
supminn: This is handled within the `setValidityFromInput` function. It has the same check.

packages/components/src/components/input/input.stories.ts
supminn: Thanks. I've added ` /** @internal */` and handled them.

packages/components/src/components/input/input.component.ts
supminn: question: when do we prefer to do this? Should I also change the properties of mdc-button & mdc-icon to refer from constants?

packages/components/src/components/input/input.component.ts
supminn: Setting it to `string` was causing TS error. Hence its set to any. 
Let me know if there is a better way to support this.
image

packages/components/src/components/input/input.component.ts
supminn: I have removed 'clear-button' class.
We are using tertiary button of size 20. It has 1px border and 1px of padding making it look similar to figma.

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,364 @@
...
+            list=${ifDefined(this.list)}
+            size=${ifDefined(this.size)}
+            @input=${this.onInput}
+            @change=${this.onChange}
+          />
+        </slot>
+       </section>
+        <slot name="trailing-button">${this.renderTrailingButton()}</slot>
+      </div>
+      ${this.renderHelperText()}
```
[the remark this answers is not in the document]
supminn: Does this cause an issue? From the spec, it is defined that the input container must be 32px in height - which is intact.

packages/components/src/components/input/input.stories.ts
supminn: The value is restricted when you type into the input field. This issue is on storybook "value" attribute. Let me know if there is a way to handle this here..

packages/components/src/components/input/input.component.ts
supminn: point 1 - agreed, we cannot do much about how the consumers would want to override the styles using id
point 2 - no, we do not allow users to change the internal id
point 3 - we could change it to `data-id` for attribute name, but the currently implementation isn't causing any issue too.

Let me know if we would want to make this change or leave it as is.

kpankaj3 (dismissed):
LGTM

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,423 @@
...
+    this.updateValue();
+    this.setValidityFromInput();
+  }
+
+  /**
+   * Handles the change event of the input field.
+   * Updates the value and sets the validity of the input field.
+   * Dispatches the change event.
+   *
+   * @param event
```
arjunkesava, line 295: To avoid the tslint, error we need to add description about the passed argument parameter.
```suggestion
   * @param event - Event which contains information about the value change.
```

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,423 @@
...
+        size="1" 
+        length-unit="rem">
+      </mdc-icon>
+    `;
+  }
+
+  /**
+   * Renders the prefix text before the input field.
+   * If the prefix text is more than 10 characters, it will not be displayed.
+   * If the prefix text is more than 10 characters, the validation messsage will be displayed.
```
arjunkesava, line 327:
```suggestion
   * If the prefix text is more than 10 characters
   * - it will not be displayed.
   * - the validation message will be displayed.
```

packages/components/src/components/input/input.component.ts
arjunkesava: Yes, we should always use types instead of strings.
The reason to do that is if in future the text type body midsize regular gets an update, then this component will gets that update.

If we use strings, then there is no way of knowing that the mdc-text exists in this component.
Please always use constants / types wherever possible.

packages/components/src/components/input/input.stories.ts
2020Deeya: Yes, it's an issue only in Storybook. We could handle it like this:

image

packages/components/src/components/input/input.stories.ts
arjunkesava: Right, but we need to ignore this error in the storybook.
Because the e2e will not let you pass if there is even a single violation in storybook.

@2020Deeya , we can skip this here, but when @supminn  works on e2e for this component, we need to fix this issue.

packages/components/src/components/input/input.stories.ts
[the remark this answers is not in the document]
arjunkesava: The value prop is derived from the `ValueMixin`, we don't need to add any desc in this component.

packages/components/src/components/input/input.stories.ts
arjunkesava: These props on the storybook doesn't make any meaningful sense to the consumer.
Please remove them in the storybook by hiding them.
image

packages/components/src/index.ts
```diff
@@ -9,7 +9,8 @@ import Button from './components/button';
 import Bullet from './components/bullet';
 import Marker from './components/marker';
 import Divider from './components/divider';
-import AvatarButton from './components/avatarbutton';
+import Avatarbutton from './components/avatarbutton';
```
arjunkesava, line 12: This might be happened due to a bad rebase, the change should be this "AvatarButton" and not "Avatarbutton".

packages/components/src/index.ts
```diff
@@ -26,7 +27,8 @@ export {
   Bullet,
   Marker,
   Divider,
-  AvatarButton,
+  Avatarbutton,
```
arjunkesava, line 30:
```suggestion
  AvatarButton,
```

You might have accidently pulled this PR https://github.com/momentum-design/momentum-design/pull/1075
Please remove this changes and you are good to go.

arjunkesava (dismissed):
LGTM 👍🏻

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,435 @@
...
+      ></mdc-button>
+    `;
+  }
+
+  public override render() {
+    return html`
+      ${this.renderLabel()}
+      <div class="input-container mdc-focus-ring" part="input-container">
+        <slot name="input-leading-icon">${this.renderLeadingIcon()}</slot>
+       <section class="input-section" part="input-section">
```
cscheffauer, line 397: issue: in terms of accessibility `section` is actually something which is used to structure your app - using the section here doesn't make sense and i am also wondering why? can't this be a div?

packages/components/src/components/input/input.component.ts
```diff
@@ -0,0 +1,435 @@
...
+  protected renderPrefixText() {
+    if (this.prevHelperText !== '' && this.helpText === PREFIX_TEXT_OPTIONS.HELPERTEXT) {
+      this.helpText = this.prevHelperText;
+      this.helpTextType = this.prevHelperTextType;
+      this.prevHelperText = '';
+    }
+    if (!this.prefixText) {
+      return nothing;
+    }
+    if (this.prefixText.length > PREFIX_TEXT_OPTIONS.MAX_LENGTH) {
```
cscheffauer, line 342: where is this logic coming from? i don't see how the specs are actually describing this?
why do we need a max character limit? a lot of custom logic where i don't know if we really want this...

packages/components/src/components/input/input.styles.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  .leading-icon{
+    aspect-ratio: 1;
+  }
+
+  .input-container{
+    height: 2rem;
+    border-radius: 0.5rem;
+    border: 0.0625rem solid var(--mdc-input-border-color);
```
cscheffauer, line 59: we should use px instead here - i can explain this in a meeting more.

arjunkesava (dismissed):

PR #1036, opened 2025-01-20
fix: text ellipsis inherit

cscheffauer:
### Description

- Fixing an issue where the text-overflow, white-space and overflow are not being passed from the host to the text part. This causes a bug where elipsis can't be applied to text easily.


Jason20020 (approved):
Looks good to me

PR #1038, opened 2025-01-20
chore: build checkbox component

arjunkesava:
### Description
- Add checkbox component in this PR.

| Screenshots |
|-------------|
|image|
| <h3>A11y screen recording of checkbox component<h3><video loop title="Checkbox A11y" src="https://github.com/user-attachments/assets/e4cb66d5-7ba0-42cc-b527-a61066431b4d" />|

### Links
- https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-487


packages/components/src/components/checkbox/checkbox.stories.ts
```diff
@@ -0,0 +1,92 @@
+import type { Args, Meta, StoryObj } from '@storybook/web-components';
+import '.';
+import { html } from 'lit';
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
+
+const render = (args: Args) => html`
```
2020Deeya, line 6: issue: please check accessibility violations of each story.

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,143 @@
...
+          class="mdc-checkbox__input"
+          name="${ifDefined(this.name)}"
+          value="${ifDefined(this.value)}"
+          ?checked="${this.checked}"
+          ?disabled="${this.disabled}"
+          aria-checked="${this.checked}"
+          aria-disabled="${this.disabled}"
+          aria-label="${ifDefined(this.label)}"
+          tabindex="${this.disabled ? -1 : 0}"
+          @click="${this.handleClick}"
```
2020Deeya, line 129: issue: we should add @change event instead of supporting any particular event.

packages/components/src/components/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,73 @@
+import { css } from 'lit';
+import { hostFitContentStyles, hostFocusRingStyles } from '../../utils/styles';
+
+const styles = [hostFitContentStyles, css`
+  :host {
```
2020Deeya, line 5: Use custom CSS properties.

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,143 @@
...
+import { ICON_NAME } from './checkbox.constants';
+import styles from './checkbox.styles';
+
+/**
+ * Checkboxes allow users to select multiple options from a list or turn an item/feature on or off.
+ * These are often used in forms, settings, and selections in lists.
+ *
+ * A checkbox component contains an optional label and an optional helper text.
+ *
+ * @dependency mdc-icon
```
2020Deeya, line 19: nitpick: Shouldn't we add @dependency mdc-labelandhelper here.

packages/components/src/components/checkboxgroup/checkboxgroup.component.ts
```diff
@@ -0,0 +1,148 @@
...
+    super();
+    this.addEventListener('change', this.toggleParentCheckboxState);
+  }
+
+  /**
+   * Handles the change event for the parent checkbox of the checkboxgroup.
+   * If all children checkboxes are checked, then we set the parent checkbox to true,
+   * If any one of the children is unchecked, then we set the parent checkbox will be indeterminate.
+   */
+  private toggleParentCheckboxState(): void {
```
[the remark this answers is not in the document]
2020Deeya: This is still there. Are we considering this issue as an expected behaviour of the checkbox?

packages/components/src/components/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,73 @@
...
+  :host([checked]) .mdc-checkbox__icon-container,
+  :host([indeterminate]) .mdc-checkbox__icon-container {
+    background: var(--mds-color-theme-control-active-normal);
+    border: unset;
+  }
+  :host([disabled]) .mdc-checkbox__container:hover {
+    background: unset;
+  }
+  :host([disabled]) .mdc-checkbox__icon-container {
+    border: 1px solid var(--mds-color-theme-outline-primary-disabled);
```
supminn, line 26: use rem unit instead of px

packages/components/src/components/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+  .mdc-checkbox__icon-container {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    position: relative;
+    flex-grow: 0;
+    flex-shrink: 0;
+    background: unset;
+    border: 1px solid var(--mds-color-theme-outline-button-normal);
```
supminn, line 53: use rem unit instead of px

packages/components/src/components/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,73 @@
...
+    cursor: pointer
+  }
+  .mdc-checkbox__icon-container {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    position: relative;
+    flex-grow: 0;
+    flex-shrink: 0;
+    background: unset;
```
supminn, line 52: question: why do we require these to be set?

packages/components/src/components/checkboxgroup/checkboxgroup.styles.ts
```diff
@@ -0,0 +1,19 @@
...
+  }
+  :host([type="parent"]) > ::slotted(mdc-checkbox), ::slotted(mdc-checkboxgroup) {
+    margin-left: 1.5rem;
+  }
+  :host([type="parent"]):dir(rtl) > ::slotted(mdc-checkbox), ::slotted(mdc-checkboxgroup) {
+    margin-right: 1.5rem;
+  }
+  .mdc-checkboxgroup__header-text {
+    margin-bottom: 0.5rem;
+  }
```
supminn, line 16: question: why do we need to set this style?

packages/components/src/components/checkboxgroup/checkboxgroup.component.ts
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * @tagname mdc-checkboxgroup
+ *
+ * @slot - Children checkboxes of the checkboxgroup.
+ */
+class Checkboxgroup extends DisabledMixin(Component) {
+  /**
+   * The type of the checkbox group, either 'parent' or nothing.
+   */
+  @property({ type: String }) type?: string;
```
supminn, line 25: suggestion: Should we change this to a boolean property? `isParent`

supminn (commented):
Haven't tested checkboxgroup yet.

supminn (changes requested):
Requested a few changes

packages/components/src/components/checkboxgroup/checkboxgroup.component.ts
```diff
@@ -0,0 +1,197 @@
...
+ * checkboxgroup component acts as a parent container for a list of checkboxes.
+ *
+ * @dependency mdc-text
+ * @dependency mdc-checkbox
+ *
+ * @tagname mdc-checkboxgroup
+ *
+ * @slot - Children checkboxes of the checkboxgroup.
+ */
+class Checkboxgroup extends DisabledMixin(Component) {
```
cscheffauer, line 21: issue: there are a lot of problems in the checkbox group - can we please split them in separate PRs?

cscheffauer (changes requested):

packages/components/src/components/checkboxgroup/checkboxgroup.component.ts
arjunkesava: Yes, we can do that. 
I will remove checkboxgroup component in this PR and create another one, when this gets merged.

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,116 @@
...
+   * @default false
+   */
+  @property({ type: Boolean, reflect: true }) indeterminate = false;
+
+  /**
+   * Determines whether the text content should be displayed or not.
+   *
+   * @default false
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'hide-text' }) hideText = false;
```
cscheffauer, line 44: suggestion: this attribute is not necessary, users can just not pass in a label - the formfieldwrapper should handle this.

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,116 @@
...
+        class="mdc-checkbox__icon"
+        name="${this.indeterminate ? ICON_NAME.INDETERMINATE : ICON_NAME.CHECKED}"
+        size="1"
+        length-unit="rem"
+      ></mdc-icon>
+    ` : nothing;
+    const textContent = this.hideText ? nothing : html`
+      <div>
+        ${this.renderLabel()}
+        ${this.renderHelperText()}
```
cscheffauer, line 90: issue: helpertext is not used (thats what the constructor says), but it is still rendered here. imho thats not necessary.

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,116 @@
...
+    const checkboxIconContent = (this.checked || this.indeterminate) ? html`
+      <mdc-icon
+        class="mdc-checkbox__icon"
+        name="${this.indeterminate ? ICON_NAME.INDETERMINATE : ICON_NAME.CHECKED}"
+        size="1"
+        length-unit="rem"
+      ></mdc-icon>
+    ` : nothing;
+    const textContent = this.hideText ? nothing : html`
+      <div>
```
cscheffauer, line 88: if the renderHelpText is removed, it will only be this.renderLabel, which can be put directly  in line 109 (without the hideText attribute) - so we don't need anything of this here.

packages/components/src/components/checkbox/checkbox.e2e-test.ts
```diff
@@ -0,0 +1,103 @@
...
+      await icon.waitFor();
+      await expect(icon).toHaveAttribute('name', 'check-regular');
+    });
+
+    await test.step('should have disabled attribute when the disabled attribute is passed', async () => {
+      await componentsPage.setAttributes(checkbox, { disabled: '' });
+      await expect(checkbox).toHaveAttribute('disabled');
+    });
+  });
+};
```
cscheffauer, line 95: issue: all interaction tests are missing here, like clicking, keyboard, tab etc. there are also no Visual Regression tests either!

packages/components/src/components/checkbox/checkbox.stories.ts
```diff
@@ -0,0 +1,105 @@
...
+  args: {
+    label: 'Email Opt In',
+    'help-text': 'I agree to receiving monthly promotional emails.',
+  },
+};
+
+export const List: StoryObj = {
+  render: () => html`
+    <div style="display: flex; flex-direction: column;">
+      <mdc-checkbox @change="${action('onchange')}" label="Black Widow" checked help-text="This is a help text"></mdc-checkbox>
```
cscheffauer, line 83: eslint issue with the line length

packages/components/src/components/checkbox/checkbox.stories.ts
```diff
@@ -0,0 +1,105 @@
...
+      <mdc-checkbox @change="${action('onchange')}" label="Hawkeye" indeterminate disabled></mdc-checkbox>
+      <mdc-checkbox @change="${action('onchange')}" label="The Hulk" help-text="This is a help text"></mdc-checkbox>
+      <mdc-checkbox @change="${action('onchange')}" label="Iron Man" indeterminate></mdc-checkbox>
+      <mdc-checkbox @change="${action('onchange')}" label="Thor" help-text="This is a help text"></mdc-checkbox>
+      <mdc-checkbox @change="${action('onchange')}" label="Captain Marvel"></mdc-checkbox>
+    </div>
+  `,
+};
+
+export const BasicList: StoryObj = {
```
cscheffauer, line 94: there is no difference between List and BasicList, please clean up

packages/components/src/components/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const styles = [hostFitContentStyles, css`
+  :host {
+    display: flex;
+    flex-direction: row;
+    align-items: flex-start;
+    margin-bottom: 0.5rem;
+  }
+  .mdc-label-text {
+    height: 1rem;
```
cscheffauer, line 12: q: why is this necessary? the form-field-wrapper should do that?

packages/components/src/components/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+  :host([disabled][checked]) .mdc-checkbox__icon-container,
+  :host([disabled][indeterminate]) .mdc-checkbox__icon-container {
+    background: var(--mds-color-theme-control-active-disabled);
+  }
+  .mdc-checkbox__input {
+    margin: 0;
+    padding: 0;
+    position: absolute;
+    opacity: 0.1%;
```
cscheffauer, line 40: where is this visually hidden style coming from? why this special opacity value? have u done research on this?

cscheffauer (changes requested):
rough review, didn't test locally yet

packages/components/src/components/checkbox/checkbox.styles.ts
arjunkesava: There are two elements inside the `.mdc-checkbox__container`
- input[type="checkbox"] native element &
- div of class mdc-checkbox__icon-container

We do need input type checkbox on the DOM for a11y but also we don't want it to be displayed on the screen.
Due to which I've found a way to set the opacity of the input to 0.1 which will make the input visually hidden but all the actions will still be available.

I can explain more about this or if you have any better way then do let me know I will consider it.

packages/components/src/components/checkbox/checkbox.e2e-test.ts
arjunkesava: I've added the interactive tests via this a6abf5cc4a8226011d0957d7b738f1b6d77d222f commit. Please take a look.

packages/components/src/components/checkbox/checkbox.component.ts
cscheffauer: aria-label is something the consumer passes in, aria-label will not be calculated by the label.

packages/components/src/components/checkbox/checkbox.styles.ts
2020Deeya: Agreed, but it's always better to create custom variables on top and use it further.

packages/components/src/components/checkbox/checkbox.constants.ts
```diff
@@ -0,0 +1,11 @@
+import utils from '../../utils/tag-name';
+import type { IconNames } from '../icon/icon.types';
+
+const TAG_NAME = utils.constructTagName('checkbox');
+
+const ICON_NAME = {
+  CHECKED: 'check-regular' as IconNames,
```
cscheffauer, line 7: issue: according to specs this should be check-bold

packages/components/src/components/checkbox/checkbox.constants.ts
```diff
@@ -0,0 +1,11 @@
...
+  INDETERMINATE: 'minus-regular' as IconNames,
```
cscheffauer, line 8: issue: according to specs this should be minus-bold

packages/components/src/components/checkbox/checkbox.stories.ts
```diff
@@ -0,0 +1,103 @@
...
+      <mdc-checkbox label="Unselected" disabled></mdc-checkbox>
+      <mdc-checkbox label="Selected" disabled checked></mdc-checkbox>
+      <mdc-checkbox label="Indeterminate" disabled indeterminate></mdc-checkbox>
+      <mdc-checkbox label="Unselected" disabled help-text="This is a help text"></mdc-checkbox>
+      <mdc-checkbox label="Selected" disabled help-text="This is a help text" checked></mdc-checkbox>
+      <mdc-checkbox label="Indeterminate" disabled help-text="This is a help text" indeterminate></mdc-checkbox>
+    </div>`,
+};
+
+export const List: StoryObj = {
```
cscheffauer, line 86: suggestion: this one can be removed in favor of the Checkbox Group component

packages/components/src/components/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,77 @@
...
+    background: var(--mds-color-theme-control-active-normal);
+    border: unset;
+  }
+  :host([disabled]) .mdc-label,
+  :host([disabled]) .mdc-checkbox__input {
+    cursor: unset;
+  }
+  :host([disabled]) .mdc-checkbox__container:hover {
+    background: unset;
+  }
```
cscheffauer, line 26: issue: in the design specs the pressed is a different color (token)

cscheffauer (changes requested):

packages/components/src/components/checkbox/checkbox.e2e-test.ts
```diff
@@ -0,0 +1,215 @@
...
+
+      await componentsPage.page.keyboard.press('Space');
+      await expect(checkbox).not.toHaveAttribute('checked');
+    });
+
+    await test.step('checkbox should be checked when clicked', async () => {
+      const checkbox = await setup({ componentsPage });
+
+      await checkbox.click();
+      await expect(checkbox).toHaveAttribute('checked');
```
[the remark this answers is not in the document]
arjunkesava: Unfortunately, this is not working with our custom checkbox.

image

packages/components/src/components/checkbox/checkbox.e2e-test.ts
```diff
@@ -0,0 +1,215 @@
...
+      });
+    });
+  });
+
+  /**
+   * ACCESSIBILITY
+   */
+  await test.step('accessibility', async () => {
+    await componentsPage.accessibility.checkForA11yViolations('checkbox-default', true, {
+      rules: ['target-size'],
```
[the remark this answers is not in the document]
arjunkesava: According to the spec, https://dequeuniversity.com/rules/axe/4.10/target-size?application=playwright
the touch point should have enough space as there is chance of consumer clicking them accidentally.

In our scenario as we deal with a custom checkbox, we have to set the native checkbox and custom icon checkbox on top of each other, due to this reason I've disabled this rule.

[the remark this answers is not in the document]
cscheffauer: this should have been discussed with the Design team, not just "decided" by our own.
I think its fine to disable yes, but i would recommend to double check first always.

packages/components/src/components/checkbox/checkbox.e2e-test.ts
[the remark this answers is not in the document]
cscheffauer: i think it can work - your locator is currently pointing to `mdc-checkbox` directly, not the input in the shadow dom. Try to modify your locator to point to the input with type="checkbox" and then try it again..

packages/components/src/components/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,77 @@
...
+const styles = [hostFitContentStyles, css`
+  :host {
+    flex-direction: row;
+    align-items: flex-start;
+  }
+  .mdc-label, .mdc-checkbox__input {
+    cursor: pointer;
+  }
+  :host, .mdc-checkbox__container {
+    margin: 0.125rem 0;
```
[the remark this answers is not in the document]
cscheffauer: issue: if no help text is there, it is still 24px.

packages/components/src/components/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,77 @@
+import { css } from 'lit';
+import { hostFitContentStyles, hostFocusRingStyles } from '../../utils/styles';
+
+const styles = [hostFitContentStyles, css`
+  :host {
+    flex-direction: row;
+    align-items: flex-start;
+  }
```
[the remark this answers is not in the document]
cscheffauer: issue: i thought u would also check the rest of the specs as well - the hover state when "checked" is also differently styled.

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+  /**
+   * Toggles the state of the checkbox element.
+   * and dispatch the new change event.
+   */
+  public handleChange(event: Event): void {
+    this.toggleState();
+    this.setFormValue();
```
2020Deeya, line 92: issue/question: Currently, setFormValue() is only called inside handleChange(), meaning it won’t update the form value if checked changes programmatically later.
My suggestion would be moving setFormValue() to updated() so it always syncs with the state.

packages/components/src/components/checkbox/__screenshots__/chrome/mdc-checkbox-ltr.png
supminn, line 1: image
styling: On Figma spec, I observe that the checked state in disabled mode also has a border seen. However, on the implementation this border is missing.

packages/components/src/components/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,99 @@
...
+    --mdc-checkbox-checked-background-color: var(--mds-color-theme-control-active-normal);
+    --mdc-checkbox-disabled-background-color: var(--mds-color-theme-control-inactive-disabled);
+    --mdc-checkbox-disabled-border-color: var(--mds-color-theme-outline-primary-disabled);
+    --mdc-checkbox-disabled-checked-icon-color: var(--mds-color-theme-control-active-disabled);
+    --mdc-checkbox-disabled-icon-color: var(--mds-color-theme-text-primary-disabled);
+    --mdc-checkbox-icon-border-color: var(--mds-color-theme-outline-input-normal);
+    --mdc-checkbox-icon-color: var(--mds-color-theme-inverted-text-primary-normal);
+    --mdc-checkbox-icon-background-color: var(--mds-color-theme-control-inactive-normal);
+    --mdc-checkbox-pressed-icon-color: var(--mds-color-theme-control-inactive-pressed);
+
```
supminn, line 17: image
image

styling: When the checkbox is unselected and is in the rest state, it should have a background color of `--color-theme-control-inactive-normal`

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,136 @@
...
+  /** @internal */
+  private internals: ElementInternals;
+
+  /** @internal */
+  static formAssociated = true;
+
+  /** @internal */
+  get form(): HTMLFormElement | null {
+    return this.internals.form;
+  }
```
supminn, line 49: suggestion: As we support form internals, we would need to add all the corresponding attributes that the native checkbox element supports. Ex: We need to support "required" on the checkbox field when it is associated with a form.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#validation

packages/components/src/components/checkbox/checkbox.e2e-test.ts
```diff
@@ -0,0 +1,209 @@
...
+  });
+
+  /**
+   * ATTRIBUTES
+   */
+  await test.step('attributes', async () => {
+    const checkbox = await setup({ componentsPage });
+
+    await test.step('should fallback to default icon and size to 32 when no attributes are passed', async () => {
+      await componentsPage.setAttributes(checkbox, {});
```
supminn, line 121: nitpick: This line is not required as we aren't setting any attributes

packages/components/src/components/checkbox/checkbox.stories.ts
```diff
@@ -0,0 +1,84 @@
...
+  render: () => html`
+    <div style="display: flex; flex-direction: column;">
+      <mdc-checkbox label="Unselected" disabled></mdc-checkbox>
+      <mdc-checkbox label="Selected" disabled checked></mdc-checkbox>
+      <mdc-checkbox label="Indeterminate" disabled indeterminate></mdc-checkbox>
+      <mdc-checkbox label="Unselected" disabled help-text="This is a help text"></mdc-checkbox>
+      <mdc-checkbox label="Selected" disabled help-text="This is a help text" checked></mdc-checkbox>
+      <mdc-checkbox label="Indeterminate" disabled help-text="This is a help text" indeterminate></mdc-checkbox>
+    </div>`,
+};
```
supminn, line 84: suggestion: It would be a nice addition to add an instance of checkbox wrapped up in a form element.

supminn (changes requested):

packages/components/src/components/checkbox/checkbox.stories.ts
arjunkesava: To achieve that we need to import button component in this component. Won't it add cost to the build of this component?

packages/components/src/components/checkbox/checkbox.styles.ts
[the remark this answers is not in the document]
cscheffauer: issue: when disabled, the checkbox is still hoverable (it changes the background) - needs to be fixed.
steps to test: uncheck checkbox, set disabled, hover over the checkbox itself.
This is not an issue in the checked state, only in unchecked.

packages/components/src/components/checkbox/checkbox.e2e-test.ts
```diff
@@ -0,0 +1,215 @@
...
+    await componentsPage.accessibility.checkForA11yViolations('checkbox-default');
+  });
+
+  /**
+   * ATTRIBUTES
+   */
+  await test.step('attributes', async () => {
+    const checkbox = await setup({ componentsPage });
+
+    await test.step('should fallback to default icon and size to 32 when no attributes are passed', async () => {
```
supminn, line 127: issue: test step description is incorrect

packages/components/src/components/checkbox/checkbox.e2e-test.ts
```diff
@@ -0,0 +1,215 @@
...
+      checked: true,
+      style: 'margin: 0.25rem',
+    });
+    await checkboxStickerSheet.createMarkupWithCombination({}, true);
+    checkboxStickerSheet.setAttributes({
+      label: 'Disabled Indeterminate Checkbox Label',
+      'help-text': 'This is a help text',
+      disabled: true,
+      indeterminate: true,
+      style: 'margin: 0.25rem',
```
supminn, line 102: suggestion: Instead of adding `style: 'margin: 0.25rem'` across all the instances of stickersheet element, you could add styles to the wrapper as follows
```
 const rowWrappers = wrapper.querySelectorAll('.componentRowWrapper') as NodeListOf<HTMLElement>;
        Array.from(rowWrappers).forEach((rowWrapper) => {
          const modifiedRowWrapper = rowWrapper;
          modifiedRowWrapper.style.margin = '0.25rem';
        });
      });
      ```

This change can be considered as a nitpick.

Comment (arjunkesava) [reply packages/components/src/components/checkbox/checkbox.e2e-test.ts:102]: This is something I was missing, we should some how find a way to add styles to the sticker sheet class.

File: packages/components/src/components/checkbox/checkbox.stories.ts [hunk 35]
```diff
@@ -0,0 +1,148 @@
...
+  render: () => {
+    const handleSubmit = (event: Event) => {
+      event.preventDefault();
+      const formData = new FormData(event.target as HTMLFormElement);
+      const selectedValues = formData.getAll('super-power');
+      action('Form Submitted')({ value: selectedValues });
+    };
+
+    return html`
+      <form @submit=${handleSubmit}>
```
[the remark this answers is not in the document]
arjunkesava: When indeterminate, the form submission should not happen. It is just a feature provided to select/unselect all children at once.

For a single checkbox component, we don't need to do anything for indeterminate state.
But for a checkbox group component (which will be done soon), we need this logic.

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * Updates the form value to reflect the current state of the checkbox.
+   * If checked, the value is set to either the user-provided value or 'on' if no value is provided.
+   * If unchecked, the value is set to null.
+   */
+  private setFormValue() {
+    let actualValue: string | null = null;
+
+    if (this.checked) {
+      actualValue = !this.value ? 'on' : this.value;
```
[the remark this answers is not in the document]
arjunkesava: It is default behavior of a checkbox.
```
If the value attribute was omitted, the default value for the checkbox is `on`,
so the submitted data in that case would be `subscribe=on`.
```

Above para is from the docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * Toggles the state of the checkbox element.
+   * and dispatch the new change event.
+   */
+  public handleChange(event: Event): void {
+    this.toggleState();
+
+    // Change event doesn't bubble out of shadow dom,
+    // Workaround to fix this: https://github.com/lit/lit-element/issues/922#issuecomment-611139629
+    const newEvent = new (event.constructor as typeof Event)(event.type, event);
```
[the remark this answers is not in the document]
arjunkesava: Both are same @2020Deeya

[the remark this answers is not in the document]
supminn: as we discussed during our standup, we should make use of the same event rather than creating a new event instance.
so, the way its mentioned currently works well.

packages/components/src/components/checkbox/checkbox.styles.ts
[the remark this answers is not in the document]
cscheffauer: issue: that is fixed, the border color in unchecked disabled is still not matching design specs - please adjust.

cscheffauer (dismissed):

supminn (dismissed):

cscheffauer (approved):

packages/components/src/components/checkbox/checkbox.component.ts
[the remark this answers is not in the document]
2020Deeya: Both ways are similar, the way of writing it is different.

PR #1048, opened 2025-01-24
test: form field wrapper with label

supminn:
### Description

E2E testing for form field wrapper

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-486


packages/components/src/components/formfieldwrapper/formfieldwrapper.styles.ts
```diff
@@ -9,9 +9,13 @@ const styles = [
       align-items: flex-start;
       gap: 0.5rem;
     }
-
+    :host([disabled]),
     :host([disabled]) .mdc-label,
-    :host([disabled]) .mdc-help-text {
+    :host([disabled]) .mdc-help-text,
+    :host([disabled][help-text-type='error']) .mdc-help-text,
```
cscheffauer, line 15: nitpick: line 15-18 is not necessary, it should style it none the less

packages/components/src/components/formfieldwrapper/formfieldwrapper.component.ts
```diff
@@ -110,14 +111,6 @@ class FormfieldWrapper extends Component {
     </div>`;
```
cscheffauer, line 111: suggestion: i can't comment on the line, but in line 53 the part should not be `mdc-label`, but just `label`. We can create a convention for the naming of parts - in general they should never be prefixed, to not confuse them with actual components.

cscheffauer (approved):
2 comments, which needs to be addressed, other than that looking good.

cscheffauer (approved):

PR #1057, opened 2025-01-27
docs: restyling and several fixes

cscheffauer:
### Description

- Restyling all pages in our Docs page to look better and closer to Momentum Design
- Removed "Docs" heading item, not providing value
- Made everything mobile responsive, tested in Chrome Dev Tools (will do another test on a real mobile device once deployed)
- Fixed background gradient image by replacing with a new one which is not cut off
- Removed Theme Toggle, not necessary anymore.
- Fixed several issues in the Assets browsers when pressing Next, Previous & changing filters:
  - When changing a filter, it is automatically resetting to Page 1
  - When the end of the filtered items is reached, the Next button will be greyed out
  - Pressing Previous button actually reduces the Current Page by 1, not falsely jumping always to the first page

### Screenshots

image
image
image
image
image


kpankaj3 (commented):
suggestion (test, blocking): It appears like the home tab has no content on mobile device(chrome dev tools) with light theme on device.

image

kpankaj3 (approved):
LGTM

PR #1059, opened 2025-01-28
fix: storybook actions naming

supminn:
### Description

Updating the naming convention for storybook actions since we are creating web components.

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-509

PR #1073, opened 2025-01-30
docs: create toggle component

2020Deeya:
### Description

1] Create Toggle Component
2] Add Storybook docs

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-506
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-507


packages/components/src/components/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  :host([size='compact']) .mdc-toggle__icon-container {
+    width: var(--mdc-toggle-icon-container-size-compact);
+    height: var(--mdc-toggle-icon-container-size-compact);
+  }
+  
+  .mdc-toggle__icon {
+    padding: 0.25rem;
+   // --mdc-icon-fill-color: var(--mds-color-theme-text-primary-normal);
```
2020Deeya, line 109: This should be added, but somehow it seems not working in light mode. Looking into it.

packages/components/src/components/toggle/toggle.component.ts
```diff
@@ -0,0 +1,124 @@
...
+   * @default default
+   */
+  @property({ type: String, reflect: true })
+  size: ToggleSize = DEFAULTS.SIZE;
+
+  /**
+   * Determines aria label on mdc-toggle.
+   */
+  @property({ type: String, reflect: true, attribute: 'data-aria-label' })
+  dataAriaLabel = '';
```
2020Deeya, line 47: dataAriaLabel is temporarily added to avoid the accessibility issue. It will be removed once checkbox PR (including the dataAriaLabel mixin) gets merged.
https://github.com/momentum-design/momentum-design/pull/1038/

packages/components/src/components/toggle/toggle.component.ts
```diff
@@ -0,0 +1,128 @@
...
+import { ifDefined } from 'lit/directives/if-defined.js';
+import styles from './toggle.styles';
+import { Component } from '../../models';
+import FormfieldWrapper from '../formfieldwrapper';
+import { ValueMixin } from '../../utils/mixins/ValueMixin';
+import { NameMixin } from '../../utils/mixins/NameMixin';
+import { DisabledMixin } from '../../utils/mixins/DisabledMixin';
+import { DEFAULTS, ICON_NAME, TOGGLE_SIZE } from './toggle.constants';
+import { ToggleSize } from './toggle.types';
+import { ValidationType } from '../formfieldwrapper/formfieldwrapper.types';
```
supminn, line 12:
```suggestion
import type { ValidationType } from '../formfieldwrapper/formfieldwrapper.types';
```

packages/components/src/components/toggle/toggle.component.ts
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Toggles the state of the toggle element.
+   * and dispatch the new change event.
+   */
+  private handleChange(event: KeyboardEvent) {
+    if (this.disabled) return;
+
+    if (['Enter', ' '].includes(event.key)) {
+      event.preventDefault();
+    }
```
supminn, line 65: issue: We are not suppose to invoke the change on 'enter'. It is restricted for form submission.
This code is not required to be added.

packages/components/src/components/toggle/toggle.component.ts
```diff
@@ -0,0 +1,128 @@
...
+          </div>
+        </div>
+        <div class="mdc-toggle__label-container">
+          ${this.renderLabel()}
+          ${this.renderHelperText()}
+        </div>
+    `;
+  }
+
+  public static override styles: Array<CSSResult> = [...Component.styles, ...styles];
```
supminn, line 125:
```suggestion
  public static override styles: Array<CSSResult> = [...FormfieldWrapper.styles, ...styles];
```

packages/components/src/components/toggle/toggle.stories.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export const WithInfoIcon: StoryObj = {
+  render: () => html`
+    <mdc-toggle
+    label="Toggle label"
+    checked
+    >
+      <mdc-icon slot="label-info" name="info-circle-filled"></mdc-icon>
+    </mdc-toggle>`,
+};
```
supminn, line 91: suggestion: Let's not show this story at the moment. We could get it added once the toggletip component is ready for consumption.

packages/components/src/components/toggle/toggle.stories.ts
supminn, line 1: question: in RTL, shouldn't the cross and check icon positions changed too?

image
image

supminn, line 1: issue: Looks like there is some styling alignment issue when we interact with the toggle.
https://github.com/user-attachments/assets/04f4ba99-b699-4f91-a898-c1367acb00fa

supminn (changes requested):

packages/components/src/components/toggle/toggle.component.ts
```diff
@@ -0,0 +1,128 @@
...
+  private handleChange(event: KeyboardEvent) {
+    if (this.disabled) return;
+
+    if (['Enter', ' '].includes(event.key)) {
+      event.preventDefault();
+    }
+
+    this.checked = !this.checked;
+    // Since change event doesn't bubble out of shadow dom, we need to explicitly dispatch it.
+    this.dispatchEvent(new CustomEvent('onToggleChange', { detail: { checked: this.checked } }));
```
cscheffauer, line 69: issue: we should not create a custom event here, but redispatch the existing change event. Check @arjunkesava 's PR for checkbox for this. This also includes removing the event.preventDefault in line 64

packages/components/src/components/toggle/toggle.stories.ts
2020Deeya: The icon is an internal part of the toggle switch, so its position should not be changed. However, I have asked the designers for confirmation. 
If needed, It will be updated.

2020Deeya: Updated.

2020Deeya: We should keep border as none instead of transparent in order to maintain overall height as 20px when set to compact.

packages/components/src/components/toggle/index.ts
```diff
@@ -0,0 +1,14 @@
+import Toggle from './toggle.component';
+import { TAG_NAME } from './toggle.constants';
+import '../icon';
+import '../button';
```
supminn, line 4: question: why do we need to import button?

packages/components/src/components/toggle/toggle.stories.ts
supminn: Replicate the same logic on both inactive and active state.
Inactive state has a border of 1px and the toggle height is 18px => 18 + 1 + 1 = 20px
Maintain the same dimensions for active state with the border of 1px and update the `border-color: transparent`. 
This will fix the issue.

packages/components/src/components/toggle/toggle.stories.ts
```diff
@@ -0,0 +1,122 @@
...
+import { DEFAULTS, TOGGLE_SIZE } from './toggle.constants';
+
+const render = (args: Args) => html`
+  <mdc-toggle
+      name="toggleName"
+      value="toggleValue"
+      @keydown="${action('onKeyDown')}"
+      @keyup="${action('onKeyUp')}"
+      @focus="${action('onFocus')}"
+      @change="${action('onToggleChange')}"
```
supminn, line 16: issue: naming convention
Refer https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-509 bug, and fixes on https://github.com/momentum-design/momentum-design/pull/1059

supminn (changes requested):

packages/components/src/components/toggle/index.ts
2020Deeya: It's used in the one of the stories.

packages/components/src/components/toggle/toggle.component.ts
2020Deeya: Updated.

packages/components/src/components/toggle/index.ts
2020Deeya: Imported the button in the story instead of index.ts

supminn (dismissed):
LGTM

supminn (dismissed):

packages/components/src/components/toggle/toggle.component.ts
2020Deeya: It's updated already as per checkbox component.

packages/components/src/components/toggle/toggle.component.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /**
+   * Toggles the state of the toggle element.
+   * and dispatch the new change event.
+   */
+  private handleChange(event: Event) {
+    this.toggleState();
+    // Re-dispatch the existing event instead of creating a new one since change event doesn't bubble out of shadow dom
+    this.dispatchEvent(event);
```
supminn, line 112:
```suggestion
     const EventConstructor = event.constructor as typeof Event;
    this.dispatchEvent(new EventConstructor(event.type, event));
```

2020Deeya: Updated.

supminn (approved):

cscheffauer (approved):

PR #1075, opened 2025-01-31
refactor: update avatarbutton name on the index

arjunkesava:
### Description

- Rename Avatarbutton component name to AvatarButton component.


cscheffauer (approved):
good catch

PR #1076, opened 2025-02-03
chore: Tab component + storybook

kpankaj3:
### Description

This PR is for Tab component and it's storybook page.

Example Tab usage with badge and icon (with all attributes):

```
<mdc-tab 
  active
  disabled
  icon-name="placeholder-bold" 
  tabindex="0"
  text="Label"
  variant="pill">
  <mdc-badge slot="badge" type="counter" counter="1"></mdc-badge>
</mdc-tab>
```
A badge for tab can be provided using named slot `'badge'`

Tab has three variants
- Glass
- Line
- Pill

Tab has two states for tab is selected or not selected
- Active (active = true) - the Tab is selected.
- Inactive (active = false) - the Tab is not selected.

Tab can be disabled using disabled attribute. It will not be clickable and focusable. The host will have aria-disabled attribute set to true.

Tab can have an icon set using icon-name attribute. Internally, the icon-name is used to load icon using mdc-icon component.

Tab component supports multiple css variables to customise the Tab UI.

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-505

#### Glass Tab (Active)
image

#### Line Tab (Active)
image

#### Pill Tab (Active)
image

#### Icon Only Tab (Active, Variant - pill)
image

#### Custom CSS Properties
image
image
image


packages/components/src/components/tab/tab.component.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Tab can have two variants:
+   * - `pill`
+   * - `line`
+   * It defines the background and foreground color of the tab.
+   * @default pill
+   */
+  @property({ type: String, attribute: 'tab-variant', reflect: true })
+  tabVariant: TabVariant = DEFAULTS.TAB_VARIANT;
```
cscheffauer, line 50: suggestion: lets use "variant" as the attribute name, same as in other components

cscheffauer (changes requested):
just scanned it quickly, left 2 comments

packages/components/src/components/tab/tab.component.ts
kpankaj3: Updated

packages/components/src/components/tab/tab.component.ts
```diff
@@ -0,0 +1,152 @@
...
+  private prevIconName?: string;
+
+  /**
+   * @internal
+   */
+  private prevTabindex : number | null = null;
+
+  constructor() {
+    super();
+    this.role = 'button';
```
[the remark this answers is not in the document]
kpankaj3: Updated

packages/components/src/components/tab/tab.styles.ts
```diff
@@ -0,0 +1,157 @@
+import { css } from 'lit';
+import { hostFitContentStyles, hostFocusRingStyles } from '../../utils/styles';
+
+const styles = [hostFitContentStyles, css`
+
+  :host {
```
cscheffauer, line 6: issue: when toggling between active and non-active, the Tab gets slightly bigger and smaller in width - we need to make sure it stays the exact same width, doesn't matter if active or non-active

This exists in both pill and line - please check the design specs and follow them exactly.

packages/components/src/components/tab/tab.stories.ts
```diff
@@ -0,0 +1,153 @@
...
+import '.';
+import '../badge';
+import { html } from 'lit';
+import { action } from '@storybook/addon-actions';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { TAB_VARIANTS } from './tab.constants';
+import { disableControls, hideControls, readOnlyControls } from '../../../config/storybook/utils';
+
+const render = (args: Args) => html`<div role="tablist">
+  <mdc-tab
```
cscheffauer, line 11: issue: when i am in storybook and i "click" on the tab, its going to switch the icon and label text styling to active, but not the background. Only when i set "active" in the controls of the storybook it does set the full active styling.
This needs to be cleaned up (either clicking does everything) or we keep the component a controlled component and only set the active state from the outside.

packages/components/src/components/tab/tab.stories.ts
```diff
@@ -0,0 +1,153 @@
...
+      control: 'boolean',
+      description: 'Tab can be active or inactive. Active tab means the tab is selected.',
+    },
+    disabled: {
+      control: 'boolean',
+    },
+    'icon-name': {
+      control: 'text',
+    },
+    role: {
```
cscheffauer, line 48: issue: the role tab should NEVER be overridable by a consumer, so we have to remove it from storybook here.
image

cscheffauer (changes requested):

kpankaj3:
Figma designs updated - New tab type - Glass added. PR will be reopened after adding tab type.

packages/components/src/components/tab/tab.component.ts
```diff
@@ -0,0 +1,169 @@
...
+    element.setAttribute('aria-selected', active ? 'true' : 'false');
+  }
+
+  public override update(changedProperties: PropertyValueMap<any> | Map<PropertyKey, unknown>): void {
+    if (changedProperties.has('active')) {
+      this.modifyIconName(this.active);
+      this.setAriaSelected(this, this.active);
+      // Remove the active property from the changedProperties
+      // to prevent the base class from using to add aria-pressed attribute.
+      changedProperties.delete('active');
```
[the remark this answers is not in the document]
kpankaj3: Buttonsimple's setAriaPressed(element: HTMLElement, active: boolean) was setting aria-pressed attribute when active state updated in update() function. This role is not valid for role tab.

We have refactored the Buttonsimple function `private setAriaPressed(element: HTMLElement, active: boolean)` to `protected setActive(element: HTMLElement, active: boolean)`

Overriding this in Tab as:
```
protected override setActive(element: HTMLElement, active: boolean) {
    element.setAttribute('aria-selected', active ? 'true' : 'false');
    this.modifyIconName(active);
}
 ```

File: packages/components/src/components/tab/tab.stories.ts (hunk 5 above)
Comment (kpankaj3) [reply packages/components/src/components/tab/tab.stories.ts]: Updated

File: packages/components/src/components/tab/tab.styles.ts [hunk 7]
```diff
@@ -0,0 +1,157 @@
...
+    --mdc-tab-content-gap: 0.5rem;
```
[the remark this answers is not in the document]
kpankaj3: Fixed - the order of hover, active and focus states in styles.

packages/components/src/components/tab/tab.stories.ts
kpankaj3: Fixed - Added reflect=true to active property of Buttonsimple.

packages/components/src/components/tab/tab.styles.ts
kpankaj3: Updated but there will be slight width increase in active state due to bold font type.

cscheffauer (approved):

PR #1077, opened 2025-02-03
fix: several icon fixes

cscheffauer:
### Description

- Fixing multiple issues in Icon:
  - Adding focus ring styles to Icon to be able to make it focusable (new accessibility requirement, see docs) 
  - Applied border-radius for cases where the icon can get focused (if tabindex is set). 
  - role and aria-label don't need to be set on the svg itself, if set on the host
  - Moving event dispatch into calling function
  - `aria-hidden` can be set in util directly instead of separate function.


deveshcisco (approved):

packages/components/src/components/icon/icon.component.ts
```diff
@@ -37,10 +37,17 @@ import type { IconNames } from './icon.types';
...
+ * - For informative icons, an `aria-label` is required, and the `role` will be set to "img" automatically.
  * - If an `aria-label` is provided, the role will be set to 'img'; if it is absent,
  *   the role will be unset.
  *
+ * ### Informative Standalone Icons
+ * - If an icon is informative (as mentioned above) and does not belong to a button (=standalone), it must
+ * have a Tooltip that describes what it means.
+ * - For informative standalone icons, an `aria-label` & `tabindex="0"` is required,
+ * and the `role` will be set to "img" automatically.
+ * - Only use this when a Icon is standalone and is not part of a button or other interactive elements.
```
supminn, line 49: suggestion: Can we highlight this line to be bold?

supminn (approved):

2020Deeya (commented):
LGTM

PR #1078, opened 2025-02-05
feat(virtualizedlist): added component and test

johnsoter13:
### Description
- Added virtualizedlist component
- Added test


packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
```diff
@@ -0,0 +1,105 @@
...
+import { VirtualizerController } from '@tanstack/lit-virtual';
+import { property } from 'lit/decorators.js';
+import { Virtualizer, VirtualizerOptions } from '@tanstack/virtual-core';
+import { StyleInfo } from 'lit/directives/style-map.js';
+import { Ref, createRef, ref } from 'lit/directives/ref.js';
+import styles from './virtualizedlist.styles';
+import { Component } from '../../models';
+import { DEFAULT_COUNT, DEFAULT_MEASURE_ELEMENT } from './virtualizedlist.constants';
+/**
+ * virtualizedlist component, which ...
```
cscheffauer, line 12: docs: proper documentation is missing here

packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @tagname mdc-virtualizedlist
+ *
+ */
+class VirtualizedList extends Component {
+  @property({ type: Function, attribute: 'onscroll' })
```
cscheffauer, line 18: docs: every public property needs to have a TSDoc

packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  constructor() {
+    super();
+    this.virtualizerController = null;
+    this.virtualizer = null;
+  }
+
+  public override update(changedProperties: PropertyValues): void {
+    super.update(changedProperties);
+    if (changedProperties.get('virtualizerprops')) {
```
cscheffauer, line 38: suggestion: please document why `.get` is used here and also what this function here does

packages/components/src/components/virtualizedwrapper/virtualizedwrapper.component.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @tagname mdc-virtualizedwrapper
+ * *
+ */
+
+const sentences = new Array(1000)
+  .fill(true)
+  .map((_, index) => `sentence number ${index}`);
+
+class VirtualizedWrapper extends Component {
```
cscheffauer, line 21: issue: since all of this here is just made for storybook (for the actual virtualized list), i would recommend to move it inside of the virtualizedlist folder and name it in a way that its clear that this is a component for storybook only. you could create a "stories" folder or something like that and put it in.

packages/components/src/components/virtualizedwrapper/virtualizedwrapper.constants.ts
```diff
@@ -0,0 +1,5 @@
+import utils from '../../utils/tag-name';
```
cscheffauer, line 1: issue: this is not necessary, since the virtualizedwrapper is not a component which is exported from the repo, but only used in storybook.

packages/components/src/components/virtualizedwrapper/virtualizedwrapper.stories.ts
```diff
@@ -0,0 +1,33 @@
...
+    ...classArgType,
+    ...styleArgType,
+  },
+};
+
+export default meta;
+
+export const Example: StoryObj = {
+  args: {
+    onScroll: () => console.log('hello'),
```
cscheffauer, line 30: suggestion: use the storybook "actions" method (see other stories)

packages/components/src/components/virtualizedwrapper/virtualizedwrapper.stories.ts
```diff
@@ -0,0 +1,33 @@
+import type { Meta, StoryObj, Args } from '@storybook/web-components';
```
cscheffauer, line 1: issue: this is now called virtualizedwrapper, but the actual component exported is virtualizedlist.
i would recommend to keep everything the same name and if u need a helper component, then only create for that a separate file. This will be very confusing for consumers otherwise, since they might search for a "virtualizedwrapper", which doesn't exist.

packages/components/src/components/virtualizedwrapper/virtualizedwrapper.stories.ts
```diff
@@ -0,0 +1,33 @@
...
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
+
+const render = (args: Args) => html`
+  <mdc-virtualizedwrapper
+    .virtualizerprops="${args.virtualizerprops}"
+    @on-scroll="${args.onScroll}"
+  ></mdc-virtualizedwrapper>`;
+
+const meta: Meta = {
+  title: 'Work In Progress/virtualizedwrapper',
```
cscheffauer, line 13: since the whole PR will "release" this component, you can change the title here to `Components/virtualizedwrapper`

packages/components/src/components/virtualizedwrapper/virtualizedwrapper.stories.ts
```diff
@@ -0,0 +1,33 @@
...
+  tags: ['autodocs'],
+  component: 'mdc-virtualizedwrapper',
+  render,
+  parameters: {
+    badges: ['wip'],
```
cscheffauer, line 18: since the whole PR will "release" this component, you can change the badges here to `['stable']`

packages/components/src/components/virtualizedwrapper/virtualizedwrapper.component.ts
```diff
@@ -0,0 +1,53 @@
...
+    this.list = html`<ul style=${styleMap(listStyle)}>${virtualItems.map((virtualItem: VirtualItem) =>
+      html`<li key=${virtualItem.key} data-index=${virtualItem.index} ref=${ref(measureElement)}>
+        ${sentences[virtualItem.index]}</li>`)}</ul>`;
+  };
+
+  override render() {
+    return html`
+    <div style="height: 500px; width: 500px;">
+      <mdc-virtualizedlist
+        @on-scroll="${this.onScroll}"
```
cscheffauer, line 42: issue: the attribute in `virtualizedlist` is `onscroll`, but here you pass it in as @on-scroll, so this shouldn't technically work at all.
In general the suggestion is to follow as much native element behaviour as possible, so the naming should always be `onscroll` (or `@onscroll` when passing in).

packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
```diff
@@ -0,0 +1,105 @@
...
+      const virtualItems = getVirtualItems();
+      const listStyle: Readonly<StyleInfo> = {
+        position: 'absolute',
+        top: 0,
+        left: 0,
+        width: '100%',
+        transform: `translateY(${virtualItems[0]?.start ?? 0}px)`,
+      };
+
+      const event = new CustomEvent('render-list', {
```
cscheffauer, line 69: suggestion: instead of an event you could just set a callback function as a attribute on the component, maybe named `onRenderList` (attribute name `onrenderlist`) - and then here instead of dispatching an event you could do `this.onRenderList(...args)`.

packages/components/src/components/virtualizedlist/virtualizedlist.constants.ts
```diff
@@ -0,0 +1,8 @@
+import utils from '../../utils/tag-name';
+
+const TAG_NAME = utils.constructTagName('virtualizedlist');
+
+const DEFAULT_MEASURE_ELEMENT = () => 0;
```
cscheffauer, line 5: nitpick: you could group those 2 constants in a `DEFAULTS` object

packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
```diff
@@ -0,0 +1,105 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
cscheffauer, line 1: issue: e2e test is missing

cscheffauer (changes requested):

packages/components/src/components/virtualizedlist/virtualizedlist.test.component.ts
```diff
@@ -0,0 +1,72 @@
+import { CSSResult, PropertyValues, TemplateResult, html } from 'lit';
```
cscheffauer, line 1: suggestion: in terms of naming please follow the other components and name this file like in the iconprovider: `virtualizedlist.stories.utils.ts`

packages/components/src/components/virtualizedlist/virtualizedlist.e2e-test.ts
```diff
@@ -0,0 +1,52 @@
...
+    expect(await currentFirstElement.getAttribute('data-index')).toBe('0');
+
+    let currentLastElement = await componentsPage.page.getByRole('listitem').last();
+    expect(await currentLastElement.getAttribute('data-index')).not.toBe('99');
+
+    const scrollElement = componentsPage.page.locator('.mdc-virtualizedlist-scroll');
+    await scrollElement.waitFor();
+
+    // Scroll to the bottom of the container
+    await componentsPage.page.evaluate((element) => {
```
cscheffauer, line 38: suggestion: playwright has actually ways to do scrolling:
https://playwright.dev/docs/input#scrolling
i would suggest using this instead.

packages/components/src/components/virtualizedlist/virtualizedlist.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+import { disableControls } from '../../../config/storybook/utils';
+
+const render = (args: Args) => html`
+  <mdc-virtualizedwrapper
+    .virtualizerprops=${args.virtualizerprops}
+    .onscroll=${args.onscroll}
+  ></mdc-virtualizedwrapper>`;
+
+const meta: Meta = {
+  title: 'Work In Progress/virtualizedlist',
```
cscheffauer, line 14: issue: since this component will be released with this pr, also change this title to "Components"

packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * @internal
+   * Renders scroll element with ref and onscroll function if defined.
+   * If virtualizerController is not defined (possible during initial rendering),
+   * don't render virtualizedlist-container until it is.
+   *
+   * @returns The template result containing the scroll element or an empty template.
+   */
+  private getVirtualizedScrollElement(): TemplateResult {
```
cscheffauer, line 151: nitpick: this does not need to be a separate function, can be moved into the render. not too concerned about it though

cscheffauer (changes requested):

packages/components/src/components/virtualizedlist/virtualizedlist.test.component.ts
johnsoter13: I use this component in both the story and the e2e test. What should I name the file in that case?

cscheffauer: virtualizedlist.helper.test.ts

What u think?

package.json
```diff
@@ -70,5 +70,8 @@
     "syncpack": "^12.4.0",
     "typescript": "^4.8.4",
     "yargs": "^17.7.2"
+  },
+  "dependencies": {
+    "@tanstack/lit-virtual": "^3.11.3"
```
cscheffauer, line 75: issue: this should only be installed as a dependency in the package.json of the components subpackage, not on the root!

cscheffauer (changes requested):

cscheffauer (approved):

PR #1080, opened 2025-02-10
test: e2e test cases for input component

supminn:
### Description

* E2E test cases for input component.
* Updated snapshots for FormFieldWrapper based on code refactoring 

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-482


packages/components/src/components/input/input.e2e-test.ts
```diff
@@ -1,66 +1,364 @@
...
+      if (browserName !== 'firefox') {
+        await componentsPage.actionability.pressTab();
+        await expect(input).not.toBeFocused();
+        await expect(trailingButton).not.toBeFocused();
+      }
+      await componentsPage.removeAttribute(input, 'readonly');
+    });
 
-      // add additional tests here, like tabbing through several parts of the component
+    await test.step('focus on input and trailing button interactions', async () => {
```
shakti97, line 245: Suggestion: Can we separate the trailing button test? The trailing button appears when the input has a value and does not depend on whether the input is focusable, disabled, or readonly. 
Otherwise, we need to add a test case to check how it behaves in different scenarios.

shakti97 (approved):
lgtm

PR #1089, opened 2025-02-11
perf: extract type, add icon name mixin and update changed props

arjunkesava:
### Description

- The problem with `Iconnames` type is that if any component has a limited usage of specific icons,
then the build file still extracts all 5400+ icon names in the typescript version of build file.
- When the icon name is given as an input for a component,
then we still need to import all 5400+ icon name strings as the consumer can enter any icon name.
- Extract icon names from iconNames type.
- Add new icon name mixin, as the `icon-name` attribute is being used in 3 different components.
- Replace property values type to a better type and add the type import for lit.


packages/components/src/utils/mixins/IconMixin.ts
```diff
@@ -0,0 +1,24 @@
+import { LitElement } from 'lit';
```
cscheffauer, line 1: suggestion: the file name should reflect the "IconName" in the name i would say.

cscheffauer (commented):
just had a minor comment, not bothered it to be addressed or not.
Rest looks good, not gonna approve or request changes - was just curious :)

packages/components/src/components/buttonsimple/buttonsimple.component.ts
```diff
@@ -1,11 +1,12 @@
...
 import { property } from 'lit/decorators.js';
-import styles from './buttonsimple.styles';
 import { Component } from '../../models';
-import { ButtonSize, ButtonType } from './buttonsimple.types';
-import { BUTTON_TYPE, DEFAULTS } from './buttonsimple.constants';
 import { DisabledMixin } from '../../utils/mixins/DisabledMixin';
 import { TabIndexMixin } from '../../utils/mixins/TabIndexMixin';
+import { BUTTON_TYPE, DEFAULTS } from './buttonsimple.constants';
+import styles from './buttonsimple.styles';
+import { ButtonSize, ButtonType } from './buttonsimple.types';
```
supminn, line 9:
```suggestion
import type { ButtonSize, ButtonType } from './buttonsimple.types';
```

packages/components/src/components/divider/divider.component.ts
```diff
@@ -11,6 +11,7 @@ import {
   DIVIDER_VARIANT,
   TEXT_TAG,
 } from './divider.constants';
+import styles from './divider.styles';
 import { Directions, DividerOrientation, DividerVariant } from './divider.types';
```
supminn, line 15:
```suggestion
import type { Directions, DividerOrientation, DividerVariant } from './divider.types';
```

packages/components/src/components/link/link.component.ts
```diff
@@ -1,11 +1,12 @@
...
 import { property } from 'lit/decorators.js';
 import { Component } from '../../models';
+import { DisabledMixin } from '../../utils/mixins/DisabledMixin';
+import { IconNameMixin } from '../../utils/mixins/IconMixin';
 import { DEFAULTS, LINK_ICON_SIZES, LINK_SIZES } from './link.constants';
-import { LinkSize } from './link.types';
 import styles from './link.styles';
-import { IconNames } from '../icon/icon.types';
-import { DisabledMixin } from '../../utils/mixins/DisabledMixin';
+import { LinkSize } from './link.types';
```
supminn, line 9:
```suggestion
import type { LinkSize } from './link.types';
```

supminn (changes requested):
Type imports are missing in some places.

supminn (approved):

PR #1100, opened 2025-02-17
feat: release common form field group component

arjunkesava:
### Description

- Remove everything from checkbox group component.
- Build new form field group component.
- Move the header text, description text and data aria label mixin into field group component.
- Add new toggle group component which extends the form field group component.

### Links

-

### Screenshots

image


packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -0,0 +1,59 @@
...
+   */
+  @property({ type: String, attribute: 'header-text', reflect: true }) headerText?: string;
+
+  /**
+   * The description of the group.
+   */
+  @property({ type: String, attribute: 'description-text', reflect: true }) descriptionText?: string;
+
+  /** @internal */
+  @state() protected isRadio = false;
```
arjunkesava, line 31: For a radio based group, the role has to be set to `radiogroup` instead of `group` and to achieve that, we have this protected internal variable.

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -154,6 +148,7 @@ class Checkbox extends NameMixin(ValueMixin(DataAriaLabelMixin(FormfieldWrapper)
           .indeterminate="${this.indeterminate}"
           .disabled="${this.disabled}"
           aria-label="${this.dataAriaLabel ?? ''}"
+          aria-describedby="help-text"
```
cscheffauer, line 151: issue: this change needs to be made on every input, not only the checkbox i think.
Also this will break if someone changes the id in the formfieldwrapper - i would suggest to put an attribute on to the renderHelperText() method, which allows u to pass in the helperTextId, which u then can make a constant inside of here.

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -0,0 +1,59 @@
+import { CSSResult, html, nothing, TemplateResult } from 'lit';
+import { property, state } from 'lit/decorators.js';
+import { Component } from '../../models';
+import { DataAriaLabelMixin } from '../../utils/mixins/DataAriaLabelMixin';
+import styles from './formfieldgroup.styles';
+import type { TextType } from '../text/text.types';
+import { TYPE as TEXT_TYPE, VALID_TEXT_TAGS as TEXT_TAGS } from '../text/text.constants';
+
+/**
+ * `mdc-formfieldgroup` component, groups any list of components together.
```
cscheffauer, line 10: docs: it does not group any list of components together, only form fields.
Also mention that this group is specifically for creating a checkbox group & toggle group.
For the radiogroup use the RadioGroup component instead.

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -56,12 +56,6 @@ class Checkbox extends NameMixin(ValueMixin(DataAriaLabelMixin(FormfieldWrapper)
   /** @internal */
   static formAssociated = true;
```
cscheffauer, line 58: suggestion: in the TSDoc of this Component, we can mention that if u want to create a group of Checkboxes, use the FormFieldGroup component

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
cscheffauer: suggestion: this can be a protected property, no need to make it a state since it will not change during runtime.

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * The header and description text are displayed above the items with accessible labels.
+ * All items in this form field group have a gap of 12px (0.75rem) each.
+ *
+ * @tagname mdc-formfieldgroup
+ *
+ * @slot default - This is a default/unnamed slot
```
cscheffauer, line 17: docs: this can be described better what we expect the consumer to pass in.

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -0,0 +1,59 @@
...
+import { DataAriaLabelMixin } from '../../utils/mixins/DataAriaLabelMixin';
+import styles from './formfieldgroup.styles';
+import type { TextType } from '../text/text.types';
+import { TYPE as TEXT_TYPE, VALID_TEXT_TAGS as TEXT_TAGS } from '../text/text.constants';
+
+/**
+ * `mdc-formfieldgroup` component, groups any list of components together.
+ *
+ * The header and description text are displayed above the items with accessible labels.
+ * All items in this form field group have a gap of 12px (0.75rem) each.
```
cscheffauer, line 13: docs: i would recommend to rephrase this to "All passed in children will have a gap of ... applied".
Otherwise consumer might not exactly understand this information.

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Component } from '../../models';
+import { DataAriaLabelMixin } from '../../utils/mixins/DataAriaLabelMixin';
+import styles from './formfieldgroup.styles';
+import type { TextType } from '../text/text.types';
+import { TYPE as TEXT_TYPE, VALID_TEXT_TAGS as TEXT_TAGS } from '../text/text.constants';
+
+/**
+ * `mdc-formfieldgroup` component, groups any list of components together.
+ *
+ * The header and description text are displayed above the items with accessible labels.
```
cscheffauer, line 12: docs: you can mention all the accessibility details, like role="group" is set and the role is "aria-labelledby" the header text and "aria-describedby" the description text if passed in.

packages/components/src/components/formfieldgroup/formfieldgroup.stories.ts
```diff
@@ -0,0 +1,41 @@
...
+  },
+  argTypes: {
+    ...classArgType,
+    ...styleArgType,
+  },
+};
+
+export default meta;
+
+export const Example: StoryObj = {
```
cscheffauer, line 35: as mentioned above, as a replacement for the actual checkbox and toggle group, its recommended to create here additional stories, one for checkbox group and one for toggle group.

packages/components/src/components/formfieldwrapper/formfieldwrapper.component.ts
```diff
@@ -80,6 +80,7 @@ class FormfieldWrapper extends DisabledMixin(Component) {
     }
     return html`
       <mdc-text
+        id="help-text"
```
cscheffauer, line 83: suggestion: as mentioned i would recommend to pass the id into here as an attribute to the renderHelpText (u can always give the attribute a default). This will allow the classes extending this formfieldwrapper to decide about the id on their own.

packages/components/src/components/togglegroup/togglegroup.component.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * togglegroup component, which ...
+ *
+ * @tagname mdc-togglegroup
+ *
+ * @slot default - This is a default/unnamed slot
+ */
+class ToggleGroup extends FormfieldGroup {
+  public static override styles: Array<CSSResult> = [...FormfieldGroup.styles];
+}
```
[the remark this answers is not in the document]
cscheffauer: suggestion: i would vote for not having the checkboxgroup and togglegroup as separate components for now, but only keep the formfield, show the usage of checkbox and toggles inside the formfield and document it properly in the formfieldgroup that thats the component to use for building a checkbox and togglegroup. If there is a need for a special component, we can always do that later.

cscheffauer (changes requested):

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  @property({ type: String, attribute: 'description-text', reflect: true }) descriptionText?: string;
+
+  /**
+  * @internal
+  * This is used to set the role of the compoonent as `radiogroup` if this is true and to 'group' if it is false.
+  */
+  protected isRadio = false;
+
+  /** @internal */
+  private headerId = 'header-id';
```
cscheffauer, line 46: nitpick: instead of setting those properties on the class, we can also move them to the `.constants.ts` file and use them as constants.

packages/components/src/components/formfieldgroup/formfieldgroup.e2e-test.ts
```diff
@@ -0,0 +1,144 @@
...
+  await formfieldgroup.waitFor();
+  return formfieldgroup;
+};
+
+test('mdc-formfieldgroup', async ({ componentsPage }) => {
+  /**
+   * ACCESSIBILITY
+   */
+  await test.step('accessibility', async () => {
+    await componentsPage.accessibility.checkForA11yViolations('formfieldgroup-default');
```
cscheffauer, line 54: issue: this runs on an empty DOM! move this step after the visual regression block.

packages/components/src/components/formfieldgroup/formfieldgroup.e2e-test.ts
```diff
@@ -0,0 +1,144 @@
...
+    await test.step('attribute `description-text` should be present on component when set', async () => {
+      await componentsPage.setAttributes(formfieldgroup, { 'description-text': descriptionText });
+      const mdcText = componentsPage.page.locator('mdc-text[id=\'description-id\']');
+      const textContent = await mdcText.textContent();
+      expect(textContent?.trim()).toBe(descriptionText);
+    });
+
+    await test.step('attribute `data-aria-label` should be present on component when set', async () => {
+      await componentsPage.setAttributes(formfieldgroup, { 'data-aria-label': ariaLabel });
+      const divContent = await componentsPage.page.getByRole(ROLE.GROUP).getAttribute('aria-label');
```
cscheffauer, line 107: nitpick: `divContent` as the constant name does not make sense here

packages/components/src/components/formfieldgroup/formfieldgroup.stories.ts
```diff
@@ -0,0 +1,88 @@
...
+  <mdc-formfieldgroup
+    header-text="${args['header-text']}"
+    description-text="${args['description-text']}"
+    data-aria-label="${args['data-aria-label']}"
+  >
+    ${args.children}
+  </mdc-formfieldgroup>`;
+
+const meta: Meta = {
+  title: 'Work In Progress/formfieldgroup',
```
cscheffauer, line 18: suggestion: since this is adding everything, we can convert this PR to already release it - so here change it to Components and change the badges as well.
Also on the PR itself set the title to `feat: ...` and put a proper description there.

packages/components/src/components/formfieldwrapper/formfieldwrapper.component.ts
```diff
@@ -40,6 +40,13 @@ class FormfieldWrapper extends DisabledMixin(Component) {
    */
   @property({ type: String, reflect: true, attribute: 'help-text' }) helpText?: string;
 
+  /**
+   * @internal
+   * The unique id of the helper text field. It is used to link the helper text field with the input.
+   * @default `helper-text-id-${uuidv4()}`
+   */
+  protected helperTextId = `helper-text-id-${uuidv4()}`;
```
cscheffauer, line 48: issue: this is a property and will always fire uuidv4, even if there is no helpertext rendered.
Pass in the helperTextId into the `renderHelpText()` method in line 84 directly.

cscheffauer (changes requested):

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  @property({ type: String, attribute: 'header-text', reflect: true }) headerText?: string;
+
+  /**
+   * The description of the group.
+   */
+  @property({ type: String, attribute: 'description-text', reflect: true }) descriptionText?: string;
+
+  /**
+  * @internal
+  * This is used to set the role of the compoonent as `radiogroup` if this is true and to 'group' if it is false.
```
supminn, line 41:
```suggestion
  * This is used to set the role of the component as `radiogroup` if this is true and to 'group' if it is false.
```

typo

packages/components/src/components/formfieldgroup/formfieldgroup.stories.ts
```diff
@@ -0,0 +1,88 @@
...
+  },
+  argTypes: {
+    children: {
+      description: 'The html content which can be placed inside the formfieldgroup component.',
+      control: 'object',
+    },
+    ...classArgType,
+    ...styleArgType,
+  },
+};
```
supminn, line 33: header-text, description-text and data-aria-label are missing in argTypes definition

packages/components/src/components/formfieldgroup/formfieldgroup.stories.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export default meta;
+
+export const Example: StoryObj = {
+  args: {
+    'header-text': 'Header text',
+    'description-text': 'This is an example description text.',
+    'data-aria-label': 'Header text aria label',
```
supminn, line 41: question: when should the consumer set `data-aria-label`? I was of the understanding that it has to be set when `header-text` is not available. 
However, in these examples, both the header-text as well as the data-aria-label is set.

arjunkesava: Header text is visible on the screen.

Aria label can be a meaningful information provided by the consumer for screen readers.

We don't impose any rules on both of these attributes, both are optional and it is decided by the consumer to set them or not.

cscheffauer: @arjunkesava just to clarify, both are optional BUT the user has to provide at least the header-text or the aria-label, like one of them HAS to be passed in always, otherwise its not accessible. Maybe we should document this better.

cscheffauer (dismissed):

supminn (approved):

PR #1101, opened 2025-02-17
fix: decorator - grabber button - incorrect outline and surface color

kpankaj3:
### Description
1. Updated border color and background color for grabber button.
3. Added a margin between divider and grabber button so that the button doesn't cover the divider line as per design.

Snapshots updated.

### Links
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-471


deveshcisco (approved):
LGTM

PR #1102, opened 2025-02-19
chore: update form field group component

shakti97:
### Description
 Due to recent design update, header and description text is replaced by label, help text & spacing between them also changed.

### Links

Screenshot 2025-02-19 at 3 13 57 PM
Screenshot 2025-02-19 at 3 14 06 PM


packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -1,11 +1,9 @@
...
-import { property } from 'lit/decorators.js';
-import { Component } from '../../models';
+import { html } from 'lit';
 import { DataAriaLabelMixin } from '../../utils/mixins/DataAriaLabelMixin';
-import { VALID_TEXT_TAGS as TEXT_TAGS, TYPE as TEXT_TYPE } from '../text/text.constants';
-import type { TextType } from '../text/text.types';
-import { DEFAULTS, ROLE } from './formfieldgroup.constants';
+import FormfieldWrapper from '../formfieldwrapper';
+import { ROLE } from './formfieldgroup.constants';
+import { DEFAULTS } from '../formfieldwrapper/formfieldwrapper.constants';
```
arjunkesava, line 6:
```suggestion
import { DEFAULTS as FORMFIELD_DEFAULTS } from '../formfieldwrapper/formfieldwrapper.constants';
```

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -1,11 +1,9 @@
 import type { CSSResult } from 'lit';
-import { html, nothing, TemplateResult } from 'lit';
-import { property } from 'lit/decorators.js';
-import { Component } from '../../models';
+import { html } from 'lit';
 import { DataAriaLabelMixin } from '../../utils/mixins/DataAriaLabelMixin';
-import { VALID_TEXT_TAGS as TEXT_TAGS, TYPE as TEXT_TYPE } from '../text/text.constants';
-import type { TextType } from '../text/text.types';
-import { DEFAULTS, ROLE } from './formfieldgroup.constants';
+import FormfieldWrapper from '../formfieldwrapper';
```
arjunkesava, line 4:
```suggestion
import FormfieldWrapper from '../formfieldwrapper/formfieldwrapper.component';
```

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -28,47 +26,30 @@ import styles from './formfieldgroup.styles';
...
-   * The header text of the group.
-   */
-  @property({ type: String, attribute: 'header-text', reflect: true }) headerText?: string;
-
-  /**
-   * The description of the group.
-   */
-  @property({ type: String, attribute: 'description-text', reflect: true }) descriptionText?: string;
-
+class FormfieldGroup extends DataAriaLabelMixin(FormfieldWrapper) {
```
arjunkesava, line 29: As we now extend ForfieldWrapper component, we need to set help-text-type to be undefined in the constructor.

https://github.com/momentum-design/momentum-design/blob/6d5119ff309d9574fd460e23442d369a162f83d2/packages/components/src/components/checkbox/checkbox.component.ts#L72

packages/components/src/components/formfieldgroup/formfieldgroup.styles.ts
```diff
@@ -1,13 +1,31 @@
...
+    & > [part="label-text"] {
+      margin-bottom: 0.25rem;
+    }
+    & > [part="help-text"] {
+      margin-bottom: 0.75rem;
+    }
+    & > ::slotted(*) {
+      margin-bottom: 0.75rem;
+    }
+  }
```
arjunkesava, line 23: Is there any better way to achieve this?

arjunkesava (commented):
Will review more tmrw.

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
shakti97: Done

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
shakti97: Done

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
shakti97: Done

packages/components/src/components/formfieldgroup/formfieldgroup.styles.ts
shakti97: I have added comments for better understanding
For better readability, I can separate the selectors, but that would increase the use of :has(), which might impact performance. Currently, they are nested to first check for the existence of the help text before adjusting the spacing.

Another approach is to use a negative margin. This requires setting a gap for the flex container and applying a negative margin to reduce the space between the label and help text. However, I'm unsure if using negative margins is the best practice.

packages/components/src/components/formfieldgroup/formfieldgroup.styles.ts
```diff
@@ -1,13 +1,41 @@
...
+    * Adjust the gap between the help text and the slot element.
+    */
+    & > [part="help-text"],
+    & > ::slotted(*) {
+      margin-bottom: 0.75rem;
+    }
+  }
+
+  :host::part(label-text){
+    font-weight: 700;
```
arjunkesava, line 34:
```suggestion
    font-weight: bold;
```

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
```diff
@@ -28,47 +27,35 @@ import styles from './formfieldgroup.styles';
...
-        aria-labelledby="${DEFAULTS.HEADER_ID}"
-        aria-describedby="${DEFAULTS.DESCRIPTION_ID}"
+        aria-labelledby="${FORMFIELD_DEFAULTS.LABEL_ID}"
+        aria-describedby="${FORMFIELD_DEFAULTS.HELPER_TEXT_ID}"
         aria-label="${this.dataAriaLabel ?? ''}"
       >
-        ${this.renderText(TEXT_TYPE.BODY_LARGE_BOLD, DEFAULTS.HEADER_ID, 'header', this.headerText)}
-        ${this.renderText(TEXT_TYPE.BODY_LARGE_REGULAR, DEFAULTS.DESCRIPTION_ID, 'description', this.descriptionText)}
+        ${this.renderLabel()}
+        ${this.renderHelpText()}
```
arjunkesava, line 52:
```suggestion
        ${this.renderHelperText()}
```

packages/components/src/components/formfieldgroup/formfieldgroup.styles.ts
```diff
@@ -1,13 +1,41 @@
...
+  }
+
+  :host::part(help-text) {
+    color: var(--mdc-group-help-text-normal);
   }
```
arjunkesava, line 39:
```suggestion
```

This is not needed as the formfieldwrapper also has the same color.
If you change the render call for help text (this.renderHelperText()), then you don't need this color declaration at all.

packages/components/src/components/formfieldgroup/formfieldgroup.styles.ts
```diff
@@ -1,13 +1,41 @@
...
+      margin-bottom: 0.25rem;
+    }
+    /*
+    * Adjust the gap between the help text and the slot element.
+    */
+    & > [part="help-text"],
+    & > ::slotted(*) {
+      margin-bottom: 0.75rem;
+    }
+  }
```
arjunkesava, line 31: We can remove this entire css by simply creating a head part.
In HTML, create a new div only for head part so that we can keep a gap of 4px in between them.
```html
<div part="group-header">
    ${this.renderLabel()}
    ${this.renderHelperText()}
</div>
```

And in `formfieldwrapper.component.ts` file,
```javascript
protected renderHelperText() {
+   if (!this.helpText) {
+      return nothing;
+   }

return html`<div class="mdc-help-text" part="mdc-help-text">
```

In the css file, you can simply set like this instead of margin bottoms on each item.
```css
  :host::part(group-header) {
    display: flex;
    flex-direction: column;
    gap: 0.25rem;
  }
```

arjunkesava (changes requested):

packages/components/src/components/formfieldgroup/formfieldgroup.styles.ts
shakti97: Done

packages/components/src/components/formfieldgroup/formfieldgroup.component.ts
shakti97: Done

packages/components/src/components/formfieldgroup/formfieldgroup.styles.ts
shakti97: Done

packages/components/src/components/formfieldgroup/formfieldgroup.styles.ts
shakti97: oh, yes we can do that. It didn't strike to me. Thanks.

arjunkesava (approved):

PR #1105, opened 2025-02-21
chore: added web api cache to iconprovider

cscheffauer:
### Description

- Allowing web-api cache to be used as an option in icon provider
- Setup Storybook to be able to show both options
- Added e2e test for web-api-cache


packages/components/src/components/iconprovider/iconprovider.component.ts
```diff
@@ -66,13 +70,32 @@ class IconProvider extends Provider<IconProviderContext> {
...
+   * Choose `js-cache` to cache icons in a JS cache (in-memory cache).
+   * Choose `web-cache-api` to cache icons using the Web Cache API.
+   *
+   * NOTE: `cache-name` must be provided if `cache-strategy` is provided.
+   *
+   * If not provided or invalid value provided, the icons will not be cached.
+   * @default undefined
+   */
+  @property({ type: String, attribute: 'cache-strategy' })
+  cacheStrategy?: 'js-cache' | 'web-cache-api';
```
maxinteger, line 84: I think `in-memory-cache` or just `memory-cache` would be better name for the `js-cache`.

`web-cache-api` could be renamed for `cache-storage` but it might be too low level for this interface so `web-cache-api` is fine.

packages/components/src/components/icon/icon.utils.ts
```diff
@@ -9,22 +18,80 @@
...
+      }
+      return response.text();
+    });
   }
 
-  return response.text();
+  return webAPIIconsCache(cacheName).then((iconsCache) =>
+    iconsCache
+      .get(request)
+      .then(async (responseText) => {
```
maxinteger, line 50: please do not mix `.then` and `async/await` code in the same function, same in line 65. It is bad for readability and following the code flow.

Also there is more nesting then actually necessary. I would recommend `async/await`, but `.then` fine as well, just do not mix and nest each `.then`

packages/components/src/components/icon/icon.component.ts
```diff
@@ -126,15 +127,18 @@ class Icon extends Component {
...
-      const { fileExtension, url, iconsCache, shouldCache } = this.iconProviderContext.value;
+      const { fileExtension, url, cacheName, cacheStrategy } = this.iconProviderContext.value;
       if (url && fileExtension && this.name) {
         // abort the previous fetch request if it is still pending
         // before retreiving from cache
         this.abortController.abort();
 
+        const iconsCache = cacheName && cacheStrategy === 'js-cache' ? jsIconsCache(cacheName) : undefined;
+
+        const iconData = iconsCache?.get(this.name);
```
maxinteger, line 138: you need to use `?.` for the method call as well otherwise JS will try to call undefined because 

`iconsCache?.get()` means `(iconsCache ? iconsCache.get : undefined)()`

```suggestion
        const iconData = iconsCache?.get?.(this.name);
```

packages/components/src/components/icon/icon.component.ts
```diff
@@ -143,16 +147,22 @@ class Icon extends Component {
...
           const iconElement = this.prepareIconElement(iconData);
 
           this.handleIconLoadedSuccess(iconElement as HTMLElement);
-          if (shouldCache) {
-            // store the fetched icon string in the iconsCache map
-            iconsCache.set(this.name, iconData);
-          }
+
+          // store the fetched icon string in the iconsCache map
+          iconsCache?.set(this.name, iconData);
```
maxinteger, line 165: same as above
```suggestion
          iconsCache?.set?.(this.name, iconData);
```

packages/components/src/components/icon/icon.component.ts
```diff
@@ -126,15 +127,18 @@ class Icon extends Component {
...
   private async getIconData() {
     if (this.iconProviderContext.value) {
-      const { fileExtension, url, iconsCache, shouldCache } = this.iconProviderContext.value;
+      const { fileExtension, url, cacheName, cacheStrategy } = this.iconProviderContext.value;
       if (url && fileExtension && this.name) {
         // abort the previous fetch request if it is still pending
         // before retreiving from cache
         this.abortController.abort();
 
+        const iconsCache = cacheName && cacheStrategy === 'js-cache' ? jsIconsCache(cacheName) : undefined;
```
maxinteger, line 136: I think you should do caching the same place either outside of `dynamicSVGImport` or inside of `dynamicSVGImport`, but not both (one here and the other one in `dynamicSVGImport`). Caching should be close to the actual fetching so I recommend to do both inside `dynamicSVGImport`

packages/components/src/utils/icon-cache/index.ts
```diff
@@ -0,0 +1,33 @@
+type IconCache = Map<string, string>;
+type Registry = Record<string, IconCache>;
+
+const registry: Registry = {};
+
+export const jsIconsCache = (name: string) => {
```
maxinteger, line 6: The 2 strategies should have the same interface:

```ts
type IconCache = Map<string, string>;
type Registry = Map<string, IconCache>;

type CacheStrategy = (bucketName: string) => Promise<CacheStrategyBucket>;
interface CacheStrategyBucket {
  put: (request: Request, response: Response) => Promise<void>;
  get: (request: Request) => Promise<string>;
  delete: (request: Request) => Promise<void>;
}

const registry: Registry = new Map();

const memoryStrategy: CacheStrategy = async (bucketName: string) => {
  if (!registry.has(bucketName)) {
    registry.set(bucketName, new Map<string, string>());
  }
  const cache = registry.get(bucketName) as IconCache;
  
  return {
    async put(request: Request, response: Response) {
      cache.set(request.url, await response.text());
    },
    async get(request: Request) {
      return cache.get(request.url) || '';
    },
    async delete(request: Request) {
      cache.delete(request.url);
    },
  };
}

const cacheStorageStrategy: CacheStrategy = async (bucketName: string) => {
  const cache = await caches.open(bucketName);
  return {
    async put(request: Request, response: Response) {
      await cache.put(request, response);
    },
    async get(request: Request) {
      const response = await cache.match(request);
      return response ? await response.text() : '';
    },
    async delete(request: Request) {
      await cache.delete(request);
    },
  };
}
```

Is there a reason to give 2 options for caching for the consumer?

If not, then we should have 1 caching which can internally fallback to in memory register when the Cache storage api is not available. It would simplify all other implementation.

```ts
const cacheStrategy = async (bucketName: string) => {
  if ('caches' in globalThis) {
    return cacheStorageStrategy(bucketName);
  }
  return memoryStrategy(bucketName);
}
```

cscheffauer: i changed the put to set in both places.
I will not do the fallback, since we do want to consumer to choose either or, without any fallback.
And the reason to have them separate is that the webAPI one is async, where the inMemory one is sync. Mixing them up in one strategy did not make sense to me since they function differently (sync vs async)

PR #1113, opened 2025-02-25
fix: reflect spinner variant property and export types

jor-row:
### Description

When using the react component for ```spinner``` changing the variant was not reflected in the styling as it was not being reflected back to the associated attribute. This PR fixes that.

Also, export Spinner types to be used in MRV2.


cscheffauer (dismissed):
lgtm!

cscheffauer (approved):

PR #1103, opened 2025-02-19
refactor: form internals and validation using mixin

supminn:
### Description

- Created a new mixin for form internals and input validation
- Updated form related components with this mixin. 

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-537

PR #1116, opened 2025-02-27
fix: form value issue

shakti97:
### Description
Fix form value issue in radio btn used inside form & fix one accessibility issue with regarding aria-described


supminn:
Incorporated these changes on https://github.com/momentum-design/momentum-design/pull/1103/commits/8cdac5147d61c0e76a895281315a446cd04ed403
This PR is marked as closed.

PR #1121, opened 2025-02-28
fix: style for toggle component

supminn:
### Description

There were extra spaces on the toggle component that wasn't it align in center when added on list item. This PR handles the fixes.

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-559


packages/components/src/components/toggle/toggle.component.ts
```diff
@@ -170,8 +170,8 @@ class Toggle extends NameMixin(ValueMixin(DataAriaLabelMixin(FormfieldWrapper)))
             ></mdc-icon>
           </div>
         </div>
-        ${this.renderLabel()}
-        ${this.renderHelperText()}
+        ${this.label ? this.renderLabel() : nothing}
```
cscheffauer, line 173: why is this necessary? the formfieldwrapper does this already...

cscheffauer (changes requested):

packages/components/src/components/toggle/toggle.component.ts
supminn: I have reverted this change and updated on the formfieldwrapper.
It was adding an extra div wrapping the label text, required label and the info slot.

arjunkesava (changes requested):
Instead of making the same change in 3 components, how about we check the label and help text on the formfieldwrapper component itself?

```diff
  protected renderLabel() {
+  if (!this.label) return nothing;
    return html`<div class="mdc-label-text" part="label-text">
      <slot name="label">${this.renderLabelElement()}</slot>
      <slot name="required-label">${this.renderRequiredLabel()}</slot>
      <slot name="label-info"></slot>
    </div>`;
  }
  ```

or similar to the way help text is done:
https://github.com/momentum-design/momentum-design/blob/03c16fcb3d254275c58d1816f4d202af54a6d3b8/packages/components/src/components/formfieldwrapper/formfieldwrapper.component.ts#L138-L146

**Review by cscheffauer — DISMISSED** [review]

File: packages/components/src/components/checkbox/checkbox.component.ts [hunk 2]
```diff
@@ -162,6 +162,14 @@ class Checkbox extends FormInternalsMixin(DataAriaLabelMixin(FormfieldWrapper))
     }
   }
 
+  private renderLabelAndHelperText = () => {
+    if (!this.label) return nothing;
```
arjunkesava, line 166:
```suggestion
  private renderLabelAndHelperText(): void {
```

arjunkesava (approved):

PR #1129, opened 2025-03-06
refactor: random generation of id added individually

supminn:
### Description

Updated the mixin to remove generating random id and setting it on the attribute
Updated individual components to do the same.

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-567-


Jason20020 (approved):
LGTM

PR #1130, opened 2025-03-06
chore: Tablist Component

deveshcisco:
### Description

Add the Tablist Component

Example usage:

```html
<mdc-tablist aria-label="Media types" role="tablist" activetabid="">
    <mdc-tab text="Photos" icon-name="add-photo-bold" tabid="photos-tab" aria-controls="photos-panel" variant="line">
    </mdc-tab>
    <mdc-tab text="Videos" icon-name="video-bold-regular" tabid="videos-tab" aria-controls="videos-panel" variant="line">
      <mdc-badge slot="badge" type="counter" counter="5"></mdc-badge>
    </mdc-tab>
    <mdc-tab text="Music" icon-name="file-music-bold" tabid="music-tab" aria-controls="music-panel" variant="line">
    </mdc-tab>
    <mdc-tab text="Documents" icon-name="document-bold" tabid="documents-tab" aria-controls="documents-panel" variant="line">
    </mdc-tab>
    <mdc-tab text="Downloads" icon-name="cloud-download-bold" tabid="downloads-tab" aria-controls="downloads-panel" variant="line">
    </mdc-tab>
</mdc-tablist>
```


packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+
+  /**
+   * @internal
+   */
+  private scrollDistance: number = 100;
+
+  constructor() {
+    super();
+    this.role = 'tablist';
+    this.attachShadow({ mode: 'open' });
```
cscheffauer, line 72: q: why is this necessary?

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+
+    /**
+     * Sets the initial active tab.
+     */
+    if (typeof this.activeTabId === 'string' && this.activeTabId.length === 0) {
+      this.activeTabId = this.tabs[0].getAttribute('tabid') || '';
+    }
+    this.selectTab(this.tabs.find((tab) => tab.getAttribute('tabid') === this.activeTabId) || this.tabs[0]);
+
+    this.scrollDistance = remToPx(6.25);
```
cscheffauer, line 94: q: why is this necessary? whenever the component is connected we set the scrollDistance to 6.25 rem? why? where is that number coming from?

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+     * Then show or hide the arrow buttons accordingly.
+     *
+     * @param entries - ResizeObserverEntry[].
+     */
+    const resizeObserver = new ResizeObserver((entries): void => {
+      const tabListWidth: number = entries[0].borderBoxSize[0].inlineSize;
+      const tabsContainerWidth: number = this.tabsContainer.scrollWidth || 0;
+
+      if (this.isRtl) {
+        this.leftArrowButton.style.right = `${tabListWidth - remToPx(1)}px`;
```
cscheffauer, line 108: issue: why are the buttons positioned absolute? you could just do a flex box and do put them at the start and end with `flex-start` / `flex-end` - this calculation of the left and right arrow button doesn't really make sense to me.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+  constructor() {
+    super();
+    this.role = 'tablist';
+    this.attachShadow({ mode: 'open' });
+
+    this.tabs = this.allTabs;
+
+    this.addEventListener('keydown', this.handleKeydown.bind(this));
+    this.addEventListener('click', this.handleClick.bind(this));
+    this.addEventListener('scroll', this.handleScroll.bind(this));
```
cscheffauer, line 78: issue: there is an eventlistener on the host here and also on the tabscontainer in line 130, so there are 2 scroll event listeners for no reason imho. this is a performance issue, we should not add event listeners which are not necessary.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+   * @param changedProperties - Map of changed properties with old values.
+   */
+  public override update(changedProperties: PropertyValues): void {
+    super.update(changedProperties);
+    if (changedProperties.has('activeTabId')) {
+      this.selectTab(this.tabs.find((tab) => tab.getAttribute('tabid') === this.activeTabId) || this.tabs[0]);
+    }
+  }
+
+  private get allTabs(): Tab[] {
```
cscheffauer, line 145: suggestion: this could be converted to the queryAll directive, see here:
https://lit.dev/docs/api/decorators/#queryAll
will help us to reduce the amount of code.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+  private get firstTab(): Tab { return this.tabs[0]; }
+
+  private get lastTab(): Tab { return this.tabs[this.tabs.length - 1]; }
+
+  private get nextTab(): Tab {
+    const newIndex = this.tabs.findIndex((tab) => document.activeElement === tab) + 1;
+    return this.tabs[newIndex % this.tabs.length];
+  }
+
+  private get tabsContainer(): HTMLDivElement {
```
cscheffauer, line 163: suggestion: all the `shadowRoot.querySelector` getters here could be replaced with query directive:
https://lit.dev/docs/api/decorators/#query

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+
+  /**
+   * @internal
+   */
+  private tabs: Array<Tab> = [];
+
+  /**
+   * @internal
+   */
+  private direction: String = '';
```
cscheffauer, line 62: nitpick: instead of saving the direction you could just save `isRTL`, like there are 2 state / properties for the same thing.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+   * Handle the keydown event. The arrow keys, Home, End, Enter, and Space keys are supported.
+   *
+   * @param event - HTML Keyboard Event.
+   */
+  private handleKeydown = (event: KeyboardEvent): void => {
+    const tab = event.target;
+
+    if (!(tab instanceof Tab)) { return; }
+
+    if (event.altKey) { return; }
```
cscheffauer, line 230: q: why is this line necessary?

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+
+    let newTab: Tab | undefined;
+
+    switch (event.code) {
+      case KEYCODES.ENTER:
+        this.selectTab(tab);
+        break;
+
+      case KEYCODES.SPACE:
+        if (event.target !== document.body) {
```
cscheffauer, line 240: q: how can the event target ever be body? the event listener is here on the host, so how could the target ever be the body? why is this line necessary?

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+          event.preventDefault();
```
cscheffauer, line 241: suggestion: the space has preventDefault, Enter has not - please explain with a comment why that is. Someone might get to this later and might think its a bug (not having preventDefault on both enter & space)

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+          this.selectTab(tab);
+        }
+        break;
+
+      case KEYCODES.LEFT:
+        if (this.isRtl) {
+          newTab = this.nextTab;
+        } else newTab = this.previousTab;
```
cscheffauer, line 249: nitpick: this syntax is not great - we either do both if & else both inline, or both with `{}` - a mix is not recommended and makes it less readable. 
same for the occurrences below, if u fix it please fix all occurrences.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+        break;
+
+      case KEYCODES.RIGHT:
+
```
cscheffauer, line 253: nitpick: unnecessary line break

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+      case KEYCODES.HOME:
+        newTab = this.firstTab;
+        break;
+
+      case KEYCODES.END:
+        newTab = this.lastTab;
+        break;
+
+      case KEYCODES.TAB:
+        event.preventDefault();
```
cscheffauer, line 268: suggestion: same as above, please explain why preventDefault is necessary here.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+   *
+   * @param event - HTML Mouse Event.
+   */
+  private handleClick = (event: MouseEvent): void => {
+    const tab = event.target;
+
+    if (!(tab instanceof Tab)) { return; }
+
+    this.selectTab(tab);
+    this.focusTab(tab);
```
cscheffauer, line 306: issue: this should be unnecessary, clicking on the tab should automatically focus it.

packages/components/src/components/tablist/tablist.styles.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  :host .tabs_container {
+    display: flex;
+    overflow-x: hidden;
+    gap: var(--mdc-tablist-gap);
+    padding: var(--mdc-focus-ring-outer-width);
+    margin: calc(var(--mdc-focus-ring-outer-width) * -1);
+  }
+
+  :host .hide-button {
```
cscheffauer, line 38: q: is there a reason why to do display none here? can we not just conditionally render the buttons in the DOM? then you wouldn't need the whole classlist add / remove thing.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+   * @param arrowButton - The arrow button to check.
+   */
+  private isArrowButtonHidden = (arrowButton: Button): boolean => arrowButton.classList.contains('hide-button');
+
+  /**
+   * Show the left arrow button.
+   */
+  private showLeftArrowButton = (): void => {
+    this.leftArrowButton.classList.remove('hide-button');
+    this.classList.add('show-left-arrow-button-padding');
```
cscheffauer, line 328: suggestion: instead of conditionally adding padding, just conditionally render the buttons and apply padding to the buttons themself - so you would not need that at all.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+  };
+
+  public override render() {
+    return html`
+      <mdc-button 
+        variant="tertiary"
+        size="32"
+        color="default" 
+        prefix-icon="arrow-left-regular" 
+        postfix-icon=""
```
cscheffauer, line 418: nitpick: this does not need to be set

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+
+  public override render() {
+    return html`
+      <mdc-button 
+        variant="tertiary"
+        size="32"
+        color="default" 
+        prefix-icon="arrow-left-regular" 
+        postfix-icon="" 
+        type="button"
```
cscheffauer, line 419: nitpick: this does not need to be set

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+  public override render() {
+    return html`
+      <mdc-button 
+        variant="tertiary"
+        size="32"
+        color="default" 
+        prefix-icon="arrow-left-regular" 
+        postfix-icon="" 
+        type="button" 
+        role="button"
```
cscheffauer, line 420: nitpick: this does not need to be set

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+    return html`
+      <mdc-button 
+        variant="tertiary"
+        size="32"
+        color="default" 
+        prefix-icon="arrow-left-regular" 
+        postfix-icon="" 
+        type="button" 
+        role="button" 
+        tabindex="0"
```
cscheffauer, line 421: nitpick: this does not need to be set

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+      behavior: 'instant',
+    });
+  };
+
+  public override render() {
+    return html`
+      <mdc-button 
+        variant="tertiary"
+        size="32"
+        color="default"
```
cscheffauer, line 416: nitpick: this does not need to be set

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+        aria-label="Scroll Tabs Left"
+        @click="${() => this.scrollTabsLeft()}"
+        class="hide-button"></mdc-button>
+      <div class="tabs_container">
+        <slot></slot>
+      </div>
+      <mdc-button 
+        variant="tertiary"
+        size="32"
+        color="default"
```
cscheffauer, line 431: nitpick: this does not need to be set

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+        class="hide-button"></mdc-button>
+      <div class="tabs_container">
+        <slot></slot>
+      </div>
+      <mdc-button 
+        variant="tertiary"
+        size="32"
+        color="default" 
+        prefix-icon="arrow-right-regular" 
+        postfix-icon=""
```
cscheffauer, line 433: nitpick: this does not need to be set

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+      <div class="tabs_container">
+        <slot></slot>
+      </div>
+      <mdc-button 
+        variant="tertiary"
+        size="32"
+        color="default" 
+        prefix-icon="arrow-right-regular" 
+        postfix-icon="" 
+        type="button"
```
cscheffauer, line 434: nitpick: this does not need to be set

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+        <slot></slot>
+      </div>
+      <mdc-button 
+        variant="tertiary"
+        size="32"
+        color="default" 
+        prefix-icon="arrow-right-regular" 
+        postfix-icon="" 
+        type="button" 
+        role="button"
```
cscheffauer, line 435: nitpick: this does not need to be set

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+      </div>
+      <mdc-button 
+        variant="tertiary"
+        size="32"
+        color="default" 
+        prefix-icon="arrow-right-regular" 
+        postfix-icon="" 
+        type="button" 
+        role="button" 
+        tabindex="0"
```
cscheffauer, line 436: nitpick: this does not need to be set

packages/components/src/components/tablist/tablist.constants.ts
```diff
@@ -0,0 +1,17 @@
...
+  ENTER: 'Enter',
+  SPACE: 'Space',
+  RIGHT: 'ArrowRight',
+  LEFT: 'ArrowLeft',
+  END: 'End',
+  HOME: 'Home',
+  TAB: 'Tab',
+};
+
+const SCROLL_BUTTON_OFFSET = 15;
```
cscheffauer, line 15: q: where is 15 coming from? are those px, rem?

packages/components/src/components/tablist/tablist.styles.ts
```diff
@@ -0,0 +1,49 @@
+import { css } from 'lit';
+
+const styles = css`
+  :host {
+    --mdc-focus-ring-inner-width: 0.125rem;
```
cscheffauer, line 5: issue: you are overriding the size of the focus ring, we should not do that unless its in the design specs

packages/components/src/components/tablist/tablist.styles.ts
```diff
@@ -0,0 +1,49 @@
...
+    --mdc-focus-ring-middle-width: calc(2 * var(--mdc-focus-ring-inner-width));
```
cscheffauer, line 6: issue: you are overriding the size of the focus ring, we should not do that unless its in the design specs

packages/components/src/components/tablist/tablist.styles.ts
```diff
@@ -0,0 +1,49 @@
...
+    --mdc-focus-ring-outer-width: calc(0.0625rem + var(--mdc-focus-ring-middle-width));
```
cscheffauer, line 7: issue: you are overriding the size of the focus ring, we should not do that unless its in the design specs

packages/components/src/components/tablist/tablist.utils.ts
```diff
@@ -0,0 +1,5 @@
+export const remToPx = (remValue: number): number => {
```
cscheffauer, line 1: suggestion: ideally we would not need this at all, like calculating between rem and px should not be necessary at all. maybe u can have a think about it - but this has some issues with it cause we are getting the documents font-size for every calculation we do seems complex and unnecessary to me.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,446 @@
...
+ *
+ * Children of the tab list are mdc-tab elements, sent to the default slot.
+ *
+ * The tabs can be navigated using the arrow keys, and selected by clicking, or pressing the Enter and Space keys.
+ *
+ * **Accessibility notes for consuming**
+ *
+ * - The element that serves as the container for the set of tabs has role `tablist`.
+ * - Each element that serves as a tab has role `tab` and is contained within the element with role `tablist`.
+ * - Each element that contains the `content panel` for a `tab` has role `tabpanel`.
```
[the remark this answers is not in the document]
deveshcisco: Separated into 2 sections

packages/components/src/components/tablist/tablist.component.ts
cscheffauer: why do we need to set a scrollDistance?

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed the first listener

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed the instances of shadowRoot for two elements and replaced the first one with query

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Moved it to a getter

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: This is to detect if the alt key is pressed, which is in the case a screen reader is active. However looking into it a bit more, I've removed it.

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: This was placed because the space button was also scrolling the page, have replaced it with preventDefault now, it also blocks the scrolling of the page

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Added prevent default to both buttons

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Refactored the function calls to use ternary operators

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Added explainer comment

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: This is needed as the function also scrolls the selected tab into view. However, I've added a check to prevent the focus() call if the element is already in focus

packages/components/src/components/tablist/tablist.styles.ts
deveshcisco: Yes, removed the logic and replaced it with a render function

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Fixed, using the css selectors now to apply the margins

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Removed

packages/components/src/components/tablist/tablist.constants.ts
deveshcisco: It is an old artefact, no longer needed. Was left in the code, now removed it

packages/components/src/components/tablist/tablist.styles.ts
deveshcisco: Renamed to use the mdc-tablist prefix

packages/components/src/components/tablist/tablist.styles.ts
deveshcisco: Renamed to use the mdc-tablist prefix

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: I took a look at this, and have replaced the logic and now we scroll to the end of the visible edge of the tabs

packages/components/src/components/tablist/tablist.utils.ts
deveshcisco: Added this calculation in case the user increases the browser font on the fly, to prevent the need for an observer.

However, this is now removed as we now use the element width to get the scroll distance

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,413 @@
...
+  private get isRtl(): boolean {
+    return (
+      document.querySelector('html')?.getAttribute('dir') === 'rtl' || window.getComputedStyle(this).direction === 'rtl'
+    );
+  }
+
+  constructor() {
+    super();
+
+    this.addEventListener('keydown', this.handleKeydown.bind(this));
```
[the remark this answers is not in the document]
deveshcisco: Correct, we don't need to bind a class field function. Thanks for pointing this out, I've removed the binding

packages/components/src/components/tablist/tablist.stories.ts
```diff
@@ -0,0 +1,158 @@
...
+      'tabvariant',
+      'currentTab',
+      'showArrowButtons',
+      'handleFocus',
+      'scrollTabs',
+      'showForwardArrowButton',
+      'showBackwardArrowButton',
+      'setActiveTab',
+      'currentActiveTab',
+      'scrollTabIntoView',
```
supminn, line 133: question: From where is this all coming? I guess something is off here. We shouldn't have so many reflective properties that should be hidden

packages/components/src/components/tablist/tablist.stories.ts
```diff
@@ -0,0 +1,158 @@
...
+    badges: ['wip'],
+  },
+  argTypes: {
+    activetabid: {
+      control: 'text',
+      description: 'ID of the active tab, defaults to the first tab if not provided',
+    },
+    onchange: {
+      description: `Function that gets called when user changes the active tab.
+      The function will receive the fired event and the new tab id can be fetched from event.detail.tabId.
```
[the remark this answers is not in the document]
supminn: This description should be moved into the components file's ts-doc under @event.

packages/components/src/components/tablist/tablist.stories.ts
```diff
@@ -0,0 +1,158 @@
+import type { Meta, StoryObj, Args } from '@storybook/web-components';
+import { action } from '@storybook/addon-actions';
+import '.';
+import { html } from 'lit';
+import { disableControls, hideControls } from '../../../config/storybook/utils';
+import { TAB_VARIANTS } from '../tab/tab.constants';
+import '../badge';
+
+const render = (args: Args) => html`
+<mdc-tablist @change="${action('onchange')}" activetabid="${args.activetabid}" aria-label="Media types">
```
supminn, line 10: suggestion: Add events of keyup and keydown as actions. 
This would reflect when the user moves between tabs

packages/components/src/components/tablist/tablist.stories.ts
```diff
@@ -0,0 +1,158 @@
...
+  <div id="music-panel" role="tabpanel" aria-labelledby="music-tab" hidden>
+    <p>Music panel</p>
+  </div>
+  <div id="documents-panel" role="tabpanel" aria-labelledby="documents-tab" hidden>
+    <p>Documents panel</p>
+  </div>
+  <div id="downloads-panel" role="tabpanel" aria-labelledby="downloads-tab" hidden>
+    <p>Downloads panel</p>
+  </div>
+<!-- End of example markup for the tab panels -->`;
```
supminn, line 64: question: Why do we require this?

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,417 @@
...
+ *
+ * @cssproperty --mdc-tablist-gap - Gap between tabs
+ * @cssproperty --mdc-tablist-width - Width of the tablist
+ * @cssproperty --mdc-container-button-padding - Padding for when an arrow button is visible
+ */
+class TabList extends Component {
+  /**
+   * ID of the active tab, defaults to the first tab if not provided
+   */
+  @property({ type: String, attribute: 'activetabid', reflect: true })
```
supminn, line 56:
```suggestion
  @property({ type: String, attribute: 'active-tabid', reflect: true })
```
Keeping it consistent with HTML attribute naming.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,417 @@
...
+   * @internal
+   */
+  @state() private showForwardArrowButton: boolean = false;
+
+  /**
+   * @internal
+   */
+  @state() private showBackwardArrowButton: boolean = false;
+
+  private get isRtl(): boolean {
```
supminn, line 82: question: why should it be a getter and not a normal function?

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,417 @@
...
+   *
+   * Fire the tab change event.
+   *
+   * @param changedProperties - Map of changed properties with old values.
+   */
+  public override update(changedProperties: PropertyValues<this>): void {
+    super.update(changedProperties);
+
+    if (changedProperties.has('activeTabId')) {
+      if (!this.tabs || !this.activeTabId) { return; }
```
supminn, line 149:
```suggestion
      if (!this.tabs || !this.activeTabId) return
```

nitpick

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,417 @@
...
+
+      const newTab = findTab(this.tabs, this.activeTabId);
+      if (!(newTab instanceof Tab)) { return; }
```
supminn, line 152:
```suggestion
      if (!(newTab instanceof Tab)) return
```

supminn (commented):
First Phase of testing.

packages/components/src/components/buttonsimple/buttonsimple.component.ts
```diff
@@ -200,6 +200,9 @@ class Buttonsimple extends TabIndexMixin(DisabledMixin(Component)) {
       if (event.key === 'Enter') {
         this.triggerClickEvent();
       }
+      if (event.key === ' ') {
+        event.preventDefault();
+      }
```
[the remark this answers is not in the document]
deveshcisco: When the space key is pressed, it also will scroll the page, so we have to prevent that behaviour.

packages/components/src/components/tab/tab.component.ts
```diff
@@ -31,7 +33,7 @@ import { IconNameMixin } from '../../utils/mixins/IconNameMixin';
  * @event keydown - (React: onKeyDown) This event is dispatched when a key is pressed down on the tab.
  * @event keyup - (React: onKeyUp) This event is dispatched when a key is released on the tab.
  * @event focus - (React: onFocus) This event is dispatched when the tab receives focus.
- *
+ * @event activechange - (React: onActiveChange) This event is dispatched when the active state of the tab changes
```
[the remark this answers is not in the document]
deveshcisco: We need this to be a custom event because it is being consumed only in case when the tab is a child of the tab list.

Ideally we wouldn't be needing anyone to use this react event because the actual on change is fired from the tab list component

packages/components/src/components/tab/tab.component.ts
```diff
@@ -120,6 +122,15 @@ class Tab extends IconNameMixin(Buttonsimple) {
...
   variant: Variant = DEFAULTS.VARIANT;
 
+  /**
+   * Id of the tab (used as a identificator when used in the tablist)
+   * Note: has to be unique!
+   *
+   * @default undefined
+   */
+  @property({ type: String, reflect: true, attribute: 'tabid' })
+  tabId?: string;
```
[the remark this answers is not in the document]
deveshcisco: It's a note for the consumer that they will need to provide unique tab IDs. We can generate them on our end because the associated tab panels has to be implemented separately.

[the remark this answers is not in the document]
deveshcisco: However, I've added logging of errors in case of non duplicate tabs

packages/components/src/components/tab/tab.stories.ts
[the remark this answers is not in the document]
deveshcisco: Removed the property

packages/components/src/components/tablist/tablist.types.ts
```diff
@@ -0,0 +1 @@
+export type ArrowButtonDirection = 'forward' | 'backward';
```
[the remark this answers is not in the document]
deveshcisco: Not sure if we need this much, it's a simple string type

packages/components/src/components/tablist/tablist.stories.ts
deveshcisco: cleaned up

packages/components/src/components/tablist/tablist.stories.ts
deveshcisco: Added

packages/components/src/components/tablist/tablist.stories.ts
deveshcisco: We need this because according to the guidelines there needs to be a corresponding tab panel to the tab.

If this markup is not provided then the automated accessibility checks and figma throw an error

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: updated

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: For better readability

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: I would be against this convention, makes the code look less like JS

packages/components/src/components/tablist/tablist.types.ts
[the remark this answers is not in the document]
supminn: Yes, its required for consistency.
If we rename any of these properties later, it would cause an issue of updating in multiple places.
Make sure to use this constant value on your components file too.

packages/components/src/components/tab/tab.component.ts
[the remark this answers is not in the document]
supminn: got it! Could we mention this in the ts-doc as well. 
This would help the consumers not use the same event unless they are using tablist.

Also, add a line mentioning `detail: { tabId: this.tabId, active },` data is being shared through the event dispatch.

packages/components/src/components/tablist/tablist.component.ts
supminn: i feel this should be a regular function instead of a getter.
It is currently being considered as a property atttribute.

packages/components/src/components/tab/tab.stories.ts
```diff
@@ -114,6 +115,7 @@ const meta: Meta = {
       'size',
       'soft-disabled',
       'type',
+      'tabId',
```
supminn, line 118: this is not the right way to fix it..
I figured the issue. Its with the way we are naming this attribute.

packages/components/src/components/tab/tab.component.ts
```diff
@@ -120,6 +122,15 @@ class Tab extends IconNameMixin(Buttonsimple) {
   @property({ type: String, reflect: true })
   variant: Variant = DEFAULTS.VARIANT;
 
+  /**
+   * Id of the tab (used as a identificator when used in the tablist)
+   * Note: has to be unique!
+   *
+   * @default undefined
+   */
+  @property({ type: String, reflect: true, attribute: 'tabid' })
```
supminn, line 131:
```suggestion
  @property({ type: String, reflect: true, attribute: 'tab-id' })
```

We would need to name it in this way so that it is consistent with the native HTML attributes.
This also resolved showing duplicate property value on storybook.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,430 @@
...
+import { KEYCODES } from './tablist.constants';
+
+import Tab from '../tab/tab.component';
+import type { ArrowButtonDirection } from './tablist.types';
+import { getFirstTab, getLastTab, getNextTab, getPreviousTab, findTab, getActiveTab } from './tablist.utils';
+
+/**
+ * Tab list organizes tabs into a container.
+ *
+ * Children of the tab list are mdc-tab elements, sent to the default slot.
```
supminn, line 16:
```suggestion
 * Children of the tab list are `mdc-tab` elements, sent to the default slot.
```
nitpick

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,430 @@
...
+ *
+ * The tabs can be navigated using the arrow keys, and selected by clicking, or pressing the Enter and Space keys.
```
supminn, line 18:
```suggestion
 * The tabs can be navigated using the left/right arrow keys, and selected by clicking, or pressing the Enter and Space keys.
```

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,430 @@
...
+ * @tagname mdc-tablist
+ *
+ * @dependency mdc-tab
+ * @dependency mdc-button
+ *
+ * @event change - (React: onChange) This event is dispatched when the tab is selected.
+ *
+ *
+ * Event that fires when user changes the active tab.
+ *     The function sent as the argument will receive the fired event
```
supminn, line 48:
```suggestion
 * @event change - (React: onChange) This event is dispatched when the tab is selected.
 * Event that fires when user changes the active tab.
 *     The function sent as the argument will receive the fired event
```

It doesn't matter if you give empty spaces. The description doesn't break into different paragraphs since its described within the attributes table. You can make use of `<br/>` if you really need to.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,430 @@
...
+ *      and the new tab id can be fetched from event.detail.tabId.
+ *
+ *    `(event: CustomEvent) => handleTabChange(event.detail.tabId);`
+ *
+ * @slot Default slot for mdc-tab elements.
+ *
+ * @cssproperty --mdc-tablist-gap - Gap between tabs
+ * @cssproperty --mdc-tablist-width - Width of the tablist
+ * @cssproperty --mdc-container-button-padding - Padding for when an arrow button is visible
+ */
```
supminn, line 58:
```suggestion
 * @cssproperty --mdc-container-button-padding - Padding for when an arrow button is visible
 * @cssproperty  --mdc-tablist-arrow-button-margin - Margin value for the arrow button
 */
```

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,430 @@
...
+class TabList extends Component {
+  /**
+   * ID of the active tab, defaults to the first tab if not provided
+   */
+  @property({ type: String, attribute: 'active-tabid', reflect: true })
```
supminn, line 63:
```suggestion
  @property({ type: String, attribute: 'active-tab-id', reflect: true })
```

Sry for the confusion, but this is the right implementation. Please get them updated on storybook side too.

packages/components/src/components/tablist/tablist.stories.ts
```diff
@@ -0,0 +1,125 @@
...
+      description: 'ID of the active tab, defaults to the first tab if not provided',
+    },
+    tabvariant: {
+      control: 'select',
+      description: 'Set the variant of tab inside the tablist',
+      options: Object.values(TAB_VARIANTS),
+    },
+    ...hideControls([
+      'tabvariant',
+      'activeTabId',
```
supminn, line 91:
```suggestion
```

This is no longer required and is incorrect to hideControls of wrongly mapped attribute.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,430 @@
...
+     */
+    if (!this.tabs) { return; }
+    if (Array.isArray(this.tabs) && this.tabs.length === 0) {
+      throw new Error('The tablist component must have at least one child tab');
+    }
+
+    const tabIds = this.tabs.map((tab) => tab.tabId);
+    if (new Set(tabIds).size !== this.tabs.length) {
+      // eslint-disable-next-line no-console
+      console.error('The tabs inside the tab list must have unique tab ids');
```
supminn, line 128:
```suggestion
     if(this.onerror){
        this.onerror('The tabs inside the tab list must have unique tab ids');
       }
```

We should handle errors in this way. It is mentioned in the convention to avoid using console

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,430 @@
...
+   * Also set up the event listeners.
+   */
+  protected override firstUpdated(): void {
+    /**
+     * If there are no tabs, skip the initialization of the active tab and the event listeners.
+     * Then throw an error.
+     */
+    if (!this.tabs) { return; }
+    if (Array.isArray(this.tabs) && this.tabs.length === 0) {
+      throw new Error('The tablist component must have at least one child tab');
```
supminn, line 122:
```suggestion
     if(this.onerror){
        this.onerror('The tablist component must have at least one child tab');
     }
```
We cannot throw errors from a web component.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,430 @@
...
+      tab.addEventListener('mousedown', this.handleMousedown);
+    });
+
+    if (!this.activeTabId) {
+      this.activeTabId = getFirstTab(this.tabs)?.tabId;
+    }
+  }
+
+  /**
+   * Observe the tablist element for changes in the activetabid attribute.
```
supminn, line 147:
```suggestion
   * Observe the tablist element for changes in the activeTabId attribute.
```
Lets be consistant with the naming conventions.

supminn (changes requested):
Added few more comments.
Functionality is working as expected.

packages/components/src/components/tablist/tablist.stories.ts
```diff
@@ -0,0 +1,125 @@
...
+  @change="${action('onchange')}"
+  @keyup="${action('onkeyup')}"
+  @keydown="${action('onkeydown')}"
+  active-tabid="${args['active-tabid']}"
+  aria-label="Media types">
+  <mdc-tab
+    variant=${args.tabvariant}
+    text="Photos"
+    icon-name="add-photo-bold"
+    tabid="photos-tab"
```
supminn, line 20:
```suggestion
    tab-id="photos-tab"
```
Please update it accordingly across this file.

packages/components/src/components/tab/tab.component.ts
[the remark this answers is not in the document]
deveshcisco: Added

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: It would show up either way, but I've made the changes

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: @cscheffauer there is no convention for this, it just felt like removing the brackets would reduce readability

packages/components/src/components/tab/tab.component.ts
deveshcisco: changed

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: removed the line breaks

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: updated

packages/components/src/components/tablist/tablist.stories.ts
deveshcisco: We need to keep this hidden in the tablist story because it is actually an attribute of the tab component.

I've added this here so that there can be examples with different tab variants

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Updated

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Replaced

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: Replaced

packages/components/src/components/tablist/tablist.stories.ts
deveshcisco: updated

packages/components/src/components/tablist/tablist.stories.ts
deveshcisco: removed

supminn (approved):

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: This has been updated to use queryAll in https://github.com/deveshcisco/momentum-design/commit/5678199dcffab39bfc2a87442fc9206351eac9f1

2020Deeya (approved):

packages/components/src/components/tab/tab.component.ts
```diff
@@ -131,6 +148,10 @@ class Tab extends IconNameMixin(Buttonsimple) {
     this.softDisabled = undefined as unknown as boolean;
     this.size = undefined as unknown as ButtonSize;
     this.type = undefined as unknown as ButtonType;
+
+    if (this.onerror) {
```
cscheffauer, line 152: issue: this fires ALL the time when the tab gets mounted. It does not check if the tab id is actually set or not.
if i set "onerror" as a method on the tab, it will immediately fire after mount (even if i set tab id or not).
So either clean this up or remove it.

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,437 @@
...
+
+  /**
+   * @internal
+   */
+  @state() private showBackwardArrowButton: boolean = false;
+
+  /**
+   * @internal
+   */
+  private isRtl = (): boolean => (
```
cscheffauer, line 91: nitpick: this is a arrow function, but all other methods are written in the normal class method syntax

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,437 @@
...
+   * the first update
+   *
+   * Also set up the event listeners.
+   */
+  protected override firstUpdated(): void {
+    /**
+     * If there are no tabs, skip the initialization of the active tab and the event listeners.
+     * Then throw an error.
+     */
+    if (!this.tabs) { return; }
```
cscheffauer, line 120: question: why is this line returning, but when the array has length 0 its erroring? shouldn't both error?

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,437 @@
...
+    }
+
+    const resizeObserver = new ResizeObserver((): void => {
+      this.handleArrowButtonVisibility();
+    });
+    resizeObserver.observe(this);
+
+    this.tabs?.forEach((tab) => {
+      tab.addEventListener('focus', this.handleFocus);
+      tab.addEventListener('mousedown', this.handleMousedown);
```
cscheffauer, line 142: question: why does every tab need a "mousedown" handler? for what reason?

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,437 @@
...
+      }
+    }
+
+    const resizeObserver = new ResizeObserver((): void => {
+      this.handleArrowButtonVisibility();
+    });
+    resizeObserver.observe(this);
+
+    this.tabs?.forEach((tab) => {
+      tab.addEventListener('focus', this.handleFocus);
```
cscheffauer, line 141: suggestion: can't we just put a focus event listener on the host of the tablist and then check which "tab" gets focused? i think adding a eventlistener on every Tab is not necessary and is not as performant as if you just have a focus event listener on the host

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -0,0 +1,437 @@
...
+            @click="${() => this.scrollTabs(direction)}"
+          ></mdc-button>`
+         : nothing}`;
+
+     return html`
+      ${arrowButton('backward')}
+      <div class="container">
+        <slot></slot>
+      </div>
+      ${arrowButton(ARROW_BUTTON_DIRECTION.FORWARD)}`;
```
cscheffauer, line 431: issue: there is a severe a11y issue right now - if you focus with the keyboard to the "forward" or "backward" buttons and press enter / space, the focus will be lost. This has to be fixed

packages/components/src/components/tablist/tablist.styles.ts
```diff
@@ -0,0 +1,35 @@
+import { css } from 'lit';
+
+const styles = css`
+  :host {
+    --mdc-tablist-focus-ring-outer-width: calc(0.0625rem + calc(2 * 0.125rem));
```
cscheffauer, line 5: issue: this is not scalable, like if we change the size of the focus-ring in the focus-ring mixin and forget about this code here, it will be broken. Instead of hard-coding values, please use the css variables from the mixin instead.

cscheffauer (changes requested):

packages/components/src/components/tablist/tablist.stories.ts
```diff
@@ -0,0 +1,123 @@
+import type { Meta, StoryObj, Args } from '@storybook/web-components';
+import { action } from '@storybook/addon-actions';
+import '.';
+import { html } from 'lit';
+import { disableControls, hideControls } from '../../../config/storybook/utils';
+import { TAB_VARIANTS } from '../tab/tab.constants';
+import '../badge';
+
+const render = (args: Args) => html`
+<mdc-tablist
```
cscheffauer, line 10: issue: inside a role "tablist", there should only be tabs, not buttons. We have to put the tablist role to the container div inside. This would mean you also have to pass the aria-label to it (using the data approach as in other components). this is a a11y issue, which needs to be addressed.

image

cscheffauer (changes requested):

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: This line is a check for if the tabs constant is `undefined`, so it's a check for if the tabs does not exist.

The second check is if the tabs constant exists and is of array type, but empty. In this case it means the tablist has no children that are mdc-tab, so we send an error that there are no children of type mdc-tab

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: This is in place because we are listening for the focus event to refocus the active tab.

However the mousedown event also fires a focus event before the clicked tab can be set to active, which caused the older tab to remain focused.

So we supress the default action of mousedown with this listener for now. 

However if we change the implementation of the listener to run on the parent then we can remove this

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: We do have the event handlers and other methods as arrow functions

packages/components/src/components/tab/tab.component.ts
deveshcisco: added check

deveshcisco: fixed in [9df84d6](https://github.com/momentum-design/momentum-design/pull/1130/commits/9df84d6eed0e740f9184771347cd5dedf64d6709)

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: fixed in [0add94d](https://github.com/momentum-design/momentum-design/pull/1130/commits/0add94d415c3dc058ddd9c131947d282032ff160)

packages/components/src/components/tablist/tablist.styles.ts
deveshcisco: fixed in [e451557](https://github.com/momentum-design/momentum-design/pull/1130/commits/e451557d3783944cda25f621ac6cedbf4c03d310)

packages/components/src/components/tablist/tablist.stories.ts
deveshcisco: fixed in [3e1d6e6](https://github.com/momentum-design/momentum-design/pull/1130/commits/3e1d6e6f46e95c629fc8f967fbbf44130159b405)

packages/components/src/components/tablist/tablist.component.ts
deveshcisco: fixed in [29c3257](https://github.com/momentum-design/momentum-design/pull/1130/commits/29c3257f077e79801bf901227949e7fad07decbc)

packages/components/src/components/tablist/tablist.component.ts
cscheffauer: ok - i will create a ticket to add a eslint rule to make this consistent.

packages/components/src/components/tablist/tablist.component.ts
cscheffauer: that is an issue and must be fixed - like supressing an event is not how it should be done.
please create a tech debt ticket for this and and it to the backlog, since i do not want to block this further cause of it and in general it works, its just not ideal.

packages/components/src/components/tablist/tablist.component.ts
cscheffauer: this is not fixed! still an issue when i tested it

cscheffauer (changes requested):
There is an issue in firefox (tablist container gets focused, but shouldn't be).
image

cscheffauer (approved):

PR #1133, opened 2025-03-06
refactor(popover): remove modalcontainer for popover

Jason20020:
### Description

Refactor Popover - remove modalcontainer


packages/components/src/components/popover/popover.utils.ts
```diff
@@ -159,11 +159,32 @@ export class PopoverUtils {
...
-    if (!this.popover.containerElement) return;
-
-    Object.assign(this.popover.containerElement.style, {
+    Object.assign(this.popover.style, {
       left: `${x}px`,
       top: `${y}px`,
     });
   }
+
+  createBackdrop() {
```
cscheffauer, line 168: suggestion: instead of creating this, can you maybe have a look at this css attribute?
https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop
it seems to work on role="dialog" elements, so we could put it on the host?
then you wouldn't need the createBackdrop function, but do it via css

cscheffauer (changes requested):
just a really quick scan - i have to test it locally, will do tomorrow

packages/components/src/components/popover/popover.styles.ts
```diff
@@ -22,20 +36,6 @@ const styles = css`
     }
```
gabrielchl, line 36: now that the content styling isn't handled by modalcontainer
i believe we need to handle the contrast colors for the popover container here
this is what it currently looks like in contrast mode
Screenshot 2025-03-12 at 00 48 41

cscheffauer (approved):

PR #1135, opened 2025-03-06
chore: specs for list and listitem component

arjunkesava:
### Description

- This PR adds e2e specs for both list and list item components.
- In list, the keyboard navigation in between the list items and inside the children of each list item is covered.
- In listitem, a few examples of using slots of leading-controls and trailing-controls is used.

### Links

- Will post soon.


packages/components/src/components/listitem/listitem.e2e-test.ts
```diff
@@ -1,66 +1,217 @@
...
+      });
+    });
+  });
+
+  /**
+   * ACCESSIBILITY
+   */
+  await test.step('accessibility', async () => {
+    await componentsPage.accessibility.checkForA11yViolations('listitem-default', true, {
+      rules: ['aria-required-parent'],
```
arjunkesava, line 130: Reason to ignore this rule: The listitem must contain a role and the above step creates a listitem without a parent list in the Stickersheet class. In stickersheet class we don't have the option to set parent for the element to take a screenshot.

packages/components/src/components/listitem/listitem.e2e-test.ts
```diff
@@ -1,66 +1,217 @@
...
+    listitemSheet.setAttributes({
+      label: primaryLabel,
+    });
+    await listitemSheet.createMarkupWithCombination({}, options);
+    listitemSheet.setChildren(`
+      <mdc-checkbox checked slot="leading-controls" data-aria-label="${primaryLabel}"></mdc-checkbox>
+    `);
+    await listitemSheet.createMarkupWithCombination({}, options);
+    listitemSheet.setChildren(`
+      <mdc-toggle checked slot="leading-controls" data-aria-label="${primaryLabel}"></mdc-toggle>
```
supminn, line 61: question: do we want to show this test case?
According to the design spec, toggle should be supported on trailing-controls only.. not on leading-controls.

I understand that we cannot restrict it in the case of web-components. However, my question here is - do we want to specifically mentioned this as a test case? Can we show it as a trailing-controls instead.

packages/components/src/components/listitem/listitem.e2e-test.ts
supminn: Alternatively, we can evaluate on the wrapper container and add the role of list using JS.

supminn (dismissed):

packages/components/src/components/listitem/listitem.e2e-test.ts
arjunkesava: Done.

packages/components/src/components/listitem/listitem.e2e-test.ts
arjunkesava: Removed it.

supminn (approved):

PR #1136, opened 2025-03-06
feat: Multiple Brand Visual updates

morgan1bathe:
**Additions:**
- webex-background: new backgrounds added
- webex-suite-wordmark: Added ‘placeholder’ wordmark

**Visual updates:**
- webex-suite-wordmark: updated wordmark to sit optically centered in frame, resized frame to fit wordmark

**No updates:**
All the rest


packages/assets/brand-visuals/src/backgrounds/webex-background-dark.svg
```diff
@@ -0,0 +1,51 @@
+<svg width="849" height="509" viewBox="0 0 849 509" fill="none" xmlns="http://www.w3.org/2000/svg">
```
cscheffauer, line 1: Note: currently our backgrounds in the app are png's, not svg's - this will cause disruption of developers switch. Also the existing tooling for brand-visuals expects all the backgrounds in the folder only as pngs, so i would need to make a tooling change as well.
Let me know what u think - can also write via direct message.

cscheffauer (changes requested):

PR #1139, opened 2025-03-07
perf: build options and options group components.

arjunkesava:
### Description

- This PR build the options and options group component.
- An option is based on the list item with restricted fields.
- As of now, this PR contains a basic version of an option:
  - An option with a label
  - An option with a leading icon.
  - An option with icon at the end when selected.

- Option group contains the role "group" and a label and a disabled state.
- If the option group is disabled, then all of it's children are disabled.
- The option group doesn't contain any form value, it acts as a wrapper for a group of options.
- The primary purpose of option group is for accessibilty and to group similar list of options.

### Screenshots:

image
image


packages/components/src/components/option/option.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Listens to changes in the default slot and updates the label of the option accordingly.
+   * This is used to set the label of the option when it is not explicitly set.
+   * It is called internally when the slot is changed.
+   */
+  private handleDefaultSlotChange() {
+    const slot = this.shadowRoot?.querySelector('slot');
+    if (slot) {
+      this.label = slot.assignedNodes()[0]?.textContent?.trim() || this.label;
+    }
```
supminn, line 66:
```suggestion
    if (slot && !this.label) {
      this.label = slot.assignedNodes()[0]?.textContent?.trim();
    }
```

`<option label="Label 1">label 2</option>` on native HTML shows Label 1. We are currently showing Label 2.

The above suggestion would fix it.

packages/components/src/components/optgroup/optgroup.component.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @dependency mdc-text
+ *
+ * @tagname mdc-optgroup
+ *
+ * @slot default - This is a default slot for mdc-option elements.
+ *
+ * @cssproperty --mdc-optgroup-disabled-color - Allows customization of the disabled option color.
+ */
+class Optgroup extends DataAriaLabelMixin(DisabledMixin(Component)) {
```
supminn, line 22:
```suggestion
class OptGroup extends DataAriaLabelMixin(DisabledMixin(Component)) {
```

packages/components/src/components/optgroup/optgroup.component.ts
```diff
@@ -0,0 +1,74 @@
...
+import { property, queryAssignedElements } from 'lit/decorators.js';
+import { Component } from '../../models';
+import { DataAriaLabelMixin } from '../../utils/mixins/DataAriaLabelMixin';
+import { DisabledMixin } from '../../utils/mixins/DisabledMixin';
+import { TYPE, VALID_TEXT_TAGS } from '../text/text.constants';
+import { HEADER_ID } from './optgroup.constants';
+import styles from './optgroup.styles';
+
+/**
+ * optgroup component, which creates a grouping of options within a listbox element.
```
supminn, line 12:
```suggestion
 * optgroup component, which creates a grouping of mdc-option within a listbox element.
```

packages/components/src/components/optgroup/optgroup.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  @property({ type: String, reflect: true }) label?: string;
+
+  /** @internal */
+  @queryAssignedElements({ selector: 'mdc-option' })
+  options!: Array<HTMLElement>;
+
+  private setDisabledForAllOptions(): void {
+    this.options.forEach((element) => {
+      if (this.disabled) {
+        element.setAttribute('disabled', 'true');
```
supminn, line 35:
```suggestion
        element.setAttribute('disabled', '');
```
Since its a boolean value, we need to set it in this manner.

packages/components/src/components/option/option.stories.ts
supminn, line 1: issue: a11y on storybook.

image

supminn (changes requested):

supminn (approved):

packages/components/src/components/option/option.stories.ts
arjunkesava: This is a common problem across all components, and not just with the option component.

PR #1142, opened 2025-03-11
docs: created chip components with storybook docs

supminn:
### Description

Chip component has 4 different components
* mdc-chip - Label chip that supports icon and 9 different colors and disabled state.
* mdc-alertchip - Alert chip with 5 variants. It doesn't support disabled state.
* mdc-filterchip - Filter chip with selected flag and disabled state.
* mdc-inputchip - Input chip with close button, error validation flag and custom event 'remove' to describe the behavior of the button.

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-569


packages/components/src/components/chip/chip.component.ts
```diff
@@ -0,0 +1,83 @@
...
+    super();
+    this.role = DEFAULTS.ROLE;
+    this.size = DEFAULTS.SIZE;
+    this.active = undefined as unknown as boolean;
+    this.softDisabled = undefined as unknown as boolean;
+  }
+
+  /**
+   * Renders the icon element if available.
+   * @returns The icon element if available, otherwise nothing.
```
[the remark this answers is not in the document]
supminn: I remember discussing this point while working on the input component. We'd need to change the rendering logic within mdc-icon. 
Until then, yes it would show a blank space.

packages/components/src/components/alertchip/alertchip.component.ts
```diff
@@ -0,0 +1,63 @@
...
+ * - It supports a leading icon along with label.
+ * - It supports 5 variants of alerts - neutral, warning, error, success, and informational
+ *
+ * This component is built by extending Buttonsimple.
+ *
+ * @tagname mdc-alertchip
+ *
+ * @dependency mdc-icon
+ * @dependency mdc-text
+ */
```
arjunkesava, line 22: Please add css props for this alert chip.

packages/components/src/components/alertchip/alertchip.component.ts
```diff
@@ -0,0 +1,63 @@
...
+   */
+  @property({ type: String }) variant: VariantType = DEFAULTS.VARIANT;
+
+  /**
+   * The visible label text of the alertchip.
+   *
+   * We recommend limiting the <b>maximum length of the label text to 20 characters</b>,
+   * including empty spaces to split words.
+   */
+  @property({ type: String }) label = '';
```
arjunkesava, line 42: If you set the label as an empty string, then by default it will show as <mdc-alertchip label="" />, instead of if you set it to undefined, then it will show as <mdc-alertchip />, the label will be set only if it's defined by the consumer.
```suggestion
  @property({ type: String }) label?: string;
```

packages/components/src/components/alertchip/alertchip.styles.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const styles = css`
+  :host {
+    height: 1.5rem;
+    gap: 0.25rem;
+    padding: 0 0.5rem;
+    border-radius: 0.25rem;
+    border: 1px solid;
+    cursor: pointer;
+  }
```
arjunkesava, line 12: please add css properties over here.

packages/components/src/components/alertchip/alertchip.utils.ts
```diff
@@ -0,0 +1,19 @@
+import { VARIANTS } from './alertchip.constants';
+import type { IconNames } from '../icon/icon.types';
+import type { VariantType } from './alertchip.types';
+
+const getAlertIcon = (type: VariantType): IconNames => {
```
arjunkesava, line 5: instead of importing all iconames, please use extract type, something similar to this.
https://github.com/momentum-design/momentum-design/blob/11a49de8326fc25b2c2dc1681b7bc8f9999d0d12/packages/components/src/components/formfieldwrapper/formfieldwrapper.types.ts#L7-L10

packages/components/src/components/chip/chip.component.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * mdc-chip is an interactive element that can be used to represent a chip. It supports a leading icon along with label.
+ * Consumers can wrap this component around a tooltip to provide additional context.
+ *
+ * It is recommended to keep the label text for the chip component concise and compact.
+ * For best results, we recommend limiting the <b>maximum length of the label text to 20 characters</b>,
+ * including empty spaces to split words.
+ *
+ * This component is built by extending Buttonsimple.
+ *-
```
arjunkesava, line 20:
```suggestion
 *
```

packages/components/src/components/chip/chip.stories.ts
```diff
@@ -0,0 +1,68 @@
...
+export default meta;
+
+export const Example: StoryObj = {
+  args: {
+    color: COLOR.DEFAULT,
+    label: 'Label',
+    'icon-name': 'placeholder-bold',
+    disabled: false,
+    class: 'custom-classname',
+    style: 'margin-top: 20px;',
```
arjunkesava, line 57: I guess these can be removed.

packages/components/src/components/filterchip/filterchip.stories.ts
```diff
@@ -0,0 +1,63 @@
...
+      'icon-name', 'color']),
+  },
+};
+
+export default meta;
+
+export const Example: StoryObj = {
+  args: {
+    class: 'custom-classname',
+    style: 'margin-top: 20px;',
```
arjunkesava, line 49: same here.

packages/components/src/components/inputchip/inputchip.component.ts
```diff
@@ -0,0 +1,74 @@
+import { CSSResult, html, nothing } from 'lit';
+import { property } from 'lit/decorators.js';
+import styles from './inputchip.styles';
+import { Component } from '../../models';
+import { DisabledMixin } from '../../utils/mixins/DisabledMixin';
+import { IconNameMixin } from '../../utils/mixins/IconNameMixin';
+import { IconNames } from '../icon/icon.types';
```
arjunkesava, line 7:
```suggestion
import type { IconNames } from '../icon/icon.types';
```

packages/components/src/components/inputchip/inputchip.component.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @event keydown - (React: onKeyDown) This event is dispatched when a key is pressed down on the close button.
+ * @event keyup - (React: onKeyUp) This event is dispatched when a key is released on the close button.
+ * @event focus - (React: onFocus) This event is dispatched when the close button receives focus.
+ */
+class Inputchip extends IconNameMixin(DisabledMixin(Component)) {
+  @property({ type: String }) label = '';
+
+  @property({ type: Boolean }) error = false;
+
+  @property({ type: String, attribute: 'clear-aria-label' }) clearAriaLabel = '';
```
arjunkesava, line 33: please add description for these 3 fields.

arjunkesava: and does, error prop won't need `reflect: true`?

packages/components/src/components/alertchip/alertchip.component.ts
```diff
@@ -0,0 +1,63 @@
...
+ * - It supports 5 variants of alerts - neutral, warning, error, success, and informational
+ *
+ * This component is built by extending Buttonsimple.
+ *
+ * @tagname mdc-alertchip
+ *
+ * @dependency mdc-icon
+ * @dependency mdc-text
+ */
+class Alertchip extends Buttonsimple {
```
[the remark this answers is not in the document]
arjunkesava: Can we extend the Alertchip with Chip instead of Buttonsimple. that way we can set the iconname with the variant and reduce the duplicate code in render part.

arjunkesava (changes requested):

packages/components/src/components/alertchip/alertchip.component.ts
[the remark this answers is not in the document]
supminn: We do not want to expose icon-name as a property. 
Also, chip has `color` whereas Alertchip has `variant`.
That's the reason I didn't extend using Chip.

packages/components/src/components/alertchip/alertchip.component.ts
supminn: Could you please clarify this?

packages/components/src/components/alertchip/alertchip.component.ts
supminn: done

packages/components/src/components/alertchip/alertchip.component.ts
arjunkesava: As the label is a mandatory field, we can't make it optional.

packages/components/src/components/inputchip/inputchip.component.ts
```diff
@@ -0,0 +1,97 @@
...
+    <mdc-icon 
+      name="${this.iconName as IconNames}" 
+      length-unit="rem" 
+      size="1"></mdc-icon>
+  `;
+  }
+
+  /**
+   * Handles the behavior of the close button on click event.
+   * @param event The event object.
```
arjunkesava, line 70:
```suggestion
   * @param event - The event object.
```

arjunkesava (approved):

arjunkesava (approved):

PR #1150, opened 2025-03-13
feat(popover): add disable-aria-expanded attribute

gabrielchl:
### Description

- add a "disable-aria-expanded" attribute to the popover component
  - will be used by coachmark and tooltip
  - default is false, so we don't affect the normal behaviour for popover

popover is sitll wip and does not have e2e tests yet, so none have been added for this change

### Links

_Please replace this line with relevant links to the changes made in this Pull Request._


Jason20020 (approved):
LGTM

PR #1152, opened 2025-03-13
fix: icon arialabelledby

cscheffauer:
### Description

- Passing aria-labelledby should also add role="img" to the icon, in case a tooltip is connected to it.
- Added e2e test


maxinteger (commented):
LGTM

maxinteger (approved):

PR #1165, opened 2025-03-18
fix: create progressbar component

2020Deeya:
### Description

Create Progressbar component
Add Storybook docs

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-581
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-582


packages/components/src/components/progressbar/progressbar.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * Renders the progress bar with dynamic width and variant styles.
+   * @returns The rendered HTML for the progress bar.
+   */
+  private renderProgressbar = () => {
+    const variant = this.getValidationVariant();
+    const isGap = this.clampedValue > 0 && this.clampedValue < 100;
+    const progressWidth = this.error ? '100' : `${this.clampedValue}`;
+    return html`
+      <div
```
deveshcisco, line 104: Instead of rendering a `div` here can you look into using the native HTML progress element? It should reduce the need to render the stye dynamically.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

2020Deeya: It was our first choice, but we also need a gap of 4px within the progressbar as per figma spec. That made us choose the div approach.

packages/components/src/components/progressbar/progressbar.component.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  constructor() {
+    super();
+    this.id = '';
+    this.disabled = undefined as unknown as boolean;
+  }
+
+  /**
+   * Ensures that the value is clamped between 0 and 100
+   * @returns The clamped value
```
supminn, line 72:
```suggestion
   * @returns The clamped value
   * @internal
```

Adding this would not show the property in storybook.

packages/components/src/components/progressbar/progressbar.component.ts
```diff
@@ -0,0 +1,143 @@
...
+      this.helpTextType = VALIDATION.SUCCESS;
+    } else {
+      this.helpTextType = VALIDATION.DEFAULT;
+    }
+    return this.helpTextType;
+  }
+
+  /**
+   * Renders the progress bar with dynamic width and variant styles.
+   * @returns The rendered HTML for the progress bar.
```
supminn, line 97:
```suggestion
   * @returns The rendered HTML for the progress bar.
   * @internal
```
same as above

packages/components/src/components/progressbar/progressbar.component.ts
```diff
@@ -0,0 +1,143 @@
...
+   * Define error state of the progressbar
+   * - **true**
+   * - **false**
+   * @default false
+   */
+  @property({ type: Boolean, attribute: 'error' }) error = false;
+
+  constructor() {
+    super();
+    this.id = '';
```
supminn, line 66: question: do you want to set it to an empty string or a fixed value (randomly generated).
Defining id is required for connecting the label field with progress bar

packages/components/src/components/progressbar/progressbar.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /**
+   * Renders the progress bar with dynamic width and variant styles.
+   * @returns The rendered HTML for the progress bar.
+   */
+  private renderProgressbar = () => {
+    const variant = this.getValidationVariant();
+    const isGap = this.clampedValue > 0 && this.clampedValue < 100;
+    const progressWidth = this.error ? '100' : `${this.clampedValue}`;
```
supminn, line 102: issue: Why is this logic required? I believe its incorrect.
If we set the value to a number less than 100 and set 'error' to true, the progresswidth is getting set to 100 due to this logic.

packages/components/src/components/progressbar/progressbar.component.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+   * Renders the component based on the variant (inline or default).
+   * @returns The rendered HTML for the component.
+   */
+  public override render() {
+    const isInline = this.variant === VARIANT.INLINE;
+    return html`
+      ${isInline
+    ? html`<div part="inline-label-container">${this.renderLabel()} ${this.renderProgressbar()}</div>`
```
supminn, line 126:
```suggestion
    ? html`<div part="inline-label-container">${this.renderLabelElement()} ${this.renderProgressbar()}</div>`
```

Use renderLabelElement instead. This would only render the label text and no other slots.
renderLabel has slots for required label and icon info which is not required in this component.

packages/components/src/components/progressbar/progressbar.component.ts
```diff
@@ -0,0 +1,143 @@
...
+    : html`
+            <div part="label-container">
+              ${this.renderLabel()}
```
supminn, line 129:
```suggestion
              ${this.renderLabelElement()}
```

same as above

packages/components/src/components/progressbar/progressbar.component.ts
```diff
@@ -0,0 +1,143 @@
+import { CSSResult, html, nothing } from 'lit';
+import { property } from 'lit/decorators.js';
+import styles from './progressbar.styles';
+import FormfieldWrapper from '../formfieldwrapper';
+import { DEFAULTS, VARIANT } from './progressbar.constants';
+import { Variant } from './progressbar.types';
```
supminn, line 6:
```suggestion
import type { Variant } from './progressbar.types';
```

packages/components/src/components/progressbar/progressbar.styles.ts
```diff
@@ -0,0 +1,82 @@
...
+  --mdc-progressbar-active-background-color: var(--mds-color-theme-control-active-normal);
+  --mdc-progressbar-success-color: var(--mds-color-theme-indicator-stable);
+  --mdc-progressbar-error-color: var(--mds-color-theme-indicator-attention);
+  --mdc-progressbar-height: 0.25rem;
+  --mdc-progressbar-border-radius: var(--mdc-progressbar-height);
+
+  --mdc-progressbar-label-color: var(--mds-color-theme-text-primary-normal);
+  --mdc-progressbar-label-line-height: var(--mds-font-lineheight-body-midsize);
+  --mdc-progressbar-label-font-size: var(--mds-font-size-body-midsize);
+  --mdc-progressbar-label-font-weight: 400;
```
supminn, line 15: issue: we should not use hard coded values. Make sure to use the corresponding font token `--mds-font-weight-regular`

packages/components/src/components/progressbar/progressbar.styles.ts
```diff
@@ -0,0 +1,82 @@
...
+
+:host([variant="inline"])::part(label-text) {
+  width: unset;
+}
+
+:host::part(progress-bar) {
+  height: 100%;
+  background-color: var(--mdc-progressbar-active-background-color);
+  border-radius: var(--mdc-progressbar-border-radius);
+  transition: width 0.3s ease-in-out;
```
supminn, line 53: question: Do we need to add this transition? How can we check this on figma spec?

supminn (changes requested):

packages/components/src/components/progressbar/progressbar.component.ts
2020Deeya: It's an expected behavior. If we set the value to a number less than 100 and set 'error' to true, the progresswidth should be 100 only. This is for error state.

packages/components/src/components/progressbar/progressbar.styles.ts
2020Deeya: Not strictly required.

packages/components/src/components/progressbar/progressbar.component.ts
2020Deeya: yes, ID would be required for connecting the label field with progress bar. We have set it's initial value which can always be overridden.

packages/components/src/components/progressbar/progressbar.e2e-test.ts
```diff
@@ -0,0 +1,143 @@
...
+  });
+  await test.step('attributes are set properly', async () => {
+    await progressbar.evaluate((el) => el.setAttribute('value', '30'));
+    await expect(progressbar).toHaveAttribute('value', '30');
+  });
+};
+
+const runVisualRegressionTest = async (componentsPage: ComponentsPage, progressbarStickerSheet: any) => {
+  // eslint-disable-next-line max-len
+  progressbarStickerSheet.setAttributes({ label: 'Label', 'help-text': 'Helper text', 'data-aria-label': 'This is a progress bar' });
```
supminn, line 51:
```suggestion
  progressbarStickerSheet.setAttributes({ 
    label: 'Label',
   'help-text': 'Helper text', 
   'data-aria-label': 'This is a progress bar' 
   });
```

Let's try not to add eslint-disable for max-len. Instead, you could split the code into multiple lines.

packages/components/src/components/progressbar/progressbar.e2e-test.ts
```diff
@@ -0,0 +1,143 @@
...
+    {
+      variant: [VARIANT.INLINE],
+      value: [100],
+    },
+    { rowWrapperStyle: 'display: flex; gap: 3rem; align-items: center; margin-bottom: 3.5rem;' },
+  );
+
+  // Error state (inline variant)
+  // eslint-disable-next-line max-len
+  await progressbarStickerSheet.setAttributes({ error: true, label: 'Label', 'help-text': 'Helper text', 'data-aria-label': 'This is a progress bar' });
```
supminn, line 108:
```suggestion
  await progressbarStickerSheet.setAttributes({ 
  error: true, 
  label: 'Label', 
  'help-text': 'Helper text', 
  'data-aria-label': 'This is a progress bar' 
  });
```

same as above

packages/components/src/components/progressbar/progressbar.e2e-test.ts
```diff
@@ -0,0 +1,143 @@
...
+    await attributeTestCases(componentsPage);
+  });
+
+  await test.step('Visual Regression and Accessibility tests', async () => {
+    const progressbarStickerSheet = new StickerSheet(componentsPage, 'mdc-progressbar');
+    await runVisualRegressionTest(componentsPage, progressbarStickerSheet);
+  });
+};
+
+test.use({ viewport: { width: 1000, height: 2250 } });
```
supminn, line 140: nitpick: Its not necessary to set a viewport I believe.

supminn (approved):

PR #1172, opened 2025-03-19
fix: connectedcallback-attribute

cscheffauer:
### Description

- Fixing all components which do attribute initialisation by moving the initialise from constructor to connectedCallback
  - This is fixing an issue we currently have in the majority of our components, which blocks Angular consumers from properly using our components. 
- Adding a new convention for initialising attribute
- Updated missing conventions in Readme.md of conventions
- Adding badge type export to index file


packages/components/src/index.ts
```diff
@@ -79,5 +80,6 @@ export type {
   SpinnerSize,
   SpinnerVariant,
   PopoverPlacement,
+  BadgeType,
```
supminn, line 83: question: Why are we doing this?
Do we also need to export types from other components?

supminn (approved):

packages/components/src/index.ts
cscheffauer: this is needed for consumption of the badge, i just didn't want to create a separate PR for a one liner.

supminn (approved):

PR #1173, opened 2025-03-20
perf: Build select component

arjunkesava:
### Description

- Build select component.

image

#### A11y guidelines:

- The keyboard interactions for the select has been followed with these rules mentioned below: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/
- For few keyboard interactions such as "Space" or "PageDown" or "Enter", we will stop them and don't propagate as these conflict with browser based default events. 

### Links

- https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-576


packages/components/src/components/listitem/listitem.styles.ts
```diff
@@ -9,10 +9,11 @@ const styles = css`
     --mdc-listitem-primary-label-color: var(--mds-color-theme-text-primary-normal);
     --mdc-listitem-secondary-label-color: var(--mds-color-theme-text-secondary-normal);
     --mdc-listitem-disabled-color: var(--mds-color-theme-text-primary-disabled);
+    --mdc-listitem-column-gap: 0.75rem;
```
supminn, line 12: nitpick: we'd need to add this cssprop on ts-docs within the components file.

packages/components/src/components/option/__screenshots__/mobile-chrome/mdc-option-high-contrast.png
supminn, line 1: This doesn't look right. The 'L' is looking different on the added screenshot. 
We need to have it as the previous one. What was the change that you've applied here?

packages/components/src/components/option/option.styles.ts
```diff
@@ -1,12 +1,29 @@
...
     height: 2.25rem;
   }
   :host::part(default-slot) {
     display: none; 
   }
+  :host::part(trailing) {
+    min-width: var(--mdc-option-trailing-min-width);
+  }
+  :host::part(leading) {
+    width: calc(100% - var(--mdc-option-trailing-min-width) - var(--mdc-listitem-column-gap));
```
supminn, line 17: could you help me understand why is this calculation done?

packages/components/src/components/optgroup/__screenshots__/firefox/mdc-optgroup-rtl.png
supminn, line 1: question: Is this update correct?
If yes, shouldn't it update all the snapshots and not just this one?

packages/components/src/components/select/select.component.ts
```diff
@@ -0,0 +1,363 @@
...
+ *
+ * @slot default - This is a default/unnamed slot for options and/or option group.
+ *
+ * @event click - (React: onClick) This event is dispatched when the select is clicked.
+ * @event change - (React: onChange) This event is dispatched when the select is changed.
+ * @event keydown - (React: onKeyDown) This event is dispatched when a key is pressed down on the select.
+ * @event keyup - (React: onKeyUp) This event is dispatched when a key is released on the select.
+ * @event focus - (React: onFocus) This event is dispatched when the select receives focus.
+ */
+class Select extends FormInternalsMixin(DisabledMixin(DataAriaLabelMixin(FormfieldWrapper))) {
```
supminn, line 35:
```suggestion
class Select extends FormInternalsMixin(DataAriaLabelMixin(FormfieldWrapper)) {
```

DisabledMixin is already a part of FormfieldWrapper

packages/components/src/components/select/select.component.ts
```diff
@@ -0,0 +1,363 @@
...
+   */
+  @property({ type: String }) placeholder?: string;
+
+  /**
+   * readonly attribute of the select field. If true, the select is read-only.
+   */
+  @property({ type: Boolean }) readonly = false;
+
+  /**
+   * readonly attribute of the select field. If true, the select is read-only.
```
supminn, line 47: issue: ts-docs is repeated. Please get it updated for this property.

packages/components/src/components/select/select.component.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  /** @internal */
+  @state() showPopover = false;
+
+  /** @internal */
+  @state() activeDescendant = '';
+
+  constructor() {
+    super();
+    this.value = undefined as unknown as string;
```
supminn, line 68: issue: Please follow the new convention to move this into a `connectedCallback`

packages/components/src/components/select/select.component.ts
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  /**
+   * Listens to changes in the default slot and updates the tabindex and selected attribute of the options accordingly.
+   * This is used to set the tabindex and selected attribute of the options when they are slotted.
+   * It is called internally when the slot is changed.
+   */
+  private handleSlotChange(): void {
+    this.updateTabIndexAndSetSelectedOnOptions();
+  }
```
supminn, line 115: nitpick: can we move these 2 functions below. After the definition of `updateTabIndexAndSetSelectedOnOptions`?

packages/components/src/components/select/select.component.ts
```diff
@@ -0,0 +1,363 @@
...
+        this.showPopover = true;
+        this.resetActivedescendant();
+        break;
+      case KEYS.ESCAPE:
+        this.showPopover = false;
+        this.resetActivedescendant();
+        break;
+      case KEYS.ENTER:
+        this.updateTabIndexAndSetSelectedOnOptions(event.target);
+        // if the popover is close, then we submit the form.
```
supminn, line 261: nitpick: typo
```suggestion
        // if the popover is closed, then we submit the form.
```

supminn (changes requested):
First phase of review. Haven't tested it locally.

packages/components/src/components/option/option.styles.ts
supminn: oh, that makes sense. Thanks
It would be great if a comment could be added here.

packages/components/src/components/option/option.stories.ts
```diff
@@ -98,6 +98,14 @@ export const Example: StoryObj = {
   },
 };
 
+export const OptionWithLongText: StoryObj = {
+  render: () => wrapWithDiv(html`
+    <mdc-option value="label">
+      This is a very long text and it should be truncated.
+    </mdc-option>
```
supminn, line 105: issue: on mouse hover the functionality is working as expected.
But, if we focus using keyboard, the tooltip is not being shown.

packages/components/src/components/select/select.component.ts
```diff
@@ -0,0 +1,521 @@
+import type { TemplateResult } from 'lit';
+import { CSSResult, html, nothing } from 'lit';
+import { property, query, queryAssignedElements, state } from 'lit/decorators.js';
+import { KEYS } from '../../utils/keys';
+import { DataAriaLabelMixin } from '../../utils/mixins/DataAriaLabelMixin';
+import { AssociatedFormControl, FormInternalsMixin } from '../../utils/mixins/FormInternalsMixin';
+import FormfieldWrapper from '../formfieldwrapper';
```
supminn, line 7:
```suggestion
import FormfieldWrapper from '../formfieldwrapper/formfieldwrapper.components.ts';
```

packages/components/src/components/select/select.component.ts
```diff
@@ -0,0 +1,521 @@
...
+import type { IconNames } from '../icon/icon.types';
+import { TAG_NAME as OPTION_GROUP_TAG_NAME } from '../optgroup/optgroup.constants';
+import { TAG_NAME as OPTION_TAG_NAME } from '../option/option.constants';
+import { POPOVER_PLACEMENT } from '../popover/popover.constants';
+import { TYPE, VALID_TEXT_TAGS } from '../text/text.constants';
+import { ARROW_ICON } from './select.constants';
+import styles from './select.styles';
+
+/**
+ * select component, represents a control that provides a menu of options.
```
supminn, line 18:
```suggestion
 * select component, represents a control that provides a menu of options. We currently support single option selection.
 * We make use of `mdc-optgroup` and `mdc-option` components to provide a structured way of representing options.
```

packages/components/src/components/select/select.stories.ts
```diff
@@ -0,0 +1,234 @@
...
+    label: 'Select option',
+  },
+  render: (args: Args) => wrapWithDiv(html`
+    <mdc-select height="${args.height}" placeholder="${args.placeholder}" label="${args.label}">
+      ${Array.from({ length: 1000 }, (_, i) => html`<mdc-option label="Option Label ${i + 1}"></mdc-option>`)}
+    </mdc-select>
+  `),
+};
+
+export const SelectWithForm: StoryObj = {
```
supminn, line 200: issue: When an option is selected and "reset" button is clicked, the value gets reset on the select component. However, if you open the drop down, the previously selected option is still marked as selected. Thus, the form gets submitted (incorrectly) when we click on "submit" button. 

Please refer to this screen recording.

https://github.com/user-attachments/assets/82e89801-8bff-4e2f-98e0-1e7c6d700dfb

packages/components/src/components/select/select.stories.ts
supminn, line 1: In all the stories that overrides the render method, please hideControls on the attributes.
Currently, the attributes can be modified. However, modifying them doesn't make any difference.

packages/components/src/components/select/select.styles.ts
supminn, line 1: height of the select component should be 32px, but it is showing 34px.

image

supminn, line 1: Styling issue related to 'readonly' state.

https://github.com/user-attachments/assets/16aae1a6-56d3-44e9-a30b-2896a77f4929

packages/components/src/components/option/option.stories.ts
supminn: https://github.com/user-attachments/assets/5250d28b-ff4a-4362-b02c-39facce5b65d

supminn (changes requested):
Few issues need attention.

packages/components/src/components/select/select.component.ts
```diff
@@ -0,0 +1,533 @@
...
+      return;
+    }
+    this.dispatchEvent(
+      new CustomEvent('change', {
+        detail: { value },
+        composed: true,
+        bubbles: true,
+      }),
+    );
+  }
```
supminn, line 212: suggestion: Instead of dispatching a new event, we could do the following
```
    const EventConstructor = event.constructor as typeof Event;
    this.dispatchEvent(new EventConstructor(event.type, event));
```

arjunkesava: If any option in the select component comes with a default selected event, we will dispatch this change event.
as that is not an mouse / keyboard event, we can't use the event constructor, we can dispatch a new event.

supminn (approved):

PR #1197, opened 2025-04-06
chore: create SideNavigation and add storybook docs

2020Deeya:
…component

### Description

This PR creates SideNavigation that consists of mainly three components:
1] SideNavigation (Host component)
2] NavItemList 
3] NavItem

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-607


cscheffauer (changes requested):
blocking review, this should not be merged

packages/components/src/components/navitem/navitem.styles.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  :host([disabled]), 
+  :host([active][disabled]) {
+    color: var(--mdc-navitem-disabled-color);
+    background-color: var(--mdc-navitem-disabled-background-color);
+  }
+
+  :host([active]) {
+    background-color: var(--mdc-navitem-active-background-color);
+  }
```
supminn, line 56:
```suggestion
  :host([active]) {
    background-color: var(--mdc-navitem-active-background-color);
  }
  
  :host([disabled]) {
    color: var(--mdc-navitem-disabled-color);
    background-color: var(--mdc-navitem-disabled-background-color);
  }
```

If you interchange the order, you don't need the extra `:host([active][disabled])` to be applied

packages/components/src/components/sidenavigation/index.ts
```diff
@@ -0,0 +1,17 @@
+import SideNavigation from './sidenavigation.component';
+import { TAG_NAME } from './sidenavigation.constants';
+import '../navitemlist';
+import '../navitem';
+import '../divider';
+import '../text';
+import '../button';
```
supminn, line 7: would we still need button as an import?
We decided to use ButtonLink or MenuItem right?

packages/components/src/components/sidenavigation/sidenavigation.context.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  // Constructor to allow setting default values
+  constructor(
+    defaultVariant?: string,
+    defaultCustomerName?: string,
+    defaultIsExpanded?: boolean,
+  ) {
+    this.variant = defaultVariant;
+    this.customerName = defaultCustomerName;
+    this.isExpanded = defaultIsExpanded;
```
supminn, line 22: shouldn't we set these values on connectedCallback?

supminn (commented):
initial review comments. Haven't tested it locally

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -0,0 +1,211 @@
...
+                part="label">${this.customerName}</mdc-text>` : nothing}
+              </div>
+          </div>
+        </div>
+        ${this.variant === VARIANTS.FLEXIBLE ? html`<mdc-divider
+            orientation=${DIVIDER_ORIENTATION.VERTICAL}
+            variant=${DIVIDER_VARIANT.GRADIENT}
+            arrow-direction=${this.arrowDirection}
+            button-position=${DIRECTIONS.POSITIVE}
+          > <mdc-button aria-label="Toggle Side Navigation" @click=${this.toggleSideNavigation}></mdc-button>
```
2020Deeya, line 203: I've added an aria-label to the grabber button for accessibility purposes, but it's currently hardcoded. Is that acceptable, or would you recommend a different approach? I'd appreciate your thoughts.

supminn: We should never hard-code it.
We can consider adding an attribute for this. Maybe data-aria-label or grabber-btn-aria-label

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -0,0 +1,211 @@
...
+   * - **fixed-expanded**: Shows icons with labels and has fixed width, 15rem.
+   * - **flexible**: Toggles between collapsed/expanded states.
+   * - **hidden**: Removes the sidenavigation from the DOM.
+   * @default flexible
+   */
+  @property({ type: String, reflect: true })
+  variant: SideNavigationVariant = DEFAULTS.VARIANT;
+
+  /**
+   * Name of the customer
```
supminn, line 75:
```suggestion
   * Name of the customer. This is displayed in the bottom section of the side-navigation component
```

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -0,0 +1,211 @@
...
+   * @default ''
+   */
+  @property({ type: String, reflect: true, attribute: 'customer-name' })
+  customerName: string = '';
```
supminn, line 79: question: should we call this 'customer-name' or 'brand-name'?
Since we have a brand logo next to it

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -0,0 +1,211 @@
...
+    this.arrowDirection = this.arrowDirection === DIRECTIONS.NEGATIVE ? DIRECTIONS.POSITIVE : DIRECTIONS.NEGATIVE;
+    this.updateExpansionState();
+  }
+
+  public override render() {
+    return html`
+        <div part="side-navigation-container">
+          <div part="scrollable-section">
+            <slot name="scrollable-section"></slot>
+          </div>
```
supminn, line 188:
```suggestion
  public override render() {
     if (this.variant === VARIANTS.HIDDEN) {
      return html``;
    }
    return html`
        <div part="side-navigation-container">
          <div part="scrollable-section">
            <slot name="scrollable-section"></slot>
          </div>
```

Should we do this instead of setting 'display: none' on styling for 'hidden' variant?
This would be appropriate according to the docs mentioned (Removes the sidenavigation from the DOM.)

packages/components/src/components/navitem/navitem.component.ts
```diff
@@ -0,0 +1,263 @@
...
+import type { ListItemVariants } from '../listitem/listitem.types';
+import { KEYS } from '../../utils/keys';
+
+/**
+ * `mdc-navitem` is a menuitem styled to work as a navigation tab.
+ * It supports a leading icon, optional badge and dynamic text rendering.
+ *
+ * Note: `mdc-navitem` is intended to be used inside an element with role="menubar" as part of the sideNavigation
+ * component. Its structure, spacing, and interactions are designed to align with
+ * the visual and functional requirements of side navigation layouts.
```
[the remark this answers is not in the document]
2020Deeya: Yes, it was initially `mdc-navItemList`. Since we are not including it in the review process for now, we will update it later when we incorporate the menu-related components as well.

packages/components/src/components/sidenavigation/sidenavigation.component.ts
2020Deeya: Sure we can give it a try and verify that nothing breaks. It's important to ensure that the layout and internal state of child components are preserved without triggering re-creation.

packages/components/src/components/sidenavigation/sidenavigation.stories.ts
```diff
@@ -0,0 +1,225 @@
...
+  tags: ['autodocs'],
+  component: 'mdc-sidenavigation',
+  render,
+  parameters: {
+    badges: ['wip'],
+    docs: {
+      source: {
+        format: 'html',
+        code: `
+        <mdc-sidenavigation customer-name="..." grabber-btn-aria-label="...">
```
2020Deeya, line 164: Please let me know if anything needs to be updated. I didn’t add all possible attributes to keep it clean.

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -0,0 +1,232 @@
...
+ * both `NavItem` and `NavItemList` components.
+ *
+ * @dependency mdc-text
+ * @dependency mdc-button
+ * @dependency mdc-divider
+ *
+ * @event click - (React: onClick) Dispatched when the grabber divider button is clicked.
+ * @event focus - (React: onFocus) Dispatched when the grabber divider button receives focus.
+ * @event keydown - (React: onKeyDown) Dispatched when a key is pressed down on the grabber divider button.
+ * @event keyup - (React: onKeyUp) Dispatched when a key is released on the grabber divider button.
```
supminn, line 39: This is incorrectly linked. If we want to emit these events only on the grabber button,  you'll have to align the listeners only to do that action. 

Focus on side navigation is incorrect. Its a container that has multiple interactive elements which receives focus individually.
 
Similarly, keyup and down events are added on navlist component and not on side navigation.

packages/components/src/components/sidenavigation/sidenavigation.stories.ts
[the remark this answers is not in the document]
arjunkesava: Nice catch, this is happening, because on escape we are not removing current for all navitems, we set it only for the parent nav item.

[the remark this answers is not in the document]
arjunkesava: I've fixed this issue with this: 19c26d5fa80d04ab2ad09c84be373ceb4dbf972b

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -0,0 +1,229 @@
...
+        this.expanded = true;
+        break;
+      case VARIANTS.FIXED_COLLAPSED:
+        this.expanded = false;
+        break;
+      default:
+        return;
+    }
+
+    this.setAttribute('aria-expanded', String(this.expanded));
```
supminn, line 169: `aria-expanded` would be set on the grabber button and not on side-navigation.
In addition to this, we also need to add `aria-controls` and the value should point to the id of side-navigation.

cscheffauer (dismissed):
approving to unblock, but haven't reviewed and check if its actually fine

supminn (dismissed):

cscheffauer (approved):

PR #1200, opened 2025-04-07
fix: prevent buttonsimple space default

cscheffauer:
### Description

- Prevent default when pressing space on the simple button to avoid scrolling behavior when focused on a button and pressing space.


deveshcisco (approved):
Looks good

PR #1210, opened 2025-04-09
fix: issues on search field component

supminn:
### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-616
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-617
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-618
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-619
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-620
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-621


packages/components/src/components/searchfield/searchfield.component.ts
```diff
@@ -27,6 +27,22 @@ class Searchfield extends Input {
...
   @state() isInputFocused = false;
 
+  /**
+   * @internal
+   */
+  @state() hasInputChips = false;
+
+  constructor() {
+    super();
+    this.addEventListener('keydown', this.clearOnEsc);
```
arjunkesava, line 37:
```suggestion
    this.addEventListener('keydown', this.handleKeyDown);
```

packages/components/src/components/searchfield/searchfield.component.ts
```diff
@@ -27,6 +27,22 @@ class Searchfield extends Input {
...
+  }
+
+  private clearOnEsc(e: KeyboardEvent) {
```
arjunkesava, line 40:
```suggestion
  private handleKeyDown(event: KeyboardEvent): void {
```

packages/components/src/components/searchfield/searchfield.component.ts
arjunkesava: This way, in future, if there is any new key combination, then we can write them in the `handleKeyDown` method.

packages/components/src/components/searchfield/searchfield.component.ts
```diff
@@ -27,6 +27,22 @@ class Searchfield extends Input {
    */
   @state() isInputFocused = false;
 
+  /**
+   * @internal
+   */
+  @state() hasInputChips = false;
+
+  constructor() {
```
arjunkesava, line 35: How about adding the even listener call inside the connectedCallback and removing it in the disconnected callback?

packages/components/src/components/searchfield/searchfield.component.ts
```diff
@@ -66,6 +83,14 @@ class Searchfield extends Input {
     };
   }
 
+  override clearInputText() {
+    this.inputChips?.forEach((element) => {
+      // Dispatch the custom 'remove' event from inputChip
+      element.dispatchEvent(new CustomEvent('remove', { bubbles: true, composed: true }));
+    });
+    super.clearInputText();
```
arjunkesava, line 91: Doesn't `super` should be called in the first line of function declaration.

2020Deeya (dismissed):

packages/components/src/components/searchfield/searchfield.component.ts
supminn: Is it mandatory to be called in the first line? I wasn't sure

packages/components/src/components/searchfield/searchfield.component.ts
supminn: I removed this and updated the existing eventlistener function created within the Input component.

PR #1212, opened 2025-04-10
fix: rm forminternalsmixin to fix angular createElement issue

arjunkesava:
### Description

- When consuming the momentum web-components in Angular.
- We are facing issue with select web-component, it is throwing exceptions on angular side as a web component.

```
Uncaught NotSupportedError: Failed to execute 'createElement' on 'Document': The result must not have attributes
    at EmulatedEncapsulationDomRenderer2.createElement (dom_renderer-B-2OOwSx.mjs:517:25)
    at EmulatedEncapsulationDomRenderer2.createElement (dom_renderer-B-2OOwSx.mjs:776:26)
    at createElementNode (core.mjs:11924:21)
    at _locateOrCreateElementNode (core.mjs:27962:12)
    at Module.ɵɵelementStart (core.mjs:27880:20)
    at SelectComponent_Template (select.component.html:26:5)
```

#### Solution:

- The removal of FormInternalsMixin solves the issue in angular.


packages/components/src/components/option/option.component.ts
```diff
@@ -27,7 +27,7 @@ import styles from './option.styles';
  * @event keyup - (React: onKeyUp) This event is dispatched when a key is released on the option.
  * @event focus - (React: onFocus) This event is dispatched when the option receives focus.
  */
-class Option extends FormInternalsMixin(ListItem) {
```
cscheffauer, line 30: The forminternals mixin is there for a reason! We should not just remove it from the Select just cause it doesn't work in Angular. My recommendation would be to find the real issue.

cscheffauer (changes requested):

cscheffauer (approved):

PR #1217, opened 2025-04-15
chore: update scrollbar track and thumb

shakti97:
### Description
 Draft pr for scrollbar update
 The solution works within our component but not globally, I have tried putting the same inside themeprovider, but it didn't work. Creating a draft pr for referrence purpose.


maxinteger:
No activities on the PR, close it for now.

Please re-open when it become relevant again

PR #1220, opened 2025-04-15
fix: refactor link component to use host approach

2020Deeya:
### Description

Refactor Link component to use host approach

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-646


packages/components/src/components/link/link.component.ts
```diff
@@ -63,11 +62,50 @@ class Link extends DisabledMixin(IconNameMixin(Component)) {
...
+  private handleNavigation = (e: MouseEvent | KeyboardEvent): void => {
+    if ((e.type === 'click' || (e instanceof KeyboardEvent && e.key === 'Enter')) && this.href) {
+      if (this.disabled) return;
+      window.open(this.href, this.target ?? '_blank', this.rel ?? 'noopener noreferrer');
+    }
+  };
+
+  public override connectedCallback(): void {
+    super.connectedCallback();
+    this.setDisabled(this.disabled);
```
2020Deeya, line 98: This could be removed.

packages/components/src/components/link/link.component.ts
```diff
@@ -63,11 +62,49 @@ class Link extends DisabledMixin(IconNameMixin(Component)) {
...
+   * so it can be restored after disabling
    * @internal
    */
-  private prevTabindex : number | null = null;
+  private prevTabindex = 0;
+
+  private handleNavigation = (e: MouseEvent | KeyboardEvent): void => {
+    if ((e.type === 'click' || (e instanceof KeyboardEvent && e.key === 'Enter')) && this.href) {
+      if (this.disabled) return;
+      window.open(this.href, this.target ?? '_blank', this.rel ?? 'noopener noreferrer');
```
arjunkesava, line 92: If the target is 'self' by default, then it will not be blank and this condition won't work right.

packages/components/src/components/link/link.component.ts
```diff
@@ -63,11 +62,49 @@ class Link extends DisabledMixin(IconNameMixin(Component)) {
...
+    this.setAttribute('role', 'link');
+    this.addEventListener('click', this.handleNavigation);
+    this.addEventListener('keydown', this.handleNavigation);
+  }
+
+  public override disconnectedCallback(): void {
+    super.disconnectedCallback();
+    this.removeEventListener('click', this.handleNavigation);
+    this.removeEventListener('keydown', this.handleNavigation);
+  }
```
arjunkesava, line 107: I suggest you move these two methods above `handleNavigation` method.

packages/components/src/components/link/link.component.ts
2020Deeya: ya, If the target is 'self' by default, we should not keep any fallback logic here.

supminn (approved):

packages/components/src/components/link/link.component.ts
```diff
@@ -63,11 +62,49 @@ class Link extends DisabledMixin(IconNameMixin(Component)) {
...
+   * Href for navigation. The URL that the hyperlink points to
+   */
+  @property({ type: String, reflect: true })
+  href: string = '#';
+
+  /**
+   * Optional target: _blank, _self, _parent, _top and _unfencedTop
+   */
+  @property({ type: String, reflect: true })
+  target: string = '_self';
```
arjunkesava, line 74:
```suggestion
  @property({ type: String, reflect: true })
  href = '#';

  /**
   * Optional target: _blank, _self, _parent, _top and _unfencedTop
   */
  @property({ type: String, reflect: true })
  target = '_self';
```

When a variable is declared with a default value, then you don't need to provide the type.

arjunkesava (approved):
Minor non blocking comment. Rest al LGTM.

packages/components/src/components/link/link.component.ts
2020Deeya: Okay, I see. But we still add the typeInterface sometimes. Let's take the size attribute as an example — should we remove it here as well, since we're already providing a default value for it?
```
  /**
   * Size of the link.
   * Acceptable values include:
   *
   * - 'small'
   * - 'midsize'
   * - 'large'
   *
   * @default large
   */
  size: LinkSize = DEFAULTS.LINK_SIZE;
```

packages/components/src/components/link/link.stories.ts
```diff
@@ -5,11 +5,7 @@ import { action } from '@storybook/addon-actions';
...
 
-const render = (args: Args) => {
-  // To allow the children to be rendered as html anchor tag, we need to parse it first
-  const htmlContent = html`${new DOMParser().parseFromString(args.children, 'text/html').body.firstChild}`;
-
-  return html`<mdc-link
+const render = (args: Args) => html`<mdc-link
     @click="${action('onclick')}"
     @keydown="${action('onkeydown')}"
     @keyup="${action('onkeyup')}"
```
supminn, line 11:
```suggestion
    @focus="${action('onfocus')}"
     @blur="${action('onblur')}"
```

2020Deeya: Sure, I can add it now. Didn't add it earlier because I was not sure if we really need to add these on storybook.

packages/components/src/components/link/link.component.ts
arjunkesava: For any property which does not have the basic types, you do need to define them.

arjunkesava (approved):

PR #1221, opened 2025-04-17
fix: add overflow for popover on update position

arjunkesava:
### Description

- Add e2e specs for select options dropdown on popover component.
- In select, on keyboard click of `escape` key, we stop the event to go any further the event is closing the popover (when the select is placed inside the popover).


packages/components/src/components/popover/popover.styles.ts
```diff
@@ -28,6 +27,16 @@ const styles = css`
...
   }
 
+  :host([strategy='absolute']) {
+    position: absolute;
+  }
+
+  :host([strategy='fixed']) {
+    position: fixed;
+    /* When position is fixed, the popover goes out of the flow of the document, so we need to set a z-index */
+    z-index: 1000;
```
cscheffauer, line 37: issue: there is the zIndex attribute on the popover itself, which should already cover this one.
We should not hardcode this.

packages/components/src/components/select/select.component.ts
```diff
@@ -82,11 +94,32 @@ class Select extends FormInternalsMixin(DataAriaLabelMixin(FormfieldWrapper)) im
     this.value = undefined as unknown as string;
 
     this.addEventListener('keydown', this.handleKeydown);
+    // Resize won't work on any single element, so we use the window.
+    window.addEventListener('resize', this.handleResize.bind(this));
```
cscheffauer, line 98: issue: this is very expensive - can we do a different approach? also why do you set resize on window?
can't we just set the resize eventlistener on the select itself? this will calculate the with on EVERY resize of the browser, doesn't matter...
so even if we agree to use the resize eventlistener, it should definitely not be on window imho.

packages/components/src/components/tooltip/tooltip.styles.ts
```diff
@@ -12,9 +12,8 @@ const styles = css`
   :host::part(popover-content) {
     padding: var(--mdc-tooltip-padding);
     color: var(--mdc-tooltip-text-color);
-    min-width: fit-content;
-    max-width: var(--mdc-tooltip-max-width);
-
+    min-width: var(--mdc-tooltip-width);
```
cscheffauer, line 15: issue: tooltip can be smaller than 400px - this will make EVERY tooltip exactly 400px - thats incorrect!

packages/components/src/components/popover/popover.component.ts
```diff
@@ -88,6 +88,13 @@ class Popover extends FocusTrapMixin(Component) {
   @property({ type: String, reflect: true })
   placement: PopoverPlacement = DEFAULTS.PLACEMENT;
 
+  /**
+   * The strategy of the popover.
+   * @default absolute
+   */
+  @property({ type: String, reflect: true })
+  strategy: PopoverStrategy = DEFAULTS.STRATEGY;
```
cscheffauer, line 96: issue: we should have a e2e test for this

packages/components/src/components/popover/popover.stories.ts
```diff
@@ -442,3 +445,31 @@ export const popoverWithBackdrop: StoryObj = {
     color: DEFAULTS.COLOR,
   },
 };
+
+export const popoverWithSelect: StoryObj = {
```
cscheffauer, line 449: suggestion: this would be really good to have in a e2e test as well, making sure the select can work inside a popover and that it looks correctly (taking a screenshot)

packages/components/src/components/select/select.constants.ts
```diff
@@ -1,10 +1,16 @@
...
 
 const TAG_NAME = utils.constructTagName('select');
 const ARROW_ICON = {
   ARROW_UP: 'arrow-up-bold' as Extract<IconNames, 'arrow-up-bold'>,
   ARROW_DOWN: 'arrow-down-bold' as Extract<IconNames, 'arrow-down-bold'>,
 } as const;
+const DEFAULTS = {
+  POPOVER_STRATEGY: POPOVER_STRATEGY.ABSOLUTE,
+  POPOVER_WIDTH: '100%',
+  HEIGHT: 'auto',
```
cscheffauer, line 13: issue: this height default is only used in storybook - the constants file is part of the component and the dist.
we should remove constants from here if they are only used in the storybook file (put it inline in the stories file)

cscheffauer (changes requested):
some issues, have not tested it locally - just scanned code for the first round

packages/components/src/components/popover/popover.styles.ts
arjunkesava: When the position is set to fixed, then the element will flyout from the DOM and the z-index will be auto by default.
We need to set the z index to 1000 (which is default) either over here on in the popover.component/utils file.

cscheffauer: disagree, we should NOT hardcode the zIndex - the zIndex is set in the popover.component.ts (line 288) already, why is that not working?

PR #1224, opened 2025-04-17
docs: card component and storybook documentation

supminn:
### Description

This PR creates the following components along with storybook and documentation
* mdc-card - static card that can be made interactive based on the contents passed within the slots
* mdc-cardcheckbox - a selectable card that acts as a checkbox
* mdc-cardradio - a selectable card that acts as a radio
* mdc-cardbutton - a clickable card that acts a button

### Snapshots

1. Static Card
image

2. Horizontal Interactive Card
image

3. Card button (without title and header)
image

4. Card checkbox (in the checked state)
image

5. Card radio (without the image)
image

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-644


packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,205 @@
...
+    *
+    * @internal
+    */
+ private handleFooterSlot = (tagname: string, variant = '') => {
+   let arrayItems: Array<HTMLElement> = [];
+   if (tagname === DEFAULTS.LINK && this.footerLink?.length) {
+     arrayItems = this.footerLink;
+   } else if (tagname === DEFAULTS.BUTTON && variant === 'primary' && this.footerButtonPrimary?.length) {
+     arrayItems = this.footerButtonPrimary;
+   } else if (tagname === DEFAULTS.BUTTON && variant === 'secondary' && this.footerButtonSecondary?.length) {
```
arjunkesava, line 128: Please use constants instead of strings.

```suggestion
   } else if (tagname === DEFAULTS.BUTTON && variant === VARIANT.PRIMARY && this.footerButtonPrimary?.length) {
     arrayItems = this.footerButtonPrimary;
   } else if (tagname === DEFAULTS.BUTTON && variant === VARIANT.SECONDARY && this.footerButtonSecondary?.length) {
```

packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,205 @@
...
+   arrayItems.forEach((element) => {
+     // remove the element if it doesn't match with the tagname
+     if (!element.matches(tagname)) {
+       element.remove();
+     }
+     // set the variant if it is provided
+     if (variant) {
+       element.setAttribute('variant', variant);
+     }
+   });
```
arjunkesava, line 146: Instead of forEach, we can try filter.

Something similar to this:
```
   const filteredArrayItems = arrayItems.filter((element) => !element.matches(tagname));
   if (variant) {
    filteredArrayItems.forEach((element) => {
      element.setAttribute('variant', variant);
    });
   }
```

packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,205 @@
...
+     arrayItems = this.footerButtonPrimary;
+   } else if (tagname === DEFAULTS.BUTTON && variant === 'secondary' && this.footerButtonSecondary?.length) {
+     arrayItems = this.footerButtonSecondary;
+   }
+   // if there are more than one instance, remove them.
+   if (arrayItems.length > 1) {
+     for (let i = 1; i < arrayItems.length; i += 1) {
+       arrayItems[i].remove();
+     }
+   }
```
arjunkesava, line 136: We can try filter here:

```suggestion
arrayItems.filter((_, index) => index === 0);
```

packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,205 @@
...
+           element.setAttribute('size', '32');
+         }
+       });
+     }
+     // limit to show only first 3 buttons defined in the slot
+     if (this.iconButtons && this.iconButtons.length > 3) {
+       for (let i = 3; i < this.iconButtons.length; i += 1) {
+         this.iconButtons[i].remove();
+       }
+     }
```
arjunkesava, line 111:
```suggestion
     this.iconButtons = this.iconButtons.filter((_, index) => index < 3);
```

packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,205 @@
...
+     if (this.iconButtons) {
+       this.iconButtons.forEach((element) => {
+         if (!element.matches(DEFAULTS.BUTTON) && element.getAttribute('data-btn-type') !== 'icon') {
+           element.remove();
+         } else {
+           element.setAttribute('variant', 'tertiary');
+           element.setAttribute('size', '32');
+         }
+       });
+     }
```
arjunkesava, line 105:
```suggestion
     this.iconButtons.filter((element) => 
        !(!element.matches(DEFAULTS.BUTTON) && element.getAttribute('data-btn-type') !== 'icon')
    ).forEach((element) => {
        element.setAttribute('variant', 'tertiary');
        element.setAttribute('size', '32');
    })
```

packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,205 @@
...
+ /**
+  * Updates the color of the footer buttons based on the variant.
+  * If the variant is promotional, the color is promotional, else default.
+  *
+  * @internal
+  */
+ private updateFooterButtonColors = () => {
+   const footerButtons = [...(this.footerButtonPrimary || []), ...(this.footerButtonSecondary || [])];
+   if (footerButtons.length > 0) {
+     footerButtons.forEach((button) => {
```
arjunkesava, line 158: The if condition is not needed, as forEach will not work if the array is empty.

```suggestion
     footerButtons.forEach((button) => {
```

packages/components/src/components/cardcheckbox/cardcheckbox.component.ts
```diff
@@ -0,0 +1,157 @@
...
+   * @default 'check'
+   */
+ @property({ type: String, attribute: 'selection-type' })
+ selectionType: SelectionType = DEFAULTS.SELECTION_TYPE;
+
+ constructor() {
+   super();
+   this.addEventListener('click', this.toggleChecked);
+   this.addEventListener('keydown', this.toggleOnEnter);
+   this.addEventListener('keyup', this.toggleOnSpace);
```
arjunkesava, line 59: Why to register both keydown and up events?
Can't we do something like this.

```suggestion
   this.addEventListener('keydown', this.handleKeyboardNavigation);
```

and in the component below,
```
private handleKeyboardNavigation(event: KeyboardEvent) {
     if (event.key === 'Enter' || event.key === ' ') {
          this.toggleChecked();
     }
}

File: packages/components/src/components/cardcheckbox/cardcheckbox.component.ts [hunk 8]
```diff
@@ -0,0 +1,157 @@
...
+   if (changedProperties.has('disabled')) {
+     this.setAttribute('aria-disabled', `${this.disabled}`);
+   }
+ }
+
+ /**
+  * Toggles the checked state
+  */
+ private toggleChecked() {
+   this.checked = !this.checked;
```
arjunkesava, line 81: Won't we need to dispatch the event here?

packages/components/src/components/cardradio/cardradio.component.ts
```diff
@@ -0,0 +1,134 @@
...
+   * @default ''
+   */
+ @property({ type: String })
+ name: string = '';
+
+ constructor() {
+   super();
+   this.addEventListener('click', this.toggleChecked);
+   this.addEventListener('keydown', this.toggleOnEnter);
+   this.addEventListener('keyup', this.toggleOnSpace);
```
arjunkesava, line 57: same here.

packages/components/src/components/cardradio/cardradio.component.ts
```diff
@@ -0,0 +1,134 @@
...
+ * @event focus - (React: onFocus) Event that gets dispatched when the card receives focus.
+ *
+ */
+class CardRadio extends DisabledMixin(TabIndexMixin(Card)) {
+ /**
+   * The checked state of the card
+   * @default false
+   */
+ @property({ type: Boolean, reflect: true })
+ checked: boolean = false;
```
arjunkesava, line 44: For primitive types, we don't need to define the type when you have a default value assigned.
```suggestion
 checked = false;
```

packages/components/src/components/cardradio/cardradio.component.ts
```diff
@@ -0,0 +1,134 @@
...
+
+ /**
+   * The name of the radio.
+   * @default ''
+   */
+ @property({ type: String })
+ name: string = '';
```
arjunkesava, line 51:
```suggestion
 name = '';
```

packages/components/src/index.ts
```diff
@@ -68,6 +72,10 @@ export {
   Badge,
   Bullet,
   Button,
+  Card,
+  Cardbutton,
```
arjunkesava, line 76:
```suggestion
  CardButton,
```

packages/components/src/index.ts
```diff
@@ -68,6 +72,10 @@ export {
...
+  CardCheckbox,
+  Cardradio,
```
arjunkesava, line 78:
```suggestion
  CardRadio,
```

packages/components/src/utils/mixins/CardComponentMixin.ts
```diff
@@ -0,0 +1,132 @@
...
+   */
+  @property({ type: String, attribute: 'image-src' })
+  imageSrc: string = '';
+
+  /**
+   * The image alt for accessibility support
+   * @default ''
+   */
+  @property({ type: String, attribute: 'image-alt' })
+  imageAlt: string = '';
```
arjunkesava, line 56:
```suggestion
  cardTitle = '';

  /**
   * The subtitle of the card - part of header section
   * @default ''
   */
  @property({ type: String })
  subtitle = '';

  /**
   * The image source URL to render on the card
   * @default ''
   */
  @property({ type: String, attribute: 'image-src' })
  imageSrc = '';

  /**
   * The image alt for accessibility support
   * @default ''
   */
  @property({ type: String, attribute: 'image-alt' })
  imageAlt = '';
```

arjunkesava (changes requested):
Haven't tested anything locally.

- Please avoid using `remove` and replace it with filter.
- Replace all Cardxxx classnames with CardXxx.

packages/components/src/components/card/card.component.ts
arjunkesava: As the `iconButtons` is readonly property which is derived from `queryAssignedElements`.
We can't assign value to it. We can only remove() it.

packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,206 @@
...
+           element.setAttribute('size', '32');
+         }
+       });
+     }
+     // limit to show only first 3 buttons defined in the slot
+     if (this.iconButtons && this.iconButtons.length > 3) {
+       for (let i = 3; i < this.iconButtons.length; i += 1) {
+         this.iconButtons[i].remove();
+       }
+     }
```
arjunkesava, line 111:
```suggestion
     this.iconButtons?.forEach((element, index) => {
       // if the element is not a button AND data-btn-type is not an icon OR if there are more than 3 elements, we remove the remaining items.
        if ((!element.matches(DEFAULTS.BUTTON) && element.getAttribute('data-btn-type') !== 'icon') || index < 3) {
            element.remove();
        } else {
            element.setAttribute('variant', BUTTON_VARIANTS.TERTIARY);
            element.setAttribute('size', '32');
        }
    });
```

packages/components/src/components/cardcheckbox/cardcheckbox.component.ts
supminn: If you notice, they are 2 different functions that fire toggle on different keys.

On keydown, the toggle is fired through "enter" key. 
On keyup, toggle is fired using "space" key.

This is how the native button as well as mdc-buttonSimple behaves. Hence, adding the same behavior over here.

packages/components/src/components/cardradio/cardradio.stories.ts
```diff
@@ -0,0 +1,218 @@
...
+    disabled: false,
+    tabIndex: 0,
+    children: html`
+    <mdc-text slot='before-body' type="body-midsize-medium" tagname="span">Content Before Body</mdc-text>
+    <img src="https://placehold.co/100x50" alt="Image Alt" slot="before-body"/>
+    ${defaultChildren}`,
+  },
+};
+
+export const ContentAfterBody: StoryObj = {
```
2020Deeya, line 199: question/issue: Are we okay with this layout when the user provides this much content here?
image

packages/components/src/components/cardradio/cardradio.stories.ts
```diff
@@ -0,0 +1,218 @@
...
+    'subtitle-tag-name': DEFAULTS.TAGNAME,
+    checked: false,
+    disabled: false,
+    tabIndex: 0,
+    variant: DEFAULTS.VARIANT,
+    orientation: DEFAULTS.ORIENTATION,
+  },
+};
+
+export const ContentBeforeBody: StoryObj = {
```
2020Deeya, line 178: question/issue: Are we okay with this layout when the user provides this much content here?
image

packages/components/src/components/cardbutton/cardbutton.stories.ts
```diff
@@ -0,0 +1,190 @@
...
+    icon-name="${args['icon-name']}"
+    title-tag-name="${args['title-tag-name']}"
+    subtitle-tag-name="${args['subtitle-tag-name']}"
+    type="${args.type}"
+    role="${args.role}"
+    tabIndex="${args.tabIndex}"
+    class="${args.class}"
+    style="${args.style}">${args.children}</mdc-cardbutton>`;
+
+const meta: Meta = {
```
2020Deeya, line 33: question: Noticed common button props like disabled, active, size, etc. are missing—any reason they weren’t included?

packages/components/src/components/cardbutton/index.ts
```diff
@@ -0,0 +1,12 @@
+import CardButton from './cardbutton.component';
+import { TAG_NAME } from './cardbutton.constants';
+
```
2020Deeya, line 3: question: Why haven't we imported the icon and text here? Is it because the CardButton component is extending ButtonSimple, which internally imports them already?

packages/components/src/components/cardradio/cardradio.stories.ts
supminn: Yes, they would need to appropriately adjust the size of the image. 
In this story, I have set the image placeholder's height to 540, they'd need to change that to support an image which is longer.

packages/components/src/components/cardbutton/index.ts
supminn: thanks for pointing this out, I have added them now

packages/components/src/components/cardradio/cardradio.stories.ts
supminn: Please suggest how can we handle this?

packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,204 @@
...
+    * @internal
+    */
+   @queryAssignedElements({ slot: 'footer-button-primary' })
+   footerButtonPrimary?: Array<HTMLElement>;
+
+   /**
+    * The secondary buttons in the footer section
+    * @internal
+    *
+    */
```
arjunkesava, line 79:
```suggestion
    */
```

packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,204 @@
...
+   }
+
+   /**
+    * Handles the icon buttons in the header section and sets its variant for styling.
+    * It also limits the number of buttons to 3.
+    * @internal
+    */
+   private handleIconButtons = () => {
+     if (this.iconButtons) {
+       this.iconButtons.forEach((element) => {
```
arjunkesava:
```suggestion
     this.iconButtons?.forEach((element) => {
```

packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,204 @@
...
+           element.remove();
+         } else {
+           element.setAttribute('variant', BUTTON_VARIANTS.TERTIARY);
+           element.setAttribute('size', '32');
+         }
+       });
+     }
+     // limit to show only first 3 buttons defined in the slot
+     if (this.iconButtons && this.iconButtons.length > 3) {
+       for (let i = 3; i < this.iconButtons.length; i += 1) {
```
arjunkesava, line 106:
```suggestion
       for (let i = 3; i < this.iconButtons?.length; i += 1) {
```

packages/components/src/components/card/card.component.ts
```diff
@@ -0,0 +1,204 @@
...
+     arrayItems = this.footerLink;
+   } else if (tagname === DEFAULTS.BUTTON && variant === BUTTON_VARIANTS.PRIMARY && this.footerButtonPrimary?.length) {
+     arrayItems = this.footerButtonPrimary;
+   } else if (tagname === DEFAULTS.BUTTON
+    && variant === BUTTON_VARIANTS.SECONDARY && this.footerButtonSecondary?.length) {
+     arrayItems = this.footerButtonSecondary;
+   }
+   // if there are more than one instance, remove them.
+   if (arrayItems.length > 1) {
+     for (let i = 1; i < arrayItems.length; i += 1) {
```
arjunkesava, line 132:
```suggestion
     for (let i = 1; i < arrayItems?.length; i += 1) {
```

packages/components/src/components/card/index.ts
```diff
@@ -0,0 +1,14 @@
+import Card from './card.component';
+import { TAG_NAME } from './card.constants';
+import '../icon';
+import '../text';
```
arjunkesava, line 4: Please organize these imports and other imports across all files in this PR.

arjunkesava (changes requested):

packages/components/src/utils/mixins/CardComponentMixin.ts
```diff
@@ -0,0 +1,132 @@
...
+         : nothing;
+     }
+
+     /**
+    * Renders the title and subtitle on the card
+    * @returns The title and subtitle elements
+    */
+     renderTitle() {
+       return html`<div part="title">
+        <mdc-text part="title" type="${DEFAULTS.TITLE_TYPE}" tagname="${this.titleTagName}">${this.cardTitle}</mdc-text>
```
arjunkesava, line 123: Question: the div and the mdc-text both have the same part "title", so if I apply any styles through part, then I will be applying for both of them. Is this valid ?

packages/components/src/components/card/index.ts
arjunkesava: In the long run, when you get back to read these files or want to understand, organized imports will help you align with sorting the imports based on there import structure.

It's a better way to maintain the code. If you think organizing all imports take time, then you can do this in your next PR, this is not a blocker for me.

arjunkesava (approved):

PR #1225, opened 2025-04-17
docs: create buttonLink component and add storybook docs

2020Deeya:
### Description

1] Create ButtonLink Component (a Link which looks like a Button)
2] Add Storybook docs
3] Add E2E tests for ButtonLink
4] Refine E2E tests for Linksimple and Link

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-647


packages/components/src/components/buttonlink/buttonlink.component.ts
```diff
@@ -0,0 +1,177 @@
+import { CSSResult, html, PropertyValues } from 'lit';
+import { property, state } from 'lit/decorators.js';
+import styles from './buttonlink.styles';
+import Link from '../link';
```
supminn, line 4:
```suggestion
import Link from '../link/link.component';
```

packages/components/src/components/buttonlink/buttonlink.component.ts
```diff
@@ -0,0 +1,177 @@
...
+  BUTTON_TYPE_INTERNAL,
+  BUTTON_VARIANTS,
+  DEFAULTS,
+  ICON_BUTTON_SIZES,
+  PILL_BUTTON_SIZES,
+} from '../button/button.constants';
+import type { IconNames } from '../icon/icon.types';
+import type { ButtonColor, ButtonVariant, IconButtonSize, PillButtonSize } from '../button/button.types';
+import type { LinkSize } from '../link/link.types';
+import Button from '../button/button.component';
```
supminn, line 16: Should we import the styles file directly from button instead of the component file?
```suggestion
import buttonStyles from '../button/button.styles';
```

packages/components/src/components/buttonlink/buttonlink.component.ts
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * @tagname mdc-buttonlink
+ *
+ * @event click - (React: onClick) Fired when the user activates the buttonLink using a mouse or assistive technology.
+ * @event keydown - (React: onKeyDown) Fired when the user presses a key while the buttonLink has focus.
+ * @event focus - (React: onFocus) Fired when the buttonLink receives keyboard or mouse focus.
+ * @event blur - (React: onBlur) Fired when the buttonLink loses keyboard or mouse focus.
+ *
+ */
+class Buttonlink extends Link {
```
supminn, line 33:
```suggestion
class ButtonLink extends Link {
```

Please make this change across all instances. We have this as a part of our [conventions](https://github.com/momentum-design/momentum-design/blob/main/packages/components/conventions/component-naming-guidelines.md).

packages/components/src/components/buttonlink/buttonlink.component.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * ButtonLink sizing is based on the buttonLink type.
+   * - **Pill buttonLink**: 40, 32, 28, 24.
+   * - **Icon buttonLink**: 64, 52, 40, 32, 28, 24.
+   * - Tertiary icon buttonLink can also be 20.
+   * @default 32
+   */
+  @property({ type: Number, reflect: true, attribute: 'buttonlink-size' })
+  buttonlinkSize: PillButtonSize | IconButtonSize = DEFAULTS.SIZE;
```
supminn, line 63: question: why do we want to create this new property?
Why not reuse 'size' property itself?

packages/components/src/components/buttonlink/buttonlink.stories.ts
```diff
@@ -0,0 +1,152 @@
...
+  ?disabled="${args.disabled}"
+  prefix-icon="${args['prefix-icon']}"
+  postfix-icon="${args['postfix-icon']}"
+  buttonlink-size="${args['buttonlink-size']}"
+  href="${args.href}"
+  target="${args.target}"
+  rel="${args.rel}"
+  variant="${args.variant}"
+  color="${args.color}"
+  type="${args.type}"
```
supminn, line 25: type attribute is not supported on the component, please get it removed from the stories.

packages/components/src/components/buttonlink/buttonlink.component.ts
2020Deeya: TypeScript does not allow overriding the type of an inherited attribute when extending a class, due to language constraints.

packages/components/src/components/buttonlink/buttonlink.component.ts
2020Deeya: Agreed but, I intentionally went with the name Buttonlink for these reasons:
✅ Buttonlink → “A link that looks like a button” – correct intent, simple and smooth naming.
🚫 ButtonLink → “A half-button, half-link” – misleading, as it overemphasizes the ‘button’ aspect, whereas the core behavior is still that of a link.

packages/components/src/components/buttonlink/buttonlink.component.ts
2020Deeya: Yes, we can go with this approach, but we haven’t been following it consistently so far. If we decide to adopt it, we should apply the change across all relevant components to maintain consistency.

packages/components/src/components/buttonlink/buttonlink.component.ts
supminn: Since we have this as a convention, we'd need to follow it for maintaining consistency.
I differ to agree that by changing the case, it would provide a different meaning.
We are doing the same on `AvatarButton` too. This component is intended to be a button and look like an avatar.

packages/components/src/index.ts
```diff
@@ -106,6 +107,7 @@ export {
   Select,
   TabList,
   Progressspinner,
+  Buttonlink,
```
arjunkesava, line 110: Please change this to `ButtonLink` and also update all classnames of the buttonlink.

packages/components/src/components/buttonlink/buttonlink.component.ts
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Button sizing is based on the button type.
+    * - **Pill button**: 40, 32, 28, 24.
+    * - **Icon button**: 64, 52, 40, 32, 28, 24.
+    * - Tertiary icon button can also be 20.
+    * @default 32
+    */
+  /* Note: TypeScript does not allow overriding the type of an inherited attribute when extending a class,
+   so we've kept it as `any` (workaround for now) */
+  @property({ type: Number, reflect: true }) override size: any = DEFAULTS.SIZE;
```
arjunkesava, line 32: This should be fixed.

Instead of setting it to `any`.
Please refactor the link component link size constant.
In link component, the link size is used only for icon size.
```
private getIconSize(): number {
    switch (this.size) {
      case LINK_SIZES.SMALL:
        return LINK_ICON_SIZES.SMALL;
      case LINK_SIZES.MIDSIZE:
        return LINK_ICON_SIZES.MIDSIZE;
      default:
        return LINK_ICON_SIZES.LARGE;
    }
  }
  ```
  
  By removing this and replacing it with size const, then we can fix this type any.
  ```
  const LINK_SIZES = {
  16: 16,
  14: 14,
  12: 12,
} as const;
```

Using the type `any` is dangerous, as the size number can be anything and we need to filter size for every time.

arjunkesava (changes requested):

packages/components/src/utils/mixins/ButtonComponentMixin.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  protected override render() {
+    return html`
+      ${this.prefixIcon ? html` <mdc-icon name="${this.prefixIcon as IconNames}" part="prefix-icon"></mdc-icon>` : ''}
+      <slot @slotchange=${this.inferButtonType}></slot>
+      ${this.postfixIcon
+    ? html` <mdc-icon name="${this.postfixIcon as IconNames}" part="postfix-icon"></mdc-icon>`
+    : ''}
+    `;
+  }
```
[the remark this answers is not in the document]
2020Deeya: Initially, we kept the render method in the mixin because Button and ButtonLink had very similar render logic, so it made sense for reuse and could always be overridden if needed in components extending the mixin.

But yeah, with the latest commit ([#1230](https://github.com/momentum-design/momentum-design/pull/1230/files)), their render methods have started to diverge. So agreed — it's better not to keep it in the mixin anymore.

packages/components/src/components/buttonlink/buttonlink.component.ts
2020Deeya: Totally agree any isn’t ideal. But ButtonLink extends Link and needs to look like a button, so button-specific sizing is required — not LINK_SIZES.

We had two options:
1] Override size with `any` which allowed us to decouple from the Link-specific size values.
2] Add a new attribute just for button size, which felt redundant.

After @supminn and I discussed this, We chose option 1 for simplicity. Using LINK_SIZES would misalign with the intended button appearance.

packages/components/src/components/buttonlink/buttonlink.component.ts
2020Deeya: Initially, I felt AvatarButton should also be named as Avatarbutton as per my reasoning. 
Okay I understand, it's not very much important for consumers though so let's stick to what we have defined so far.

arjunkesava (approved):
LGTM

PR #1231, opened 2025-04-22
feat: introduce dialog component

zachraymer:
### Description

Introduces the dialog component, along with storybook and playwright testing. The dialog component is a modal dialog with a partially-opaque fullscreen backdrop. Focus is trapped within the dialog when it is visible. Unlike popover, trigger actions are not delegated; visibility of the dialog is controlled by the consumer with the `?visible` boolean attribute. It accepts an optional title and header. Body and footer content are provided with individual slots.

image

This PR also has updates on close-button position on popover. With this change, the snapshots for coachmark and toggletip needed to be updated.

### Links

https://confluence-eng-gpk2.cisco.com/conf/pages/viewpage.action?pageId=646959672
https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-625


packages/components/src/components/dialog/dialog.styles.ts
```diff
@@ -0,0 +1,103 @@
...
+    box-sizing: content-box;
+    justify-content: flex-end;
+  }
+
+  .dialog-close {
+    position: absolute;
+    right: 1rem;
+    top: 1rem;
+    cursor: pointer;
+    z-index: 9999;
```
supminn, line 95: issue: This is incorrect. We shouldn't hard code the zIndex value. Instead you could do something similar to the backdrop.
```
        z-index: ${this.dialog.zIndex + 1};
```

packages/components/src/components/dialog/dialog.utils.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!this.dialog.ariaLabel) {
+      if (this.dialog.headerText) this.dialog.ariaLabel = this.dialog.headerText;
+      else if (!this.dialog.ariaLabelledById) this.dialog.ariaLabelledById = this.dialog.triggerElement?.id || '';
+    }
+  }
+
+  createBackdrop() {
+    const backdrop = document.createElement('div');
+    backdrop.classList.add('dialog-backdrop');
+    this.dialog.parentElement?.appendChild(backdrop);
```
supminn, line 25: suggestion: You can move this line after appending styleElement to backdrop.
After line 39.

packages/components/src/components/dialog/dialog.styles.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+    
+  :host([visible]) {
+    display: flex;
+      
+  }
+
+  :host([size='medium']) {
+    --mdc-dialog-width: 41rem;
+    --mdc-dialog-height: 22.25rem;
```
supminn, line 36: issue: we will not restrict the height of the component. The size  value will only restricts the width.

packages/components/src/components/dialog/dialog.styles.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  :host([size='large']) {
+    --mdc-dialog-width: 62rem;
+    --mdc-dialog-height: 25.75rem;
```
supminn, line 41: same as above

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -0,0 +1,240 @@
...
+  args: {
+    id: 'dialog',
+    'z-index': DEFAULTS.Z_INDEX,
+    role: DIALOG_ROLE[1],
+    size: DIALOG_SIZE[0],
+    triggerId: 'dialog-trigger-btn',
+    'close-button-aria-label': 'Close dialog',
+    visible: false,
+  },
+};
```
supminn, line 240: suggestion: Can we add another story where the dialog would have "Cancel" and "Save" as buttons.
Clicking on "Save" will close dialog. However, clicking on "Cancel" should open an alert window and ask for confirmation. Once confirmed, the dialog window closes.

packages/components/src/components/dialog/dialog.component.ts
```diff
@@ -0,0 +1,317 @@
...
+/**
+ * Dialog component is a modal dialog that can be used to display information or prompt the user for input.
+ * It can be used to create custom dialogs where content for the body and footer actions is provided by the consumer.
+ * The dialog is available in three sizes: small, medium, and large. It may also receive custom styling/sizing.
+ * The dialog interrupts the user and will block interaction with the rest of the application until it is closed.
+ *
+ * **Accessibility notes for consuming (have to be explicitly set when you consume the component)**
+ *
+ * - If the dialog is provided with header text, it will be set as the aria-label string value if aria-label
+ *   is not provided.
```
supminn, line 22: issue: This is incorrect. We should not set the fallback values.

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -0,0 +1,240 @@
...
+        <mdc-button variant="secondary" @click="${toggleVisibility}">Secondary</mdc-button>
+        <mdc-button variant="primary" @click="${toggleVisibility}">Primary</mdc-button>
+      </div>`)}
+  `;
+};
+
+const renderNoFooter = (args: Args) => {
+  const toggleVisibility = () => {
+    const dialog = document.getElementById(args.id) as HTMLElement;
+    dialog.toggleAttribute('visible');
```
supminn, line 71: Instead of doing this, can we have 2 functions (similar to popover) for 'showDialog' and 'hideDialog'.
This abstraction pattern would be consistent to maintain across components.

packages/components/src/components/dialog/dialog.component.ts
```diff
@@ -0,0 +1,317 @@
...
+      changedProperties.has('ariaLabel')
+      || changedProperties.has('ariaLabelledById')
+    ) {
+      this.utils.setupAccessibility();
+    }
+  }
+
+  /**
+   * Handles the escape keydown event to close the dialog.
+   *
```
supminn, line 209:
```suggestion
   * @internal
```

This would hide the function on storybook.

packages/components/src/components/dialog/dialog.component.ts
```diff
@@ -0,0 +1,317 @@
...
+      this.deactivateFocusTrap?.();
+      this.triggerElement.removeAttribute('aria-expanded');
+      this.triggerElement.removeAttribute('aria-haspopup');
+      this.triggerElement?.focus();
+      DialogEventManager.onHideDialog(this);
+    }
+  }
+
+  /**
+   * Hides the dialog.
```
supminn, line 259:
```suggestion
   * Hides the dialog.
   * @internal
```

supminn (changes requested):
First phase of review

packages/components/src/components/dialog/dialog.component.ts
zachraymer: how can we make properties required? .. or should we here in this case ?

packages/components/src/components/dialog/dialog.stories.ts
zachraymer: can you point to that example, I dont see it in popover

supminn: You already have a function to hideDialog, I was suggesting to use a similar function to 'showDialog' too.

packages/components/src/components/dialog/dialog.e2e-test.ts
```diff
@@ -0,0 +1,207 @@
...
+    });
+
+    await test.step('accessibility attributes', async () => {
+      await expect(dialog).toHaveAttribute('aria-label', 'dialog-attribute');
+      await expect(dialog).not.toHaveAttribute('aria-expanded');
+      await expect(dialog).not.toHaveAttribute('aria-describedby');
+
+      const closeDialogButton = componentsPage.page.locator('.dialog-close');
+      await expect(closeDialogButton).toHaveAttribute('aria-label', 'Close button label');
+    });
```
supminn, line 141: suggestion: Could you add tests to check for other attributes as well?
size, header, description, aria-labelledby are all missing.

packages/components/src/components/dialog/dialog.e2e-test.ts
```diff
@@ -0,0 +1,207 @@
...
+  /**
+   * INTERACTIONS
+   */
+  await test.step('interactions', async () => {
+    await test.step('programmatic', async () => {
+      await test.step('dialog should close/open when the visible attribute is changed', async () => {
+        const { dialog } = await setup({ componentsPage, ...dialogWithAllSlots });
+        await expect(dialog).toBeVisible();
+        await dialog.evaluate((dialog) => {
+          dialog.removeAttribute('visible');
```
supminn, line 153: suggestion: rather than doing this, we could use `dialog.hideDialog()` which does the same thing.

packages/components/src/components/dialog/dialog.e2e-test.ts
```diff
@@ -0,0 +1,207 @@
...
+        const { dialog } = await setup({ componentsPage, ...dialogWithAllSlots, visible: false });
+        await dialog.evaluate((dialog) => {
+          dialog.toggleAttribute('visible');
+        });
+        await expect(dialog).toBeVisible();
+        const closeButton = componentsPage.page.locator('.dialog-close');
+        await expect(closeButton).toBeFocused();
+        await componentsPage.page.keyboard.press('Tab');
+        await expect(closeButton).not.toBeFocused();
+        await componentsPage.page.keyboard.press('Shift+Tab');
```
supminn, line 179:
```suggestion
        await expect(closeButton).toBeFocused();
        await componentsPage.actionability.presssTab();
        await expect(closeButton).not.toBeFocused();
        await componentsPage.actionability.pressShiftTab();
```

packages/components/src/components/dialog/dialog.e2e-test.ts
```diff
@@ -0,0 +1,207 @@
...
+        await dialog.evaluate((dialog) => {
+          dialog.toggleAttribute('visible');
+        });
+        await expect(dialog).toBeVisible();
+        const closeButton = componentsPage.page.locator('.dialog-close');
+        await expect(closeButton).toBeFocused();
+        await componentsPage.page.keyboard.press('Tab');
+        await expect(closeButton).not.toBeFocused();
+        await componentsPage.page.keyboard.press('Tab');
+        await componentsPage.page.keyboard.press('Tab');
```
supminn, line 202:
```suggestion
         await componentsPage.actionability.pressTab();
        await expect(closeButton).not.toBeFocused();
        await componentsPage.actionability.pressTab();
         await componentsPage.actionability.pressTab();
```

supminn (changes requested):

packages/components/src/components/dialog/dialog.component.ts
```diff
@@ -0,0 +1,310 @@
...
+      return;
+    }
+
+    if (newValue) {
+      this.enabledFocusTrap = true;
+      this.enabledPreventScroll = true;
+      this.utils.createBackdrop();
+
+      await this.handleCreateDialogFirstUpdate();
+      document.addEventListener('keydown', this.onEscapeKeydown);
```
supminn, line 231: suggestion: could we add this listener on the constructor instead of adding and removing it here.

packages/components/src/components/dialog/dialog.component.ts
```diff
@@ -0,0 +1,310 @@
...
+    if (!this.triggerElement) return;
+
+    this.removeEventListener('focus-trap-exit', this.hideDialog);
+  }
+
+  protected override async updated(changedProperties: PropertyValues) {
+    super.updated(changedProperties);
+
+    if (changedProperties.has('visible')) {
+      const oldValue = (changedProperties.get('visible') as boolean | undefined) || false;
```
supminn, line 191: we should not set the fallback to false. If it is undefined, we need to handle it appropriately on `isOpenUpdated` function

packages/components/src/components/dialog/dialog.stories.ts
zachraymer: Im not sure I see why we should do this, it is expected that consumer will show and hide dialog with visible attribute. The reason there is a hideDialog function is because there are internal actions such as x button and escape button that should close the dialog but the logic for these actions is inside the component. There is nothing inside the component that would show it, because it is attribute driven.

packages/components/src/components/dialog/dialog.e2e-test.ts
zachraymer: The test is testing whether or not the the attribute successfully closes the dialog, we should not use the hideDialog method.

PR #1235, opened 2025-04-23
test: tablist e2e test cases

supminn:
### Description

E2E test cases for tablist component

### Links

https://jira-eng-gpk2.cisco.com/jira/browse/MOMENTUM-534


packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -166,8 +153,6 @@ class TabList extends Component {
       }
     }
 
-    this.setAriaLabelledByOrAriaLabel();
-
     const resizeObserver = new ResizeObserver(async () => {
       const activeElement = this.tabsContainer?.shadowRoot?.activeElement;
```
deveshcisco, line 157:
```suggestion
      const activeElement = document?.activeElement;
```
Accessing the shadowroot's active element returns `null` here and is a bug. This selector should access the document's active element instead

packages/components/src/components/tablist/tablist.component.ts
deveshcisco, line 1: As per https://web.dev/articles/custom-elements-best-practices#events, it is a better practice to emit events in one direction only.

I suggest moving the `this.fireTabChangeEvent(newTab);`  from the `update` method to the `handleNestedTabActiveChange ` method.

At present a change event will fire even if someone changes the active tab id attribute, which is redundant because the host already knows the new tab id

deveshcisco, line 1: In place of the payload of the change event being `detail: newTab.tabId,`, it should be 

```js
detail: {
        tabId: newTab.tabId,
      },
```

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  /**
+   * INTERACTIONS
+   */
+  await test.step('interactions', async () => {
+    await test.step('resize', async () => {
+      await tabs.nth(2).click();
+
+      await test.step('resizing the viewport to 320px width should show atleast one arrow button', async () => {
+        await componentsPage.page.setViewportSize({ width: 320, height: 600 });
```
deveshcisco, line 181: This will not work on all the devices

---  


Playwright runs with its own viewport, so the setViewPortSize function doesn't always trigger the resize event.

This causes bugs in tests where we rely on the resize event to trigger.

see https://github.com/microsoft/playwright/issues/20721#issuecomment-1421644560

---

Instead resize by changing the mdcTablist widths

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,315 @@
...
+        await expect(arrowButtons.first()).toBeVisible();
+      });
+
+      await test.step('resizing the viewport to 2000px width should remove both arrow buttons', async () => {
+        await componentsPage.page.setViewportSize({ width: 2000, height: 600 });
+        await expect(arrowButtons).toHaveCount(0);
+      });
+      await componentsPage.page.setViewportSize({ width: 800, height: 600 });
```
deveshcisco, line 189:
```suggestion
      await componentsPage.page.setViewportSize({ width: 800, height: componentsPage.page.viewportSize().width });
```

If you decide to resize the viewport , then instead of a hardcoded height, keep the height of the original viewport

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,315 @@
...
+          expect(xAxisPositionInitial).not.toEqual(xAxisPositionFinal);
+        }
+        await componentsPage.page.setViewportSize({ width: 800, height: 600 });
+      });
+    });
+
+    await test.step('keyboard', async () => {
+      await test.step('component should change active tab on pressing space/enter', async () => {
+        await tabs.first().focus();
+        await componentsPage.page.keyboard.press('Enter'); // activate this tab.
```
deveshcisco, line 223: Also needs a test case for the `Space` key

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,315 @@
...
+
+    await test.step('focus', async () => {
+      await test.step('right arrow should focus next tab', async () => {
+        await tabs.nth(2).focus();
+        await componentsPage.page.keyboard.press('ArrowRight');
+        await expect(tabs.nth(3)).toBeFocused();
+      });
+
+      await test.step('left arrow should focus previous tab', async () => {
+        await componentsPage.page.keyboard.press('ArrowLeft');
```
deveshcisco, line 238: This test should also have a `await tabs.nth(3).focus();` as the first action.

This case is depending on the previous case to have the 4th tab focused, which will break the test case if any order is shuffled

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,315 @@
...
+
+    // variant: icon only tabs
+    stickerSheet.setChildren(renderTabs(TAB_VARIANTS.PILL, true));
+    await stickerSheet.createMarkupWithCombination({});
+
+    await stickerSheet.mountStickerSheet();
+    const container = stickerSheet.getWrapperContainer();
+
+    await test.step('matches screenshot of element', async () => {
+      await componentsPage.visualRegression.takeScreenshot('mdc-tablist', { element: container });
```
deveshcisco, line 123:
```suggestion
            await componentsPage.visualRegression.takeScreenshot('mdc-tablist', {
        element: container,
        assertionAfterSwitchingDirection: async (page) => {
          await page.evaluate(() => {
            const tablists = Array.from(document.querySelectorAll('mdc-tablist') || []);

            tablists.forEach((tablist, index) => {
              const tabs = Array.from(tablist.querySelectorAll('mdc-tab')) as Tab[];
              tabs[0]?.click();
              tabs[1]?.click();
            });
          });
        },
```

The `assertionAfterSwitchingDirection ` call is needed because the arrows only re-render when the component state changes.
In an ideal scenario, the direction of the page is not changed on the fly. It is usually preset.
However in this case we need to force a re-render to get the correct arrow icon to show. Otherwise there will be flipped arrow buttons.

packages/components/src/components/tablist/tablist.component.ts
deveshcisco, line 1: All calls to the focus method need to have the focusVisible option set to true, otherwise due to an inconsistency in Webkit's implementation, the focus outline does not return to the tabs on keyboard interactions.

```js
focus(
        {
          preventScroll: true,
          focusVisible: true,
        },
      );
```

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -64,18 +61,8 @@ class TabList extends Component {
   activeTabId?: string;
 
   /**
-   * Selector for the element that provides a label for the tablist.
-   */
-  @property({ type: String, attribute: 'data-aria-labelledby' })
-  dataAriaLabelledby?: string;
```
deveshcisco, line 70: Why is the `aria-labelledby` property removed?

The data- approach is needed because the tablist is a child of the mdc-tablist component

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,315 @@
...
+    `,
+    clearDocument: true,
+  });
+
+  const mdcTablist = componentsPage.page.locator('mdc-tablist');
+  await mdcTablist.waitFor();
+
+  return mdcTablist;
+};
+
```
deveshcisco, line 95: Are the event listener tests not needed?

deveshcisco (commented):
Left some notes

packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -572,7 +537,11 @@ class TabList extends Component {
         : nothing}`;
 
     return html` ${arrowButton('backward')}
-      <div class="container" role="tablist" tabindex="-1">
+      <div 
+      class="container" 
+      role="tablist" 
+      tabindex="-1" 
+      aria-label="${ifDefined(this.dataAriaLabel)}">
```
deveshcisco, line 544: Is there any fallback label in case the consumer misses adding one?

There should be a fallback to prevent any accidental accessibility failures

deveshcisco (commented):
One more note

packages/components/src/components/tablist/tablist.component.ts
cscheffauer: disagree, there should not be any fallback cause a fallback is not translated. If a consumer misses passing the aria-label, its up to them, not on us. If we provide a fallback, you are setting a aria-label to a constant value like "Aria-label" which is very bad for a user. Like we should not provide any aria-label instead of a random, non-meaningful one.

packages/components/src/components/tablist/tablist.component.ts
cscheffauer: aria-labelledby does not work across shadow dom, thus it can't be used (only aria-label).
The reason for this is cause the role is on the div inside - ideally this should be changed in the future to have the tablist role on the host directly. Thats tech debt though

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,320 @@
...
+        await componentsPage.page.keyboard.press('Enter');
+        await componentsPage.visualRegression.takeScreenshot(
+          'mdc-tablist',
+          { source: 'userflow',
+            fileNameSuffix: '4' },
+        );
+        await componentsPage.visualRegression.takeScreenshot(
+          'mdc-tablist',
+          { source: 'userflow',
+            fileNameSuffix: '1' },
```
cscheffauer, line 294: issue: using numbers here is not great, since it doesn't really mean anything - i would vote for a better naming.

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,320 @@
...
+      await componentsPage.setAttributes(mdcTablist, { 'data-aria-label': 'Media types tabs' });
+      await expect(tablist).toHaveAttribute('aria-label', 'Media types tabs');
+    });
+  });
+
+  /**
+   * ACCESSIBILITY
+   */
+  await test.step('accessibility', async () => {
+    await componentsPage.accessibility.checkForA11yViolations('tablist-default');
```
cscheffauer, line 170: issue: this should be run in multiple scenarios, like when the arrow buttons are visible etc etc.

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,320 @@
...
+  tab-id="meetings-tab"
+  aria-controls="meetings-panel">
+  </mdc-tab>`;
+
+const setup = async (args: SetupOptionsType) => {
+  const { componentsPage, ...restArgs } = args;
+  await componentsPage.mount({
+    html: `
+    <div id="mdc-tablist-example">
+      <mdc-tablist
```
cscheffauer, line 59: issue: we have to test the events / onChange callback also

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,345 @@
...
+      await componentsPage.page.keyboard.press('Enter');
+      const activeTab = tabs.first();
+      await expect(activeTab).toHaveAttribute('aria-selected', 'true');
+      await expect(activeTab).toHaveAttribute('active');
+    });
+
+    await test.step('component should fire change event when active tab changes', async () => {
+      await mdcTablist.evaluate((element) => {
+        element.addEventListener('change', () => {
+          element.classList.toggle('onchange');
```
cscheffauer, line 198: issue: this is not necessary - you can return something from the evaluate (just return the eventListener).

cscheffauer (changes requested):
there is one more comment left, other than that looking good

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,345 @@
...
+        });
+      });
+      await expect(mdcTablist).not.toHaveClass('onchange');
+      await componentsPage.actionability.pressTab();
```
deveshcisco, line 202: There is no guarantee that just pressing tab would actually focus a tab element.

Either have a check that makes sure that a tab element is in focus or programatically focus it.

deveshcisco: Instead of setting a class, prefer a test id.

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,345 @@
...
+        await expect(mdcTablist).toHaveAttribute('active-tab-id', 'music-tab');
+        await componentsPage.actionability.pressTab();
+        await expect(arrowButtons.last()).toBeFocused();
+        await componentsPage.page.keyboard.press('Shift+Tab');
+        await expect(tabs.nth(2)).toBeFocused();
+        await componentsPage.page.keyboard.press('Shift+Tab');
+        await expect(arrowButtons.first()).toBeFocused();
+      });
+
+      await test.step(`Given a tablist component with 4 tabs and a smaller viewport,
```
deveshcisco, line 274:
```suggestion
      await test.step(`Given a tablist component with 5 tabs and a smaller viewport,
```

deveshcisco (commented):
Some notes on tests

packages/components/src/components/tablist/tablist.e2e-test.ts
```diff
@@ -0,0 +1,345 @@
...
+      await componentsPage.page.keyboard.press('ArrowRight');
+      await componentsPage.page.keyboard.press('Enter');
+      await expect(tabs.nth(1)).toHaveAttribute('aria-selected', 'true');
+      await expect(tabs.nth(1)).toHaveAttribute('active');
+      await expect(mdcTablist).toHaveAttribute('active-tab-id', 'videos-tab');
+    });
+
+    // resizing viewport only on desktop devices.
+    const deviceName = test.info().project.name;
+    if (['chrome', 'firefox', 'msedge', 'webkit'].includes(deviceName)) {
```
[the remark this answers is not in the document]
cscheffauer: Not a requirement for now, closing

packages/components/src/components/tablist/tablist.e2e-test.ts
deveshcisco: Also, as per the logic implemented in the tablist, tabbing into the tablist automatically focuses the active tab.

We have the code set to not fire a change event when a tab is already active. This test should not work anyway. 
There is an issue with this test.

@supminn @cscheffauer

cscheffauer: @deveshcisco the setup of the tablist happens already in line 187, then there is a tab press and enter happening before reaching this step (read about playwrights "step" function in case its unclear how it works).

@supminn can u please check - i wonder how this can work cause the focus to me is already inside of the tablist once this step is reached, so pressing Tab in line 202 should bring the focus out of the tablist (maybe to the arrow button?) and then enter is pressed. Not sure if this is what we want. If yes, please write the step's description more clearly.

cscheffauer: resolved now, new setup happens at the beginning the step, then pressing tab and arrow right to get the tab next to the interactive one as the newly activated tab.

cscheffauer (approved):

PR #1240, opened 2025-04-23
fix: popover & select style

cscheffauer:
### Description

- Fix styling of popover and select:
  - Fix an issue where the popover does not calculate position properly by setting top / left to 0
  - Select should be not user-selectable, fixed

### DIFF
--- a/packages/components/src/components/popover/popover.styles.ts
+++ b/packages/components/src/components/popover/popover.styles.ts
@@ -11,11 +11,14 @@ const styles = css`
     --mdc-popover-inverted-border-color: var(--mds-color-theme-inverted-outline-primary-normal);
     --mdc-popover-inverted-text-color: var(--mds-color-theme-inverted-text-primary-normal);
     --mdc-popover-elevation-3: var(--mds-elevation-3);
-    --mdc-popover-max-width: unset;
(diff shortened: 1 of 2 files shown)


packages/components/src/components/popover/popover.styles.ts
```diff
@@ -11,11 +11,14 @@ const styles = css`
     --mdc-popover-inverted-border-color: var(--mds-color-theme-inverted-outline-primary-normal);
     --mdc-popover-inverted-text-color: var(--mds-color-theme-inverted-text-primary-normal);
     --mdc-popover-elevation-3: var(--mds-elevation-3);
-    --mdc-popover-max-width: unset;
+    --mdc-popover-max-width: max-content;
```
supminn, line 14: I believe it was set to 'unset' on purpose. Something related to how the options were being rendered within the dropdown for select.. (not sure)
@arjunkesava could you confirm on this?

cscheffauer: this has been clarified, like it will get overridden in the select anyways.

arjunkesava: Yeah, it's been clarified.

arjunkesava (approved):

PR #1249, opened 2025-04-28
perf: build menu item component

arjunkesava:
### Description

- Add menuitem component.

### Screenshot

image

### Links

- Will post soon.


packages/components/src/components/menuitem/menuitem.component.ts
```diff
@@ -0,0 +1,36 @@
+import type { CSSResult, PropertyValues } from 'lit';
+import ListItem from '../listitem/listitem.component';
+
+/**
+ * menuitem component is inherited by listitem component with the role set "menuitem".<br/>
+ * A menu item can contain an icon on the leading or trailing side.
+ *
```
arjunkesava, line 7: PR reviewers, please help me with rewriting this text. I've spent 30 mins here but I was unable to jot it down anything...

packages/components/src/components/menuitem/menuitem.component.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.tertiaryLabel = undefined as unknown as string;
+    this.sublineText = undefined as unknown as string;
+  }
+
+  override updated(changedProperties: PropertyValues): void {
+    super.updated(changedProperties);
+
+    if (changedProperties.has('disabled')) {
+      this.setAttribute('aria-disabled', `${this.disabled}`);
+    }
```
supminn, line 30: question: Are we not doing this on the Listitem component? If not, then why? Shouldn't it be present on listitem too?

packages/components/src/components/menuitem/menuitem.stories.ts
```diff
@@ -0,0 +1,159 @@
...
+    disabled: {
+      control: 'boolean',
+    },
+    'data-aria-label': {
+      control: 'text',
+    },
+    ...hideControls([
+      'variant',
+      'tertiary-label',
+      'subline-text',
```
supminn, line 52: question: Why are we hiding them on stories?
If they are not supported on this component, shouldn't we update the same on `connectedCallback` as well?

packages/components/src/components/menuitem/menuitem.stories.ts
```diff
@@ -0,0 +1,159 @@
...
+      '--mdc-listitem-background-color-active',
+      '--mdc-listitem-background-color-hover',
+      '--mdc-listitem-default-background-color',
+      '--mdc-listitem-primary-label-color',
+      '--mdc-listitem-secondary-label-color',
+      '--mdc-listitem-disabled-color',
+      '--mdc-listitem-column-gap',
+      'default',
+      'keyup',
+      'keydown',
```
supminn, line 63: question: Why is this done? where are these variables set?

arjunkesava: For every new component, these events are automatically generated, I'm just hiding them from the storybook as we don't support those events as of now.

arjunkesava: The above stmt is false, the events are not generated automatically, as this menuitem component is extended from listitem component, the above events are pulled from listitem component, I've reverted my changes via https://github.com/momentum-design/momentum-design/pull/1249/commits/07451121c3fc4a6fdde344808ecbf904ac41f167

supminn (approved):

PR #1256, opened 2025-04-30
fix: update listitem to select only active elements

arjunkesava:
### Description

- When navigating across the listitems in a list, the current `listItems` variable selects all list items even the disable items.
- This PR updates the selector which now selects the item which are not disabled.


2020Deeya (dismissed):
LGTM

packages/components/src/components/list/list.component.ts
```diff
@@ -19,7 +19,11 @@ import { KEYS } from '../../utils/keys';
  * @slot default - This is a default/unnamed slot
  */
 class List extends DataAriaLabelMixin(Component) {
-  @queryAssignedElements({ selector: LISTITEM_TAGNAME })
+  /**
+   * @internal
+   * Get all listitem elements which are not disabled in the list.
+   */
+  @queryAssignedElements({ selector: `${LISTITEM_TAGNAME}:not([disabled])` })
```
cscheffauer, line 26: suggestion: an e2e test covering this would be great

cscheffauer (changes requested):

cscheffauer (approved):

PR #1266, opened 2025-05-06
feat(assets): Device Image Update

tricarro:
Increased the size of device images.


cscheffauer (changes requested):
this PR should not go in - png's are now 1MB big each, not optimised at all. This will cause issues.

PR #1276, opened 2025-05-12
fix: create menusection component

2020Deeya:
### Description

1] Create MenuSection component
2] Add Storybook docs

### Jira Tickets

- MOMENTUM-672


packages/components/src/index.ts
```diff
@@ -130,6 +131,7 @@ export {
   Tooltip,
   MenuItemRadio,
   MenuItemCheckbox,
+  Menusection,
```
arjunkesava, line 134:
```suggestion
  MenuSection,
```

packages/components/src/components/menusection/menusection.stories.ts
```diff
@@ -0,0 +1,61 @@
...
+    ...classArgType,
+    ...styleArgType,
+  },
+};
+
+export default meta;
+
+export const Example: StoryObj = {
+  render,
+};
```
arjunkesava, line 61: As there are no inputs from the user, we can hide all controls.
```suggestion
export const Example: StoryObj = {
  render,
  ...hideAllControls(),
};
```

packages/components/src/components/menusection/menusection.stories.ts
```diff
@@ -0,0 +1,61 @@
...
+  title: 'Work In Progress/menusection',
+  tags: ['autodocs'],
+  component: 'mdc-menusection',
+  parameters: {
+    badges: ['wip'],
+  },
+  argTypes: {
+    ...disableControls([
+      'default',
+    ]),
```
arjunkesava, line 51: I would suggest to hide this, instead of disabling it.

```suggestion
    ...hideControls([
      'default',
    ]),
```

packages/components/src/components/menusection/menusection.stories.ts
```diff
@@ -0,0 +1,61 @@
...
+const wrapWithDiv = (htmlString: TemplateResult) => html`<div role="menubar" style="width: 25rem;">${htmlString}</div>`;
+
+const render = () => wrapWithDiv(html`
+    <mdc-menusection>
+      <mdc-menuitemcheckbox label="Checkbox1" aria-label="Checkbox 1">Checkbox1</mdc-menuitemcheckbox>
+      <mdc-menuitemcheckbox label="Checkbox2" aria-label="Checkbox 2">Checkbox2</mdc-menuitemcheckbox>
+      <mdc-menuitemcheckbox label="Checkbox3" aria-label="Checkbox 3">Checkbox3</mdc-menuitemcheckbox>
+      <mdc-menuitemradio label="Radio1" aria-label="Radio 1">Radio1</mdc-menuitemradio>
+      <mdc-menuitemradio label="Radio2" aria-label="Radio 2">Radio2</mdc-menuitemradio>
+      <mdc-menuitemradio label="Radio3" aria-label="Radio 3">Radio3</mdc-menuitemradio>
```
arjunkesava, line 21:
```suggestion
      <mdc-menuitemcheckbox label="Checkbox1" aria-label="Checkbox 1"></mdc-menuitemcheckbox>
      <mdc-menuitemcheckbox label="Checkbox2" aria-label="Checkbox 2"></mdc-menuitemcheckbox>
      <mdc-menuitemcheckbox label="Checkbox3" aria-label="Checkbox 3"></mdc-menuitemcheckbox>
      <mdc-menuitemradio label="Radio1" aria-label="Radio 1"></mdc-menuitemradio>
      <mdc-menuitemradio label="Radio2" aria-label="Radio 2"></mdc-menuitemradio>
      <mdc-menuitemradio label="Radio3" aria-label="Radio 3"></mdc-menuitemradio>
```

menu items / listitems won't take text from the slot, so this will show a bad example for the consumer.
Also, you can remove aria label also, as the label and aria label have both text, the screen reader will read it 2 times.

packages/components/src/components/menusection/menusection.stories.ts
```diff
@@ -0,0 +1,61 @@
+import type { Meta, StoryObj } from '@storybook/web-components';
+import type { TemplateResult } from 'lit';
+import { html } from 'lit';
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
+import { disableControls } from '../../../config/storybook/utils';
+import '.';
+import '../menuitem';
+import '../menuitemcheckbox';
+import '../menuitemradio';
+import '../divider';
```
arjunkesava, line 10: Please organize these imports.

packages/components/src/components/menusection/menusection.constants.ts
```diff
@@ -0,0 +1,14 @@
...
+const TAG_NAME = utils.constructTagName('menusection');
+
+const ARIA_CHECKED_STATES = {
+  TRUE: 'true',
+  FALSE: 'false',
+} as const;
+
+const DEFAULTS = {
+  ROLE: 'group',
+} as const;
```
arjunkesava, line 12: Please use role from this file: `packages/components/src/utils/roles.ts`
Don't create a new role here.

packages/components/src/components/menusection/menusection.component.ts
```diff
@@ -0,0 +1,94 @@
...
+   * @internal
+   */
+  @queryAssignedElements({ selector: `${MENUITEMRADIO_TAGNAME}` })
+  radios!: MenuItemRadio[];
+
+  override connectedCallback(): void {
+    super.connectedCallback();
+    this.setAttribute('role', DEFAULTS.ROLE);
+    this.addEventListener('click', this.handleChildEvent);
+    this.addEventListener('keydown', this.handleChildEvent);
```
arjunkesava, line 36: Writing generic functions and calling them with separate names will yield a long life.
```
    this.addEventListener('click', this.handleMouseClickEvent);
    this.addEventListener('keydown', this.handleKeyDownEvent);

    private handleMouseClickEvent(event: MouseEvent): void {
        this.handleMenuItemState(event.target);
    }

    private handleKeyDownEvent(event: KeyboardEvent): void {
        if (event.key === KEYS.SPACE) {
            event.preventDefault();
        }
        // check for keys here.
        if (event.key === KEYS.ENTER || event.key === KEYS.SPACE) {
            this.handleMenuItemState(event.target);
        }
    }

    private handleMenuItemState(target: EventTarget | null) {
        // code to check and update the state of menuitem
    }
```

By writing like this, the code will become easy to understand,
you don't need to check weather the call is made from keyboard or mouse.

packages/components/src/components/menusection/menusection.component.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @slot - Default slot for inserting `menuitem`, `menuitemcheckbox`, or `menuitemradio`
+ */
+class MenuSection extends Component {
+  /**
+   * Query assigned `menuitemradio` elements from the default slot.
+   * These elements are used to enforce single-selection behavior.
+   *
+   * @internal
+   */
+  @queryAssignedElements({ selector: `${MENUITEMRADIO_TAGNAME}` })
```
arjunkesava, line 29:
```suggestion
  @queryAssignedElements({ selector: MENUITEMRADIO_TAGNAME })
```

packages/components/src/components/menuitemcheckbox/menuitemcheckbox.component.ts
```diff
@@ -5,9 +5,11 @@ import { ROLE } from '../../utils/roles';
 import MenuItem from '../menuitem/menuitem.component';
 import { TYPE } from '../text/text.constants';
 import { TOGGLE_SIZE } from '../toggle/toggle.constants';
-import { ARIA_CHECKED_STATES, DEFAULTS, INDICATOR } from './menuitemcheckbox.constants';
+import { DEFAULTS, INDICATOR } from './menuitemcheckbox.constants';
+import type { Indicator } from './menuitemcheckbox.types';
+import { ARIA_CHECKED_STATES } from '../menusection/menusection.constants';
+import type { AriaCheckedStates } from '../menusection/menusection.types';
```
arjunkesava, line 11: Please organize these imports.

packages/components/src/components/menuitemcheckbox/menuitemcheckbox.stories.ts
```diff
@@ -6,7 +6,8 @@ import '../icon';
 import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
 import { disableControls, hideAllControls, hideControls } from '../../../config/storybook/utils';
 import { POPOVER_PLACEMENT } from '../popover/popover.constants';
-import { ARIA_CHECKED_STATES, INDICATOR } from './menuitemcheckbox.constants';
+import { INDICATOR } from './menuitemcheckbox.constants';
+import { ARIA_CHECKED_STATES } from '../menusection/menusection.constants';
```
arjunkesava, line 10: Please organize these imports.

packages/components/src/components/menusection/menusection.component.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    if (!isActivationEvent) return;
+
+    if (isKeyboardEvent && event.key === ' ') {
+      event.preventDefault(); // Prevent page scroll
+    }
+
+    const target = event.target as HTMLElement | null;
+    if (!target) return;
+    const { nodeName } = target;
```
arjunkesava, line 65: Please use tagName instead of node name.

packages/components/src/components/menusection/menusection.component.ts
```diff
@@ -0,0 +1,94 @@
...
+    const currentChecked = target.getAttribute('aria-checked') === ARIA_CHECKED_STATES.TRUE;
+
+    switch (nodeName) {
+      case MENUITEMCHECKBOX_TAGNAME.toUpperCase():
+        target.setAttribute('aria-checked', String(!currentChecked));
+        break;
+
+      case MENUITEMRADIO_TAGNAME.toUpperCase():
```
arjunkesava, line 73: Instead of converting it to uppercase, it would be better to convert the tagname to lowercase.
```suggestion
    switch (tagName?.toLowerCase()) {
      case MENUITEMCHECKBOX_TAGNAME:
        target.setAttribute('aria-checked', String(!currentChecked));
        break;

      case MENUITEMRADIO_TAGNAME:
```

arjunkesava (changes requested):

packages/components/src/components/menusection/menusection.stories.ts
```diff
@@ -0,0 +1,61 @@
...
+    </mdc-menusection>
+    <mdc-divider></mdc-divider>
+    <mdc-menusection>
+      <mdc-menuitemcheckbox label="Toggle6" indicator="toggle" aria-label="Checkbox 6">Checkbox6</mdc-menuitemcheckbox>
+      <mdc-menuitemcheckbox label="Checkmark7" indicator="checkmark" 
+      aria-label="Checkbox 7">Checkbox7</mdc-menuitemcheckbox>
+    </mdc-menusection>`);
+
+const meta: Meta = {
+  title: 'Work In Progress/menusection',
```
arjunkesava, line 42: Please place this inside `menu` folder.
```suggestion
  title: 'Work In Progress/menu/menusection',
```

packages/components/src/components/menusection/menusection.stories.ts
2020Deeya: I don't think it's needed, since there are no controls present at the moment. Also, I noticed that adding it introduces a new "parameters" control. Do we really require this?

image

packages/components/src/components/menusection/menusection.stories.ts
2020Deeya: I see that MenuSection has a default slot, which could be a MenuItem, MenuItemRadio, or MenuItemCheckbox. Keeping it disabled as a placeholder could help convey this to the consumer. I'd like to know your thoughts on this.

packages/components/src/components/menusection/menusection.stories.ts
2020Deeya: It seems to be organized using the command palette.

packages/components/src/components/menusection/menusection.component.ts
```diff
@@ -0,0 +1,117 @@
...
+   * Activates menu items when the Enter or Space keys are pressed.
+   *
+   * @param event - The keyboard event from a child menu item
+   *
+   * @internal
+   */
+  private handleKeydown = (event: KeyboardEvent) => {
+    if (event.key !== 'Enter' && event.key !== ' ') return;
+
+    if (event.key === ' ') {
```
arjunkesava, line 69:
```suggestion
    if (event.key !== KEYS.ENTER && event.key !== KEYS.SPACE) return;

    if (event.key === KEYS.SPACE) {
```

packages/components/src/components/menusection/menusection.stories.ts
arjunkesava: As the default slot doesnot have any name, it would be better to hide it. Consumer can add a slot with the name default, if we show this on story.

packages/components/src/components/menusection/menusection.component.ts
arjunkesava:
```suggestion
  @queryAssignedElements({ selector: `${MENUITEMRADIO_TAGNAME}:not([disabled])` })
```

If the menu item is disabled, we should not set it to true or false.

arjunkesava (changes requested):
Please organize imports, and add disabled check for the radios property.

packages/components/src/components/menusection/menusection.component.ts
```diff
@@ -0,0 +1,117 @@
...
+  private toggleCheckedState(target: EventTarget | null) : void {
+    if (!(target instanceof HTMLElement)) return;
+
+    const tagName = target.tagName.toLowerCase();
+    const currentChecked = target.getAttribute('aria-checked') === ARIA_CHECKED_STATES.TRUE;
+
+    switch (tagName) {
+      case MENUITEMCHECKBOX_TAGNAME:
+        target.setAttribute('aria-checked', String(!currentChecked));
+        break;
```
arjunkesava, line 94: If the `menuitemcheckbox` is disabled, then aria-checked should not work, but it's working. Please add that condition.

packages/components/src/components/menuitemcheckbox/menuitemcheckbox.component.ts
2020Deeya: Sure. 
Are we following it everywhere? I don't see it mentioned in any of our conventions.

2020Deeya:
> Please organize imports, and add disabled check for the radios property.
Sure, Let me check.
Question: Are we following it everywhere? I don't see it mentioned in any of our conventions. If it's really needed, we should definitely add it somewhere, since not everyone has the same local settings.

arjunkesava (approved):

PR #1282, opened 2025-05-13
fix: progressbar margin styles

supminn:
### Description

Fixing styles for progressbar component

TICKET_NO: MOMENTUM-671


packages/components/src/components/progressbar/progressbar.styles.ts
```diff
@@ -16,13 +17,17 @@ const styles = [css`
   --mdc-progressbar-label-font-weight: var(--mds-font-weight-regular);
   --mdc-progressbar-help-text-color: var(--mds-color-theme-text-secondary-normal);
 
-  display: block;
   width: 100%;
 }
 
+:host([variant="inline"]) {
+ display: block;
```
cscheffauer, line 24: can this be done without changing the "display" on different variants? can we use the same display in all variants?
i am not sure if this imposes an issue for consumers - plus for extensibility it will be better to stick to one display and build all variants with that

cscheffauer (approved):

PR #1284, opened 2025-05-13
fix: removed scroll into view

supminn:
### Description

Removed the logic of having "scrollIntoView" which causes issues.

MOMENTUM-669


packages/components/src/components/tablist/tablist.component.ts
```diff
@@ -154,16 +154,6 @@ class TabList extends Component {
...
     const resizeObserver = new ResizeObserver(async () => {
-      const { activeElement } = document;
-
-      /**
-       * Keep the focused element in view.
-       */
-      if (activeElement instanceof Tab) {
-        // @ts-ignore : https://github.com/Microsoft/TypeScript/issues/28755
-        activeElement.scrollIntoView({ behavior: 'instant', block: 'start', inline: 'center' });
-      }
```
cscheffauer, line 165: q: can it be removed without causing a reduction in terms of the feature set? like what exactly is now not working anymore cause of this?

cscheffauer (approved):

PR #1289, opened 2025-05-13
fix: select icon option

supminn:
### Description

Updated select component to show icon of the selected value.
Added a test case on visual-regression.
Updated all the snapshots for select component.

MOMENTUM-654


packages/components/src/components/select/select.component.ts
```diff
@@ -520,6 +526,9 @@ class Select extends FormInternalsMixin(DataAriaLabelMixin(FormfieldWrapper)) im
           aria-labelledby="${this.label ? FORMFIELD_DEFAULTS.HEADING_ID : ''}"
           aria-expanded="${this.displayPopover ? 'true' : 'false'}"
         >
+      ${this.selectedIcon
```
cscheffauer, line 529: issue: this change has to be reflected in a e2e test (visual regression)

cscheffauer (changes requested):

cscheffauer (approved):

PR #1290, opened 2025-05-13
feat: introduced close event in dialog component

balaram36:
Added close event for close icon in dialog component.

### Description

Close event will fire if we click on "close icon"


packages/components/src/components/dialog/dialog.component.ts
```diff
@@ -273,6 +273,7 @@ class Dialog extends FocusTrapMixin(CardAndDialogFooterMixin(Component)) {
    */
   public hideDialog = () => {
     this.visible = false;
+    DialogEventManager.onCloseDialog(this);
```
cscheffauer, line 276: issue: we should not add a new event, but try to understand why "hidden" is not firing. Line 266 should be fired if the visible property gets false.
Please investigate and fix that instead of adding a new event.

cscheffauer (changes requested):

packages/components/src/components/dialog/dialog.component.ts
balaram36: Okay, I will re-check

balaram36: Removed new event and using existing event for close

PR #1293, opened 2025-05-13
feat(components): animation

maxinteger:
### Description

Add new `<mdc-animation />` component


packages/components/src/components/animation/animation.component.ts
```diff
@@ -0,0 +1,163 @@
...
+import lottie, { AnimationItem } from 'lottie-web/build/player/lottie_light';
+import * as animationManifest from '@momentum-design/animations/dist/manifest.json';
+import styles from './animation.styles';
+import { Component } from '../../models';
+import { AnimationNames, LoopType } from './animation.types';
+import { DEFAULTS } from './animation.constants';
+
+/**
+ * The `mdc-animation` component is a wrapper around the Lottie animation library.
+ * It fetches the animation data dynamically based on the provided name and renders it.
```
cscheffauer, line 14: suggestion: i think it would be good to mention here that it displays animations only from the "@momentum-design/animations" library

packages/components/src/components/animation/animation.styles.ts
```diff
@@ -0,0 +1,12 @@
+import { css } from 'lit';
+import { hostFitContentStyles } from '../../utils/styles';
+
+const styles = [
+  hostFitContentStyles,
+  css`
```
cscheffauer, line 6: suggestion: these 4 lines (6-9) can be removed, since not necessary. If we need to apply styling later, someone could add it again.

cscheffauer (approved):

cscheffauer (approved):

PR #1301, opened 2025-05-19
chore: creating avatar button screenshots

supminn:
### Description

Adding visual regression snapshots for avatarButton

MOMENTUM-651


packages/components/src/components/avatarbutton/avatarbutton.e2e-test.ts
```diff
@@ -161,4 +158,66 @@ test('mdc-avatarbutton', async ({ componentsPage }) => {
...
+      });
+    });
+    await avatarButtonStickerSheet.mountStickerSheet();
+    const container = avatarButtonStickerSheet.getWrapperContainer();
+    const avatars = await container.locator('mdc-avatar[src]').all();
+    for (const avatarComp of avatars) {
+      const image = avatarComp.locator('img');
+      await image.waitFor();
+      await expect(avatarComp).toHaveAttribute('src', src);
+      await expect(image).toHaveAttribute('src', src);
```
2020Deeya, line 209: You’re testing for the presence of img[src], but not for the alt attribute. Shouldn't that be included as well?

supminn: not required here.
We are testing for the img src to confirm that the image has been loaded.
We are waiting for the image to be loaded first to capture the snapshot correctly.

2020Deeya (approved):

PR #1306, opened 2025-05-21
refactor: adding a custom header slot for dialog component

supminn:
### Description

* Updated dialog component to support 'dialog-header' slot
* Added a new storybook example for this change
* Updated visual regression snapshot with custom header

MOMENTUM-676


arjunkesava (changes requested):
Please rebase this PR with latest upstream or cherry and pick only your commits for dialog related PR.

packages/components/src/components/dialog/dialog.component.ts
```diff
@@ -288,6 +289,7 @@ class Dialog extends FocusTrapMixin(CardAndDialogFooterMixin(Component)) {
 
   public override render() {
     return html`
+    <slot name="dialog-header"></slot>
```
cscheffauer, line 292: suggestion: this slot name is misleading, it indicates that it replaces the header. But if a consumer passes in the header text and uses this slot, it will not replace it. I would vote for a different slot name here i.e. `dialog-header-prefix` or similar

supminn: oh, I have understood this requirement incorrectly.
I was of the understanding that we'd need to provide a slot for replacing the default header attributes that we pass in.
I'll connect with you to understand this correctly.

supminn:
closing this PR
I have incorrectly created the branch against menu

PR #1313, opened 2025-05-22
perf: add e2e specs for menuitem component

arjunkesava:
### Description

- PR to add e2e specs for menu item component.
- As there are no interactive elements in this component, the "Interactions" check is ignored.

### Links

- MOMENTUM-660


supminn:
Could we get an interaction testcase for menuitem or listitem to check if the tooltip is shown when "tooltip-text" is passed?

PR #1314, opened 2025-05-22
refactor: dialog component

supminn:
### Description

* Updated the dialog component to support "header-prefix" slot which is placed before the header text. The component styling can be updated using "header" part.
* Added a visual regression test case and story for this change
* Updated the dialog component to fire 'onshown' event when the consumer uses it without a triggerId.

MOMENTUM-676
MOMENTUM-681


packages/components/src/components/dialog/dialog.e2e-test.ts
```diff
@@ -107,7 +126,7 @@ test('mdc-dialog', async ({ componentsPage }) => {
       await componentsPage.visualRegression.takeScreenshot('mdc-dialog', { element: dialog });
```
cscheffauer, line 126: suggestion: we also want to create a e2e test which opens the dialog without a trigger, just with using visible. We are supporting it as a use case, so we have to test it as well.

cscheffauer (changes requested):

packages/components/src/components/dialog/dialog.e2e-test.ts
supminn: We have covered this already in the interactions. I have tweaked it to remove the triggerid.

cscheffauer (approved):

PR #1327, opened 2025-05-29
refactor: chip orange color

supminn:
### Description

Updating chip component to support 'orange' color.

MOMENTUM-695


packages/components/src/components/chip/chip.component.ts
```diff
@@ -38,6 +38,7 @@ class Chip extends IconNameMixin(Buttonsimple) {
    * - mint
    * - pink
    * - purple
+   * - orange
```
arjunkesava, line 41: - The `orange` should come after the `mint` as a sorted way.

packages/components/src/components/chip/chip.constants.ts
```diff
@@ -12,6 +12,7 @@ const COLOR = {
   MINT: 'mint',
   PINK: 'pink',
   PURPLE: 'purple',
+  ORANGE: 'orange',
```
arjunkesava, line 15: Same here, `ORANGE` should come after `MINT`.

arjunkesava (changes requested):
Minor Comment, rest all Good.

PR #1337, opened 2025-06-06
refactor: logic on Select and Popover component

supminn:
### Description

* Fixed the logic of setting active-descendant on select component
* Adding aria-controls on combobox when listbox is open
* Fixed popover component's logic for hoverBridge
* Updated aria-modal on popover to be applied only when role is dialog or alertdialog
* Placing mdc-option, mdc-optgroup and mdc-select inside the select directory on storybook.
* User stories for select component - select.feature file
* Replaced all hard-coded roles to use the constant ROLE from utils

**Note**: This PR does not fix the SR focus issue (MOMENTUM-702) on select component.


packages/components/src/components/select/select.component.ts
```diff
@@ -372,25 +355,47 @@ class Select extends FormInternalsMixin(DataAriaLabelMixin(FormfieldWrapper)) im
...
     this.displayPopover = true;
-    this.resetActivedescendant();
+
+    // Find the currently selected option or the first option
+    const options = this.getAllValidOptions();
+    const selectedOption = options.find((option) => option.hasAttribute('selected'));
+    const focusedOption = options.find((option) => option.getAttribute('tabindex') === '0');
+
+    // Set activeDescendant to the selected/focused option or first option
+    this.activeDescendant = (selectedOption || focusedOption || options[0])?.id ?? '';
```
arjunkesava, line 399: The problem with this approach is that if an option is selected, we still check for a `focusedOption`.
For a select with a very limited list of options, this won't make any difference, but for a select with 50K options this will cost a delay in performance.

Please check the `selectedOption` first, and if nothing is selected (i.e. selectedOption = undefined),
then look for `focusedOption` and then if no focusedOption is present, then we set the default to first option (i.e. options[0]).

arjunkesava (changes requested):

arjunkesava (approved):

PR #1338, opened 2025-06-09
docs: add user stories for the MenuPopover component

2020Deeya:
### Description

Add user stories for the MenuPopover component to improve test coverage and clarify expected behaviors. These user stories cover key interaction scenarios and accessibility considerations for each component.

### DIFF
--- a/packages/components/src/components/menupopover/menupopover.feature
+++ b/packages/components/src/components/menupopover/menupopover.feature
@@ -141,7 +141,7 @@ Feature: MenuPopover Accessibility and User Interaction
       When I navigate with arrow keys
       Then focus should move sequentially through all menuitems, regardless of groups
 
-  Rule: ✅ Nested Submenus
+  Rule: ✅ Nested Submenu Interaction
 
     Background:
       Given the MenuPopover contains a nested submenu under "Settings"
@@ -153,8 +153,6 @@ Feature: MenuPopover Accessibility and User Interaction
         | Advanced | menuitem | disabled |
       And the nested submenu is closed by default
       And the MenuPopover is open
-
-  Rule: ✅ Nested Submenu Interaction
 
     Scenario: Open nested submenu with mouse
       Given the focus is on Settings menuitem


packages/components/src/components/menupopover/menupopover.mdx
```diff
@@ -0,0 +1,263 @@
...
+    **Scenario: Separators are skipped by the screen reader**
+    > Given the menupopover contains section dividers (separators)  
+    > When I navigate using screen reader   
+    > Then the screen reader skips directly to the next content item afterwards
+
+    **Scenario: Group headings are announced for context**
+    > Given the menupopover contains groups of menu items with headings  
+    > When I navigate into a group using screen reader   
+    > Then the screen reader announces the group heading label  
+    > And then continues announcing the contained menu items
```
arjunkesava, line 261: SR's won't announce the menu items directly, the user should navigate on each item.

2020Deeya (commented):
LGTM

supminn (approved):

maxinteger (changes requested):
I think all the cases are covered, but I found some format related problems.

In general, 
- there is not background specification for the rules (except the first one)
- when an "or" appear in any clause then it should be `Scenario Outline: ` with `Example` table, See my comment on "Scenario: Open nested submenu using keyboard"

maxinteger:
@2020Deeya VO cases are really nice additions to the test cases, but it is also very specific to one technology. do we plane to add other screen readers or drop VO and just test accessibility values?

2020Deeya:
> @2020Deeya VO cases are really nice additions to the test cases, but it is also very specific to one technology. do we plane to add other screen readers or drop VO and just test accessibility values?

Maybe later, once we integrate automation testing more tightly with user stories, we can look into adding broader screen reader coverage. For now, VO gives us a reasonable starting point and help catch some behavioral issues.

packages/components/src/components/menupopover/menupopover.feature
```diff
@@ -200,50 +252,27 @@ Feature: MenuPopover Accessibility and User Interaction
...
       Given VoiceOver interaction is active inside the MenuPopover
-      When VoiceOver navigates right (VO + Right) to "Profile"
-      Then VoiceOver announces: "Profile, menu item"
-      When VoiceOver navigates right (VO + Right) to "Settings"
-      Then VoiceOver announces: "Settings, menu item"
-      When VoiceOver navigates right (VO + Right) to "Notifications"
-      Then VoiceOver announces: "Notifications, 3 new, menu item"
-      When VoiceOver navigates right (VO + Right) to "Logout"
-      Then VoiceOver announces: "Logout, menu item, dimmed"
+      When VoiceOver navigates to the "<menuitem>" menuitem
```
2020Deeya, line 267: I understand the Scenario Outline format. Should we include the command (VO + Right) here for reference?

maxinteger:
@2020Deeya Not sure, but looks like each rules has different menu setup, but only 2 of them has backgrounds. Is it Ok?

packages/components/src/components/menupopover/menupopover.feature
```diff
@@ -0,0 +1,306 @@
...
+      And it includes visual separators
+      When I navigate with arrow keys
+      Then focus should skip over the separator
+
+    Scenario: Moves focus correctly across groups
+      Given the MenuPopover includes grouped menuitems
+      When I navigate with arrow keys
+      Then focus should move sequentially through all menuitems, regardless of groups
+
+  Rule: ✅ Nested Submenus
```
2020Deeya, line 144: @supminn why we have a rule that only contains a background without any scenarios? 
Background alone does nothing without Scenarios.

maxinteger (approved):
It looks good now, thanks for the fixes

packages/components/src/components/menupopover/menupopover.feature
supminn: Yes, you could add this for reference.

packages/components/src/components/menupopover/menupopover.feature
supminn: We have scenarios below the background.

Rule -> background -> scenarios.  That's the order I have followed

PR #1339, opened 2025-06-09
fix: dialog, focustrap and popover fixes

cscheffauer:
### Description

- Fixing issues in dialog, focustrap and popover


packages/components/src/utils/mixins/FocusTrapMixin.ts
```diff
@@ -258,15 +301,15 @@ export const FocusTrapMixin = <T extends Constructor<HTMLElement>>(superClass: T
...
-      const wrapFocus = this.shouldWrapFocus();
 
       if (currentIndex === -1) {
         return step > 0 ? 0 : length - 1;
       }
 
       let nextIndex = currentIndex + step;
 
-      if (wrapFocus) {
+      console.log(this.shouldFocusTrapWrap, nextIndex, length);
```
arjunkesava, line 311: I guess this must be removed.

cscheffauer: thanks! i missed that, pushed the change

PR #1343, opened 2025-06-10
chore: add user stories for popover component

arjunkesava:
### Description

- Add user stories for popover component based on "Given When Then" syntax structure.


packages/components/src/components/popover/popover.feature
```diff
@@ -0,0 +1,159 @@
...
+      Then the trigger should have `aria-haspopup` set to `dialog` (by default),
+      And the user can override `aria-haspopup` if needed.
+      When the popover is open,
+      Then the trigger should have `aria-expanded="true"`.
+      When the popover is closed,
+      Then the trigger should have `aria-expanded="false"`.
+      Then the `disable-aria-expanded` attribute should be used to remove `aria-expanded`.
+      When the popover has interactive attribute set to `true`
+      And has a role of `dialog` or `alertdialog`,
+      Then the `aria-modal` attribute should be set to `true`.
```
supminn, line 109:
```suggestion
    Scenario: Default accessibility attributes
      Given a popover component,
      Then it should have a default role of `dialog`.
      And the `aria-modal` attribute should be set to `true`
      And it can support `aria-label`, `aria-labelledby`, and `aria-describedby`.
       When the popover is open,
      Then the trigger should have `aria-expanded="true"`.
      When the popover is closed,
      Then the trigger should have `aria-expanded="false"`.

Scenario: Accessibility on interactive popover
     Given a popover component
     When the popover has `interactive` attribute set to `true`,
     Then the trigger should have `aria-haspopup` set to `dialog` (by default),
      And the user can override `aria-haspopup` if needed.
 
Scenario: Popover with disabled aria expanded
     Given a popover component
     When the `disable-aria-expanded` attribute is set to `true`
     Then the `aria-expanded` attribute should not be removed.
```

Its easier to understand when they are split into smaller scenarios.

packages/components/src/components/popover/popover.feature
```diff
@@ -0,0 +1,159 @@
...
+
+  Rule: Accessibility with screen readers
+
+    Scenario: Announcing popover content
+      Given a popover component,
+      When the popover trigger element is clicked,
+      Then the popover should open,
+      And screen readers should announce the text content inside the popover.
```
supminn, line 117:
```suggestion
    Scenario: Announcing popover content
      Given a popover component is closed,
      When the popover trigger element is focused using screen reader
      Then the screen reader should announce as a button which has an action.
      When the trigger element is clicked,
      Then the popover should open,
      And screen readers should announce the text content inside the popover.
```

packages/components/src/components/popover/popover.feature
```diff
@@ -0,0 +1,159 @@
...
+
+    Scenario: Navigation inside popover content
+      Given a popover component with list items inside it.
+      When the popover trigger element is clicked,
+      Then the popover should open,
+      And screen readers should read the text of first list item.
+      When the user navigates to next item with **Tab** key,
+      Then screen readers should read the text of the second list item.
+      When the user press **Escape** key then popover should get closed.
```
supminn, line 126:
```suggestion
    Scenario: Navigation inside popover content
      Given a popover component with list items inside it.
      When the popover trigger element is activated using screen reader,
      Then the popover should open and screen reader focus moves to the 1st list item.
      And screen readers should read the text of first list item.
      When the user navigates to next item with **Tab** key,
      Then screen readers should read the text of the second list item.
      When the user press **Escape** key then popover should get closed.
```

supminn (changes requested):

2020Deeya (approved):

PR #1359, opened 2025-06-17
refactor: menupopover component and test cases

2020Deeya:
### Description

* Add e2e tests for the Menupopover component based on menupopover.feature file.
* We are following TDD approach. 
* Updated menupopover component according to the user test behavior
* Refactored keyup and keydown behavior on menuitem


arjunkesava (changes requested):

packages/components/src/components/menupopover/menupopover.component.ts
```diff
@@ -27,39 +30,158 @@ import styles from './menupopover.styles';
...
+      .flat()
+      .filter((node) => !!node)
+      .filter((node) => !node.hasAttribute('disabled'));
+  }
+
   override connectedCallback() {
     super.connectedCallback();
     this.role = ROLE.MENU;
 
-    this.focusBackToTrigger = true;
```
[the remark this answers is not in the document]
supminn: Its set below, on line 43/63

packages/components/src/components/menuitemradio/menuitemradio.component.ts
```diff
@@ -35,11 +35,50 @@ class MenuItemRadio extends MenuItem {
...
+      container = container.parentElement;
+    }
+    if (container) {
+      const radios = Array.from(container.querySelectorAll('mdc-menuitemradio')) as MenuItemRadio[];
+      radios.forEach((item) => {
+        const radio = item;
+        if (radio !== this && radio.name === this.name && !radio.disabled) {
+          radio.ariaChecked = ARIA_CHECKED_STATES.FALSE;
+        }
+      });
```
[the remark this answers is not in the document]
supminn: I differ to agree. I feel the current implementation is easier to read as well as more performant.

Here, we have 1 foreach loop which runs the complete logic once over an array.
With multiple `.filter` we are running the loop repeatedly on a smaller set of array.

packages/components/src/components/menupopover/menupopover.utils.ts
```diff
@@ -0,0 +1,21 @@
...
+import { TAG_NAME as MENUPOPOVER_TAGNAME } from './menupopover.constants';
+
+/**
+ * Checks if the given menu item is a valid menu item.
+ * @param menuItem - The menu item to check.
+ * @returns True if the menu item is a valid menu item, false otherwise.
+ */
+const isValidMenuItem = (menuItem: Element | null): boolean =>
+  [MENUITEM_TAGNAME, MENUITEMCHECKBOX_TAGNAME, MENUITEMRADIO_TAGNAME]
+    .includes(menuItem?.tagName?.toLowerCase() as typeof MENUITEM_TAGNAME);
```
maxinteger, line 13: TS error is a bit misleading here:

> TS2345: Argument of type string | undefined is not assignable to parameter of type

Unfortunate `as typeof MENUITEM_TAGNAME` is resolve the TS error, but hide the real problem.

The actual problem is when `menuItem.tagName.toLowerCase` is undefined then you can not call it so this code will fail.

You need to do the "Optional chaining" (`?.`) on the call operator (`()`) as well:

```suggestion
    .includes(menuItem?.tagName?.toLowerCase?.()
```

This will work because if any part of the expression is `undefined` it just return `undefined` and it Ok for the `includes` which will return with `false` in this case

packages/components/src/components/menupopover/menupopover.component.ts
```diff
@@ -27,39 +30,168 @@ import styles from './menupopover.styles';
...
+        if (node.tagName.toLowerCase() === MENUSECTION_TAGNAME) {
+          return Array.from(node.children)
+            .filter((child) => isValidMenuItem(child)) as Array<HTMLElement>;
+        }
+        return isValidMenuItem(node) ? node : [];
+      },
+    )
+      .flat()
+      .filter((node) => !!node)
+      .filter((node) => !node.hasAttribute('disabled'));
```
maxinteger, line 56: chained `filter`s should be combine into one. Unfortunately compilers/minifiers can not merge this pure functions into one and we iterate on a list 2 times for no real reason.

packages/components/src/components/menupopover/menupopover.component.ts
```diff
@@ -27,39 +30,168 @@ import styles from './menupopover.styles';
...
+    }
+  }
+
+  private handleMouseClick(event: MouseEvent): void {
+    const target = event.target as HTMLElement;
+    const role = target.getAttribute('role');
+    if (
+      isValidMenuItem(target)
+      && !isValidPopover(target.nextElementSibling)
+      && role === ROLE.MENUITEM
```
maxinteger, line 120: is this mean, we allow any element as menu item in a menu?

For example `<div role="menuitem">File</div>`

Most of the place we check the tagName why do we do it differently here?

packages/components/src/components/menupopover/menupopover.component.ts
```diff
@@ -27,39 +30,168 @@ import styles from './menupopover.styles';
...
+        return key;
+    }
+  }
+
+  private handleKeyDown(event: KeyboardEvent) {
+    const currentIndex = this.getCurrentIndex(event.target);
+    if (currentIndex === -1) return;
+
+    const isRtl = document.querySelector('html')?.getAttribute('dir') === 'rtl'
+     || window.getComputedStyle(this).direction === 'rtl';
```
maxinteger, line 144: What is the point to test the HTML element as well? 
The consumer can change anywhere the direction of the text in the DOM. For example the HTML is RTL, but the menu (eg. for demo reason) is LTR.

This should be enough
```suggestion
    const isRtl = window.getComputedStyle(this).direction === 'rtl';
```

maxinteger (changes requested):
I found some minor problems but over all it is a really nice work!

packages/components/src/components/menupopover/menupopover.component.ts
supminn: No, `isValidMenuItem` checks if it is either: `mdc-menuitem`, `mdc-menuitemcheckbox` or `mdc-menuitemradio`
However, `role === ROLE.MENUITEM` is added to restrict the closing behavior only for `mdc-menuitem`

We could make use of the same tagname instead of the role here. I'll update this, thanks

maxinteger: I see, I did not notice the `isValidMenuItem`

maxinteger (dismissed):

maxinteger (approved):

PR #1364, opened 2025-06-20
chore(components): update repo hygiene

maxinteger:
### Description

- add husky to manage git pre commit hook
- add lint-stage to run scripts on staged code
- add eslint-config-prettier to hand over formatting related rule to prettier
- fix eslintrc
- updated prettier config
- re-run prettier on all files


packages/components/package.json
```diff
@@ -126,9 +128,12 @@
     "concurrently": "^8.2.2",
     "esbuild": "^0.23.0",
     "eslint": "8.27.0",
+    "eslint-config-prettier": "^10.1.5",
     "eslint-plugin-import": "^2.26.0",
     "eslint-plugin-lit": "^1.14.0",
     "fs-extra": "^11.2.0",
+    "husky": "^9.1.7",
```
cscheffauer, line 135: I would like to review this and get this investigated first - husky has caused issues before in Mrv2, so would like to get this evaluated before we make those change

cscheffauer (changes requested):

arjunkesava:
Hey @maxinteger, with the help of lint-staged, can we add VS Code's organize imports cli as well?

- There is a package which we need to install as dev dependency: https://www.npmjs.com/package/@havenmoney/organize-imports-cli
- In the `lint-staged.config.js` file, add this below line
```
'*': ['organize-imports-cli', 'eslint --fix', 'prettier --write'],
```

packages/components/package.json
maxinteger: I removed this version since it is already in the root of the project.

Could you please tell me what were the issues in MRv2?

BTW, MRv2 uses 7.x, while we are using 9.x here

cscheffauer (approved):

PR #1367, opened 2025-06-23
fix: toggletip test and updated snapshot

supminn:
### Description

* Modified the logic of `hidePopover`
* Updated snapshots of toggletip - We can open only 1 toggletip at a time
* Updated toggletip test cases
* Modified toggletip component to handle placement change.
* Modified menupopover component to update default placement


arjunkesava (approved):
LGTM.

PR #1372, opened 2025-06-26
feat(menu): menu component improvements

maxinteger:
### Description

- add action event to menu popover
- add onAction and onChange events to react menu popover
- change aria-checked to checked in menu radio and checkbox
- add checkmark indicator to menu radio
- add label to menu section
- add onChange event to react menu section

### DIFF
--- a/packages/components/src/components/menuitemcheckbox/menuitemcheckbox.component.ts
+++ b/packages/components/src/components/menuitemcheckbox/menuitemcheckbox.component.ts
@@ -1,4 +1,4 @@
-import type { CSSResult, TemplateResult } from 'lit';
+import type { PropertyValues, CSSResult, TemplateResult } from 'lit';
 import { html, nothing } from 'lit';
 import { property } from 'lit/decorators.js';
 import { ROLE } from '../../utils/roles';
@@ -8,14 +8,13 @@ import { TOGGLE_SIZE } from '../toggle/toggle.constants';
 import { DEFAULTS, INDICATOR } from './menuitemcheckbox.constants';
 import type { Indicator } from './menuitemcheckbox.types';
 import { ARIA_CHECKED_STATES } from '../menusection/menusection.constants';
-import type { AriaCheckedStates } from '../menusection/menusection.types';
 import styles from './menuitemcheckbox.styles';
 
 /**
  * A menuitemcheckbox component is a checkable menuitem.
  * There should be no focusable descendants inside this menuitemcheckbox component.
  *
- * The `aria-checked` attribute indicates whether the menuitemcheckbox is checked or not.
+ * The `checked` attribute indicates whether the menuitemcheckbox is checked or not.
  *
  * The `indicator` attribute is used to differentiate between <b>checkbox</b>, <b>checkmark</b> and <b>toggle</b>.
  * By default the `indicator` is set to <b>checkbox</b>.<br/>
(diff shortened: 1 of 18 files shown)


packages/components/src/components/menusection/menusection.component.ts
```diff
@@ -14,22 +18,51 @@ import { ROLE } from '../../utils/roles';
...
+      // more details: https://nolanlawson.com/2022/11/28/shadow-dom-and-accessibility-the-trouble-with-aria/
+      this.ariaLabel = this.label;
+    }
+  }
+
+  private renderLabel() {
+    if (this.label) {
+      return html`<mdc-text
+        class="mdc-menusection__label"
+        role="presentation"
```
cscheffauer, line 51: why was role="presentation" set here?

packages/components/src/components/menusection/menusection.component.ts
```diff
@@ -14,22 +18,51 @@ import { ROLE } from '../../utils/roles';
...
+      // Since the later one just released in the major browsers, we do the first one for now.
+      // more details: https://nolanlawson.com/2022/11/28/shadow-dom-and-accessibility-the-trouble-with-aria/
+      this.ariaLabel = this.label;
+    }
+  }
+
+  private renderLabel() {
+    if (this.label) {
+      return html`<mdc-text
+        class="mdc-menusection__label"
```
cscheffauer, line 50: suggestion: instead of using class, use part. This is mentioned in our styling convention.
setting class here will stop consumers from overridding the style

packages/components/src/components/menuitemradio/menuitemradio.component.ts
```diff
@@ -52,43 +63,103 @@ class MenuItemRadio extends MenuItem {
...
+   */
+  private radioHandleClick = (event: Event) => {
+    event.stopPropagation();
+
+    if (this.disabled || this.checked) return;
+
+    this.updateOtherRadiosCheckedState();
+    this.checked = true;
+
+    this.dispatchEvent(new Event('change', { bubbles: true, composed: true }));
```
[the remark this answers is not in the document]
maxinteger: 1. It is correct the `menupopover` does not handle it, it is needed for the consumer. In the other hand, MdcMenuPopover get the `onChange` prop to capture these events.
2. Yes you can access to the `name` and `value` via the `target` member of the event

I used `Event` because all native `<input />` elements do the same

packages/components/src/components/menusection/menusection.component.ts
```diff
@@ -14,22 +18,51 @@ import { ROLE } from '../../utils/roles';
...
  * @slot - Default slot for inserting `menuitem`, `menuitemcheckbox`, or `menuitemradio`
+ *
+ * @event change - (React: onChange) This event is dispatched when a `menuitemcheckbox`, or `menuitemradio` changes.
  */
 class MenuSection extends Component {
+  /**
+   * The primary label of the list item.
+   * This appears on the leading side of the list item.
+   */
+  @property({ type: String, reflect: true }) label: string | null = null;
```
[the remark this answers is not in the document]
maxinteger: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes) only 3 native element using it <optgroup>, <option>, <track>. And we are already using same common attributes like `name` and `value`.

Still want me to change it?

packages/components/src/components/menusection/menusection.component.ts
```diff
@@ -15,22 +19,46 @@ import { ROLE } from '../../utils/roles';
...
-  override disconnectedCallback(): void {
-    super.disconnectedCallback();
+  override update(changedProperties: PropertyValues): void {
+    super.update(changedProperties);
+    if (changedProperties.has('headerText')) {
+      // Because IDREF attribute reflection does not work across light and shadow DOM, we either set the
+      // `aria-label` directly or use the `ariaLabelledByElements`.
+      // Since the later one just released in the major browsers, we do the first one for now.
+      // more details: https://nolanlawson.com/2022/11/28/shadow-dom-and-accessibility-the-trouble-with-aria/
+      this.ariaLabel = this.headerText;
```
supminn, line 44: Should we make use of `dataAriaLabel` here instead of hard-coding the headertext?

packages/components/src/components/menuitemcheckbox/menuitemcheckbox.component.ts
```diff
@@ -60,6 +67,7 @@ class MenuItemCheckbox extends MenuItem {
   constructor() {
     super();
     this.addEventListener('click', this.menuitemcheckboxHandleClick);
+    // this.ariaChecked = this.checked ? ARIA_CHECKED_STATES.TRUE : ARIA_CHECKED_STATES.FALSE;
```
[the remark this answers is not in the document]
supminn: my bad. I think i missed to remove this and commented instead

packages/components/src/components/menusection/menusection.component.ts
maxinteger: not with `dataAriaLabel`, but fixed

arjunkesava (approved):
LGTM.

cscheffauer (approved):

PR #1373, opened 2025-06-26
fix: dialog focustrap expose

cscheffauer:
### Description

- Allow focus trap to be set for Dialog. It should be avoided though at all costs, given that the Dialog component is a Modal Dialog. Only set this to false in case u handle the focustrap manually on the implementation side.


packages/components/src/components/dialog/dialog.component.ts
```diff
@@ -401,7 +407,10 @@ class Dialog extends PreventScrollMixin(FocusTrapMixin(CardAndDialogFooterMixin(
       this.activatePreventScroll();
 
       await this.updateComplete;
-      this.activateFocusTrap?.();
+      console.log(this.focusTrap)
```
gabrielchl, line 410: log

PR #1380, opened 2025-07-01
fix(select): select component enable attribute 'value' and input event

kpankaj3:
### Description

- the value attribute has been updated to reflect the value of current selected option. (if options do not have value attribute set, the value attribute on select component will be empty)
- input event will also be dispatched when an option is selected.
- fix for click function getting called twice when pressing space/enter key on an option.

### Links
https://developer.mozilla.org/en-US/docs/Web/API/Element/input_event


packages/components/src/components/select/select.component.ts
```diff
@@ -59,6 +64,21 @@ class Select extends FormInternalsMixin(DataAriaLabelMixin(FormfieldWrapper)) im
...
   @property({ type: String }) height = 'auto';
 
+  /**
+   * Indicates the current selected value of the select component,
+   * which is same as the `value` attribute of the selected `mdc-option`.
+   *
+   * This attribute can also be used to set the initial selected option.
+   *
+   * Note:
+   * - If mdc-option has a `selected` attribute initially, it will override this value.
```
cscheffauer, line 74: q: should we get rid of the "selected" in the option? having 2 ways to indicate the initial selected value i think can cause confusion

cscheffauer: we will use the "selected" on the option only.

PR #1389, opened 2025-07-09
fix: chip-styling-issue

cscheffauer:
### Description

If there is no label passed in to a chip, there is currently a styling issue in terms of appling the gap.
This PR is about to fix that and also add Screenshots in e2e tests


nataliadelmar (approved):
Approved with 1 comment

PR #1397, opened 2025-07-14
fix: update dot badge styling in HC

2020Deeya:
### Description

Update `DOT` badge styling in High Contrast mode


supminn:
Please update the HC screen shots of all the components that make use of the badge. This would fix the CI pipline.

supminn (approved):

PR #1414, opened 2025-07-22
fix: several select, option and popover fixes

cscheffauer:
#### API Breaking Changes:
- [Option] Component does not support slot / children anymore, the visible label has to be passed as `label` attribute only.
- [Option] Component requires `value` attribute for Select component to work properly.

#### New Features:
- [Option] Added support for secondaryLabel

#### Fixes:
- [Popover] Fix hideOnOutsideClick not considering popover itself (clicking in popover was considered as outside, instead inside)
- [Select] onChange / onInput only fire on initial mount if there is not initial option selected and no placeholder provided (first option will be selected by default then and events fired). If initial value is set (by setting selected), no events are fired on initial mount.
- [Select] Better typing for onChange / onInput customEvents in Select
- [Select] Fix formValidation issue where in case of invalidity the browser was not focusing the invalid select
- [Select] Resetting as part of a form will reset the value to the initial selected value (on mount) instead of setting to empty.
- [Select] Refactoring to code to be more performant / needs less code
- [All] Updated Events in Storybook Controls to start with "Event Name: " to avoid duplication issues with attributes. This is storybook only (applies for all stories)

#### Internal Chores:
- Added vscode task for running Component E2E tests quicker locally


packages/components/src/components/select/select.stories.ts
supminn, line 1: image
`SelectWithLongOptionText` story is not showing the tooltip on hover/focus.

packages/components/src/components/select/select.component.ts
```diff
@@ -536,24 +494,48 @@ class Select extends FormInternalsMixin(DataAriaLabelMixin(FormfieldWrapper)) im
...
           <div part="icon-container">
-            <mdc-icon size="1" length-unit="rem" name="${this.baseIconName}"></mdc-icon>
+            <mdc-icon
+              size="1"
+              length-unit="rem"
+              name="${this.displayPopover ? ARROW_ICON.ARROW_UP : ARROW_ICON.ARROW_DOWN}"
+            ></mdc-icon>
           </div>
         </div>
+        <input
```
supminn, line 520: question: what was the reason to replace <select> by <input>?

supminn: With this change, the native validation message changes

Using select
image

Using input
image

packages/components/src/components/select/select.component.ts
```diff
@@ -338,186 +353,129 @@ class Select extends FormInternalsMixin(DataAriaLabelMixin(FormfieldWrapper)) im
...
+        const currentIndex = options.findIndex(option => option === event.target);
+        const newIndex = Math.max(currentIndex - 1, 0);
+        this.focusAndUpdateTabIndexes(options[newIndex]);
+        event.preventDefault();
+        break;
+      }
+      case KEYS.PAGE_DOWN: {
+        const options = this.getAllValidOptions();
+        const currentIndex = options.findIndex(option => option === event.target);
+        const newIndex = Math.min(currentIndex + 10, options.length - 1);
```
supminn, line 433: question: are we using a fixed number (10) here or do we want to expose this to the consumers too?

PR #1415, opened 2025-07-23
chore: create toast component and add storybook docs

2020Deeya:
### Description

This PR creates Toast component and adds storybook docs.


packages/components/src/components/toast/toast.e2e-test.ts
```diff
@@ -53,176 +64,355 @@ const setup = async (options: ToastSetupOptions) => {
...
+
+    /**
+     * ATTRIBUTES
+     */
     await test.step('Rule: ✅ Attributes', async () => {
       await test.step('Toast sets default attributes', async () => {
         const toast = await setup({ componentsPage, headerText: 'Action Completed' });
         await expect(toast).toHaveAttribute('header-tag-name', 'h2');
-        const header = componentsPage.page.locator('mdc-toast [part="toast-header"]');
-        await expect(header).toHaveAttribute('tagname', 'h2');
```
supminn, line 68: This was also fine, in fact much readable.
Why did you refactor it into single line?

packages/components/src/components/toast/toast.e2e-test.ts
```diff
@@ -53,176 +64,355 @@ const setup = async (options: ToastSetupOptions) => {
...
+        const toast = await setup({ componentsPage, ariaLabel: 'Toast notification' });
+        await expect(toast).toHaveAttribute('aria-label', 'Toast notification');
       });
 
       await test.step('Toast sets the close-button-aria-label attribute', async () => {
-        const closeButtonAriaLabel = 'Dismiss notification';
-        await setup({ componentsPage, closeButtonAriaLabel });
+        await setup({ componentsPage, closeButtonAriaLabel: 'Dismiss notification' });
         const closeBtn = componentsPage.page.locator('mdc-toast [part="toast-close-btn"]');
-        await expect(closeBtn).toHaveAttribute('aria-label', closeButtonAriaLabel);
```
supminn, line 108: this was correct too. why change it to string again?

packages/components/src/components/toast/toast.e2e-test.ts
```diff
@@ -53,176 +64,355 @@ const setup = async (options: ToastSetupOptions) => {
...
+          children: '<mdc-text tagname="span" slot="toast-body-normal">Your file has been successfully uploaded</mdc-text>',
+        });
+
         const closeBtn = componentsPage.page.locator('mdc-toast [part="toast-close-btn"]');
         await closeBtn.waitFor();
+
         const closePromise = componentsPage.waitForEvent(toast, 'close');
         await closeBtn.click();
         await closePromise;
-        await expect(toast).not.toBeVisible();
```
supminn, line 160: why is this line removed?

packages/components/src/components/toast/toast.e2e-test.ts
```diff
@@ -53,176 +64,355 @@ const setup = async (options: ToastSetupOptions) => {
...
+        });
+
         await componentsPage.actionability.pressTab();
         const closeBtn = componentsPage.page.locator('mdc-toast [part="toast-close-btn"]');
         await expect(closeBtn).toBeFocused();
+
         const closePromise = componentsPage.waitForEvent(toast, 'close');
         await closeBtn.press('Enter');
         await closePromise;
-        await expect(toast).not.toBeVisible();
```
supminn, line 205: we need this check, right?

packages/components/src/components/toast/toast.styles.ts
```diff
@@ -0,0 +1,88 @@
...
+    align-items: flex-start;
+    gap: 0.25rem;
+    flex: 1 0 0;
+    align-self: stretch;
+  }
+
+  :host::part(toast-header) {
+    display: -webkit-box;
+    -webkit-box-orient: vertical;
+    -webkit-line-clamp: 2;
```
supminn, line 61: issue: This might work only on webkit browsers.
Can we work with an alternative styling that is supported on all browsers

packages/components/src/components/toast/toast.e2e-test.ts
2020Deeya: No, we don't. The toast will remain open until an external component closes it. It will only emit a close event. This behavior will occur when we write user stories for that external component.

packages/components/src/components/toast/toast.component.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  private renderIcon(iconName: string): TemplateResult {
+    return html`
+      <mdc-icon
+        name="${ifDefined(iconName as IconNames)}"
+        size="${DEFAULTS.PREFIX_ICON_SIZE}"
+        part="toast-prefix-icon"
+      ></mdc-icon>
+    `;
+  }
```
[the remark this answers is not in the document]
2020Deeya: I think, Both approaches are acceptable because iconName will always be present.

packages/components/src/components/toast/toast.e2e-test.ts
```diff
@@ -14,12 +17,17 @@ type ToastSetupOptions = {
...
 };
 
 const SHOW_MORE_TEXT = 'Show more';
 const SHOW_LESS_TEXT = 'Show less';
 
+const isDesktopDevice = () => {
+  const deviceName = test.info().project.name;
+  return ['chrome', 'firefox', 'msedge', 'webkit', 'tablet chrome', 'tablet safari'].includes(deviceName);
+};
+
```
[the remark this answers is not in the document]
2020Deeya: I'm not sure if we really need mobile support, we'd need to confirm that with our team.

packages/components/src/components/toast/toast.component.ts
```diff
@@ -0,0 +1,225 @@
...
+ * @cssproperty --mdc-toast-elevation-3 - Elevation effect applied to the toast.
+ * @cssproperty --mdc-toast-width - Width of the toast.
+ * @cssproperty --mdc-toast-padding - Padding inside the toast.
+ */
+class Toast extends FooterMixin(Component) {
+  /**
+   * The unique ID of the toast
+   */
+  @property({ type: String, reflect: true })
+  override id: string = '';
```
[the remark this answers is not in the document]
supminn: if we don't need it, then we could omit overriding this property altogether

packages/components/src/components/toast/toast.e2e-test.ts
```diff
@@ -53,176 +64,355 @@ const setup = async (options: ToastSetupOptions) => {
...
+        const toggleBtn = toast.locator('mdc-linkbutton[part="footer-button-toggle"]');
+        const detailedSlot = toast.locator('mdc-text[slot="toast-body-detailed"]');
+
+        await toggleBtn.focus();
+
+        if (isDesktopDevice()) {
+          await componentsPage.visualRegression.takeScreenshot('mdc-toast', {
+            source: 'userflow',
+            fileNameSuffix: 'collapsed-view',
+          });
```
supminn, line 386:
```suggestion
          await componentsPage.visualRegression.takeScreenshot('mdc-toast', {
            source: 'userflow',
            fileNameSuffix: 'collapsed-view',
            element: toast
          });
```

packages/components/src/components/toast/toast.component.ts
```diff
@@ -0,0 +1,225 @@
...
+      </mdc-linkbutton>
+    `;
+  }
+
+  protected renderHeader() {
+    return this.headerText
+      ? html`
+          <mdc-text
+            part="toast-header"
+            tagname="${VALID_TEXT_TAGS[this.headerTagName.toUpperCase() as keyof typeof VALID_TEXT_TAGS]}"
```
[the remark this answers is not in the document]
2020Deeya: I disagree, we would need this check.
This mapping ensures that only valid HTML tag names are passed to tagname, satisfying the strict string literal union type required by the component. Without it, passing a plain string like this.headerTagName would throw a TypeScript error because it's not assignable to the expected union type.

packages/components/src/components/toast/toast.component.ts
```diff
@@ -0,0 +1,221 @@
+import type { CSSResult, PropertyValues } from 'lit';
+import { html, nothing } from 'lit';
+import { property, queryAssignedElements, state } from 'lit/decorators.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+import { Component } from '../../models';
+import { FooterMixin } from '../../utils/mixins/FooterMixin';
+import type { IconNames } from '../icon/icon.types';
+import { TYPE, VALID_TEXT_TAGS } from '../text/text.constants';
+
```
supminn, line 10:
```suggestion
import type { TagName } from '../text/text.types';

```

packages/components/src/components/toast/toast.component.ts
```diff
@@ -0,0 +1,221 @@
...
+   */
+  @property({ type: String, reflect: true, attribute: 'header-text' })
+  headerText?: string;
+
+  /**
+   * The html tag to be used for the header text
+   * @default 'h2'
+   */
+  @property({ type: String, reflect: true, attribute: 'header-tag-name' })
+  headerTagName: string = DEFAULTS.HEADER_TAG_NAME;
```
supminn, line 75:
```suggestion
  headerTagName: TagName = DEFAULTS.HEADER_TAG_NAME;
```

packages/components/src/components/toast/toast.component.ts
```diff
@@ -0,0 +1,221 @@
...
+      </mdc-linkbutton>
+    `;
+  }
+
+  protected renderHeader() {
+    return this.headerText
+      ? html`
+          <mdc-text
+            part="toast-header"
+            tagname="${VALID_TEXT_TAGS[this.headerTagName.toUpperCase() as keyof typeof VALID_TEXT_TAGS]}"
```
supminn, line 169:
```suggestion
            tagname="${this.headerTagName}"
```

supminn (changes requested):
Suggestion to handle mdc-text tagname Type

supminn (commented):
image

image


There's extra spacing on these 2 toasts which is not there on figma spec. We would need to remove the empty gap.

packages/components/src/components/toast/toast.stories.ts
```diff
@@ -0,0 +1,175 @@
...
+        render({
+          ...DefaultArgs,
+          ...args,
+        })
+      )}
+    </div>
+  `,
+  parameters: {
+    controls: { disable: true },
+  },
```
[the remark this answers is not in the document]
arjunkesava: It depends on the requirement here.
For a story with all variants, hideAllControls is the only way to prevent user for making this change. 
For a normal story, blocking the user by hiding the controls is not an ideal solution.

arjunkesava (approved):

PR #1428, opened 2025-07-25
chore: convert default as type const

arjunkesava:
### Description

- Add `as const` for all DEFAULT exports in all missing components.
- Replace string `button` with `ROLE.BUTTON`.


packages/components/src/components/button/button.component.ts
```diff
@@ -3,6 +3,8 @@ import { CSSResult, html } from 'lit';
 import { property, state } from 'lit/decorators.js';
 
 import { ButtonComponentMixin } from '../../utils/mixins/ButtonComponentMixin';
+import { ROLE } from '../../utils/roles';
+import type { RoleType } from '../../utils/roles';
```
supminn, line 7: question: do we import RoleType across other components too?
If yes, could you get this updated on the same PR?

PR #1429, opened 2025-07-27
fix: multiple fixes to popover, tooltip, select, formfieldwrapper

cscheffauer:
### New Features:
- [Popover] Added support for tooltip and popover connected to the same trigger and keeping tooltip closed (by using `keepConnectedTooltipClosed` attribute). This is set to true by default.

### Fixes:

**Memory / Detached objects fixes**
- [Popover] Fixing severe Memory Leaks by closing all open event listeners, upgrading floating-ui to latest version etc.
- [Dialog] Remove existing backdrops before creating a new one, to avoid keeping backdrop in detached nodes.
- [Animation] Destroying Lottie instance on unmount to prevent memory leak.
- [Icon] Destroying Abortcontroller instance on unmount to prevent memory leak.
- [Tablist] Removing event listeners to avoid memory leak.

**Several fixes**
- [Popover] Fix for not closing the tooltip when a popover opens a dialog and the focus moves back.
- [Formfieldwrapper] Styling update when disabled, required asterisk is now also grey when disabled instead of staying red
- [Select] Stop propagation of keyboard event if arrow down/ up are pressed (for scenarios where selects are used in a ListItem
- [Select] adding backdrop to make sure outside click is first closing the select, then other actions can be clicked after

**Typescript fixes**
- [Popover] better Typescript types for Popover events
- [Select] better Typescript types for Select events and re-exporting types for bubbled up events of Popover
- [Textarea] fixed typing on rows and cols attribute
- [Spinner] fixed typing on inverted attribute

**Internal Chores:**
- Adding/updating e2e tests and stories for above changes


packages/components/src/components/popover/popover.e2e-test.ts
```diff
@@ -803,6 +803,123 @@ const userStoriesTestCases = async (componentsPage: ComponentsPage) => {
...
+    });
+
+    // Attach click handler to #open-dialog to open dialog and close popover
+    await componentsPage.page.evaluate(() => {
+      const openDialogBtn = document.querySelector('#open-dialog');
+      if (openDialogBtn) {
+        openDialogBtn.addEventListener('click', () => {
+          const dialog = document.querySelector('#dialog') as any;
+          if (dialog) dialog.visible = true;
+          const popover = document.querySelector('#popover') as any;
```
supminn, line 891:
```suggestion
          const dialog = document.querySelector('#dialog') as Dialog;
          if (dialog) dialog.visible = true;
          const popover = document.querySelector('#popover') as Popover;
```

packages/components/src/components/listitem/listitem.events.ts
```diff
@@ -37,7 +37,9 @@ export class ListItemEventManager {
   static onCreatedListItem(instance: ListItem) {
     // Use setTimeout to ensure the event is dispatched after the component is fully initialized
     // Inherited Element might execute its own lifecycle methods
-    setTimeout(() => this.dispatchListItemEvent('created', instance), 0);
+    setTimeout(() => {
+      this.dispatchListItemEvent('created', instance);
+    }, 0);
```
supminn, line 42: question: are they not doing the same thing? Was there a reason to make this change?

packages/components/src/components/menubar/menubar.component.ts
```diff
@@ -78,6 +79,24 @@ class MenuBar extends Component {
...
+  private getVisibleSubMenusOfItems(items: Array<HTMLElement>): Array<MenuPopover> {
+    const subMenus = <Array<MenuPopover>>[];
+    if (!items || items.length === 0) return subMenus;
+
+    items.forEach(item => {
+      const id = item.getAttribute('id');
+      if (!id) return;
+
+      const submenu = this.querySelector?.(`${MENUPOPOVER_TAGNAME}[triggerid="${id}"]`) as MenuPopover;
+      if (submenu && submenu.visible) {
```
supminn, line 92: nitpick:
```suggestion
      if (submenu?.visible) {
```

packages/components/src/components/menupopover/menupopover.component.ts
```diff
@@ -297,7 +349,7 @@ class MenuPopover extends Popover {
       // `disabled` could bubble up, but it is more consistent to handle it the same way as `destroyed`
       item.addEventListener('disabled', this.handleItemChangeEvent);
     }
-  }
+  };
 
   private handleItemChangeEvent = (event: Event) => {
     event.stopImmediatePropagation();
```
supminn, line 355: question: Are we not changing all the arrow function implementation?

packages/components/src/components/popover/popover.component.ts
```diff
@@ -351,158 +360,220 @@ class Popover extends PreventScrollMixin(FocusTrapMixin(Component)) {
...
-      hoverBridge?.addEventListener('mouseenter', this.showPopover);
     }
     if (this.trigger.includes('focusin')) {
-      this.triggerElement.addEventListener('focusin', this.showPopover);
+      this.triggerElement.addEventListener('focusin', this.show);
       if (!this.interactive) {
         this.triggerElement.addEventListener('focusout', this.handleFocusOut);
       }
     }
-    this.addEventListener('focus-trap-exit', this.hidePopover);
```
supminn, line 435: question: Do we not require this event anymore?

supminn (approved):
Added 2 questions and nitpick.

packages/components/src/components/listitem/listitem.events.ts
cscheffauer: they are, for me typing it properly out as a void function in this case was better to read and also its not necessary correct - if u do `() => returnValue` you get something as a return, whereas `() => {}` is a void

PR #1438, opened 2025-07-31
fix: Align image height with Card body content

2020Deeya:
### Description
Adjusts the image height in the Card component to match the dynamic height of the Card body. This ensures consistent alignment and resolves layout mismatches in horizontally aligned content.

#### Fixes

- Image height not matching dynamic Card body height
- Misalignment in horizontal Card layout due to static image height

### DIFF
--- a/packages/components/src/components/cardbutton/cardbutton.stories.ts
+++ b/packages/components/src/components/cardbutton/cardbutton.stories.ts
@@ -183,10 +183,10 @@ export const ContentBeforeBody: StoryObj = {
     'title-tag-name': DEFAULTS.TAGNAME,
     'subtitle-tag-name': DEFAULTS.TAGNAME,
     orientation: DEFAULTS.ORIENTATION,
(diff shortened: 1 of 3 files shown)


packages/components/src/components/cardcheckbox/cardcheckbox.component.ts
```diff
@@ -167,7 +167,7 @@ class CardCheckbox extends DisabledMixin(TabIndexMixin(Card)) {
 
   public override render() {
     return html`
-      ${this.renderImage()}
+      <div part="image-wrapper">${this.renderImage()}</div>
```
cscheffauer, line 170: why not doing this in the `this.renderImage()` function?
same applies for all other occurrences

cscheffauer: another issue, if there is NO image this div will still be there unnecessarily and might cause problems

2020Deeya: Fixed.

PR #1441, opened 2025-08-01
fix: remove nested css selector in popover.styles.css

dominiccarrington:
### Description

Remove the nested selector in popover.styles.css and replace with the equivalent expanded version. This appears to be causing a crash in Safari 17

(Also, update the badge for menupopover from wip to stable since it is in the stable area of the storybook) 

### Links

WebKit Bug: https://bugs.webkit.org/show_bug.cgi?id=276942


chrisadubois (approved):
nice find

PR #1442, opened 2025-08-04
fix: provide clear icon click event

2020Deeya:
### Description

Introduces a dedicated clear icon click event in the SearchField component. 

#### Fixes
Lack of a specific event for clear icon click in SearchField, leading to conflicts with existing focus and change handlers used for separate logic.


packages/components/src/components/searchfield/searchfield.types.ts
```diff
@@ -0,0 +1,15 @@
+import { TypedEvent } from "../../utils/types";
+import type Input from '../input/input.component';
+
+type InputClearEvent = TypedEvent<Input>;
```
2020Deeya, line 4: Note: `TypedEvent` will be renamed to `TypedCustomEvent` in an upcoming PR. For now, we are retaining the existing name.

cscheffauer (approved):

PR #1451, opened 2025-08-06
fix(button): make ellipsis work for child slot instead of just mdc-text component

jor-row:
### Description

The current implementation only applies the ellipsis styling if the child is a mdc-text component. This PR fixes that so that plain text can be passed as a child to the button and the ellipsis will still work.

MOMENTUM-755


packages/components/src/components/button/button.component.ts
```diff
@@ -43,7 +43,7 @@ import { getIconNameWithoutStyle } from './button.utils';
  *
  * @tagname mdc-button
  *
- * @slot - Text label of the button.
+ * @slot button-text - Text label of the button.
```
supminn, line 46:
```suggestion
 * @slot - Text label of the button.
 *
 * @csspart button-text - Text label of the button, passed in default slot
```

supminn (commented):
suggested changes on ts-doc. 
Implementation looks good!

supminn (approved):

PR #1465, opened 2025-08-12
test: user stories, test cases and attributes for Slider component

supminn:
### Description

Slider component - Template, User stories, Test cases and attributes are defined on components file

MOMENTUM-756


packages/components/src/components/slider/slider.feature
```diff
@@ -0,0 +1,214 @@
...
+
+    Examples:
+      | key        | action               |
+      | ArrowRight | increase by 5 steps  |
+      | ArrowUp    | increase by 5 steps  |
+      | ArrowLeft  | decrease by 5 steps  |
+      | ArrowDown  | decrease by 5 steps  |
+      | Home       | set to minimum value |
+      | End        | set to maximum value |
+
```
arjunkesava, line 98: The edge cases are not covered.

- If the step value is 5, and if the user set and even number, how will the slider respond?
- If the user provide a number which is out of the range (not in between the start and end), how will the slider show?
- If the user provides not a number (NaN) (or anything else other than number).

PR #1478, opened 2025-08-20
fix: duplicate id issue in form related components by introducing data-id

2020Deeya:
### Description
Previously, the same `this.id` was being applied both to the host `<mdc-input>` and the internal native `<input>`.  
This caused duplicate `id` attributes in the DOM, leading to issues when querying elements by `id`.

#### Fix
- Introduced `data-id` as the attribute to be passed down to the native `<input>`. 
- Ensured all form-related input components follow the same pattern.

#### BREAKING CHANGE
Passing `id` to `<mdc-input>` **no longer propagates to the native `<input>`**.
If you relied on the host `id` also being the native input’s `id`, you must now provide `data-id`.


2020Deeya:
Switched to different approach of fixing this, hence closing the PR.

PR #1481, opened 2025-08-21
fix: update Text default type

2020Deeya:
#### Description
Set default mdc-text type to body-midsize-medium (14px) for consistency.

### Fix
- Set mdc-text default type to body-midsize-medium (14px).

### Breaking Change ⚠️

The default `mdc-text` type has been changed to body-midsize-medium (14px).
Consumers relying on the previous default will need to update their usage accordingly if required.


packages/components/src/components/input/input.component.ts
```diff
@@ -24,6 +24,18 @@ import styles from './input.styles';
  * - leading icon - displayed before the input field.
  * - clear-aria-label - aria label for the trailing button.
  * - all the attributes of the input field.
+ * 
+ * ### Help Text & Validation
```
cscheffauer, line 28: issue: this is now for the input component only, but in reality it applies for other form fields too.
Instead of adding it here i would recommend to think about grouping all input related components into a group in storybook and add a separate generic docs page where this will be mentioned

supminn (approved):

PR #1483, opened 2025-08-22
fix(progressspinner): handle `--mdc-spinner-size` correctly

dominiccarrington:
### Description

Update the `mdc-progressspinner` component to still be correctly styled when the size is changed by the `--mdc-spinner-size` css variable.


dominiccarrington:
Reproduce: Open https://momentum-design.github.io/momentum-design/storybook-static/index.html?path=/story/components-progressspinner--example and set the `--mdc-spinner-size` to any css size.

| Before | After |
| --- | --- |
| Screenshot 2025-08-22 at 13 57 29 | Screenshot 2025-08-22 at 13 59 42 |

PR #1485, opened 2025-08-25
refactor: building single value slider component 

supminn:
### Description

This PR covers the implementation for single-value slider. Please review only this component
We will extend the same component to support range slider as well (this is WIP)


packages/components/src/components/slider/slider.component.ts
```diff
@@ -1,153 +1,592 @@
...
+    this.dispatchEvent(new CustomEvent('change', { detail: { valueStart: this.valueStart } }));
+  }
+
+  /**
+   * Handles the input event for the end thumb of the range value slider.
+   * @param e - The input event.
+   */
+  onInputEnd() {
+    this.handleInputEnd();
+    this.dispatchEvent(new CustomEvent('input', { detail: { valueEnd: this.valueEnd } }));
```
2020Deeya, line 419: suggestion: Better to include full state ({ valueStart, valueEnd }) for consumers.

packages/components/src/components/slider/slider.component.ts
```diff
@@ -1,153 +1,592 @@
...
+    this.dispatchEvent(new CustomEvent('change', { detail: { value: this.value } }));
+  }
+
+  /**
+   * Handles the input event for the start thumb of range value slider.
+   * @param e - The input event.
+   */
+  onInputStart() {
+    this.handleInputStart();
+    this.dispatchEvent(new CustomEvent('input', { detail: { valueStart: this.valueStart } }));
```
2020Deeya, line 400: suggestion: Better to include full state ({ valueStart, valueEnd }) for consumers.

packages/components/src/components/slider/slider.component.ts
```diff
@@ -1,153 +1,592 @@
...
+    `;
+  }
+
+  /**
+   * Updates the styling of the slider track.
+   * The progress value is calculated and updated using appropriate tokens
+   * In a range slider, both thumbs are considered.
+   * The track is filled between the two thumbs.
+   */
+  updateTrackStyling() {
```
2020Deeya, line 342: `updateTrackStyling` sets inline styles directly. Harder to theme or override.
suggestion: If possible, let's move to CSS.

packages/components/src/components/slider/slider.component.ts
```diff
@@ -1,153 +1,592 @@
...
+  }
+
+  /**
+   * Handles the change event for the end thumb of the range slider.
+   * @param e - The change event.
+   */
+  onChangeEnd(e: Event) {
+    const input = e.target as HTMLInputElement;
+    this.valueEnd = Number(input.value);
+    this.dispatchEvent(new CustomEvent('change', { detail: { valueEnd: this.valueEnd } }));
```
2020Deeya, line 429: suggestion: Better to include full state ({ valueStart, valueEnd }) for consumers.

packages/components/src/components/slider/slider.component.ts
```diff
@@ -1,153 +1,592 @@
...
+  }
 
   /**
-   * Aria value text for the slider's value displayed when range is false.
+   * Handles input changes for the start thumb of the range slider.
+   * The start thumb will never cross over the end thumb
    */
-  @property({ type: String, attribute: 'data-aria-valuetext' }) dataAriaValuetext: string = '';
+  private handleInputStart() {
+    const input = this.inputElements[0];
```
2020Deeya, line 266:
```suggestion
    const input = this.inputElements?.[0];
    if (!input) return;
```

packages/components/src/components/slider/slider.component.ts
```diff
@@ -1,153 +1,592 @@
...
+  }
 
   /**
-   * Aria label for the slider's handle displayed when range is false.
+   * Handles input changes for the end thumb of the range slider.
+   * The end thumb will never cross over the start thumb
    */
-  @property({ type: String, attribute: 'data-aria-label' }) dataAriaLabel: string = '';
+  private handleInputEnd() {
+    const input = this.inputElements[1];
```
2020Deeya, line 281:
```suggestion
    const input = this.inputElements?.[1];
    if (!input) return;
```

packages/components/src/components/slider/slider.types.ts
```diff
@@ -1,5 +1,11 @@
-// import type { ValueOf } from '../../utils/types';
+import { Slider } from '../..';
+import { OverrideEventTarget, TypedCustomEvent } from '../../utils/types';
```
2020Deeya, line 2:
```suggestion
import type { OverrideEventTarget, TypedCustomEvent } from '../../utils/types';

import type Slider from './slider.component';
```

packages/components/src/components/slider/slider.component.ts
supminn: As we discussed, this will still let us update according to the theme - since we are using theme tokens.
The styling logic is based on the latest value of the slider thumb, hence it has to be set using JavaScript.

packages/components/src/components/slider/slider.styles.ts
```diff
@@ -1,9 +1,194 @@
...
+    border-radius: 0.25rem;
+    outline: none;
+    margin: 0;
+    position: absolute;
+    pointer-events: none;
+  }
+
+  input[type='range']::-webkit-slider-thumb {
+    -webkit-appearance: none;
+    appearance: none;
```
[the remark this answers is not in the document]
supminn: Line 70 and 71, we are applying it for the slider line. Here, we are applying it for the slider thumb.

This is how the slider thumb would look if this code is removed. The custom styles do not get applied.
image

packages/components/src/components/slider/slider.styles.ts
```diff
@@ -1,9 +1,204 @@
...
+    display: flex;
+    justify-content: space-between;
+  }
+
+  :host::part(leading-icon),
+  :host::part(trailing-icon) {
+    flex-shrink: 0;
+  }
+
+  #start-slider {
```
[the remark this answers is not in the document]
supminn: I discussed this with you earlier. This styles wasn't working with parts. Maybe its specific to the slider component (native)

PR #1486, opened 2025-08-25
chore: add static image fixtures for Storybook and Playwright tests

2020Deeya:
### Description
Previously, multiple components in Storybook and Playwright tests relied on 3rd party image URLs. This added external dependencies, caused potential instability, and risked inconsistent rendering.

### Changes
- Added `pngFixtures/` under Playwright setup to host all static images.
- Created a `pngToDataUrl` script to convert PNGs into data URLs.
- Generated `imageFixtures.ts` to export data URLs for reuse across components.
- Updated Storybook stories and Playwright tests to use these fixtures consistently.


packages/components/config/playwright/setup/pngToDataurl.js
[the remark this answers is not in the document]
2020Deeya: I agree. We should do that.

packages/components/config/playwright/setup/pngToDataurl.js
```diff
@@ -0,0 +1,42 @@
...
+ * AUTO-GENERATED FILE CREATOR
+ *
+ * This script reads all `.png` files inside `pngFixtures/`,
+ * converts them into base64 data URLs, and writes them into
+ * `imageFixtures.ts` for use in Storybook + Playwright tests.
+ *
+ * Run with: `yarn generate:image-fixtures`
+ */
+
+const fs = require('fs');
```
cscheffauer, line 11: issue: this is not clear at all in terms of folder structure what this file is about. i would highly recommend to structure better, for example create a scripts folder here in playwright or move this file to the already existing `src/scripts`

packages/components/config/playwright/setup/pngToDataurl.js
```diff
@@ -0,0 +1,42 @@
...
+const path = require('path');
+
+const inputFolder = path.resolve(__dirname, './pngFixtures');
```
cscheffauer, line 14: suggestion: this script might not work on windows machine, check the other scripts in `src/scripts` - they make use of `path.posix` to overcome those issues.

packages/components/package.json
```diff
@@ -71,6 +71,7 @@
     "figma:unpublish:webcomponent": "yarn figma connect unpublish --config figma-webcomponent.config.json",
     "figma:unpublish": "yarn figma:unpublish:webcomponent && yarn figma:unpublish:react",
     "generate:component": "yarn build:plop && plop --plopfile config/plop/dist/plopfile.mjs",
+    "generate:image-fixtures": "node ./config/playwright/setup/pngToDataurl.js",
```
cscheffauer, line 74: issue: on CI this will break, since this is not running before the e2e tests. Checkout the github actions file and this package.json and identify where the best spot is to put this script.

packages/components/scripts/pngToDataurl.js
```diff
@@ -0,0 +1,60 @@
...
+let existingFixtures = {};
+if (fs.existsSync(outputFile)) {
+  try {
+    const content = fs.readFileSync(outputFile, 'utf8');
+    const match = content.match(/export const imageFixtures = (.*) as const;/s);
+    if (match) {
+      existingFixtures = JSON.parse(match[1]);
+    }
+  } catch (e) {
+    console.warn('⚠️ Could not parse existing imageFixtures.ts, regenerating...');
```
supminn, line 26: question: Does this script automatically regenerate after going through this catch block?

```suggestion
    console.warn('⚠️ Could not parse existing imageFixtures.ts, try again...');
```

packages/components/scripts/pngToDataurl.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+});
+
+if (changed) {
+  const output = `/**
+ * AUTO-GENERATED FILE – DO NOT EDIT
+ * Run \`yarn generate:image-fixtures\` to regenerate.
+ */
+
+export const imageFixtures = ${JSON.stringify(newFixtures, null, 2)} as const;
```
supminn, line 54: question: Is it really necessary to pass the 'space' value?
```suggestion
export const imageFixtures = ${JSON.stringify(newFixtures)} as const;
```

2020Deeya: It’s not necessary for the script to work, but it’s useful for maintainability. I think, we should keep it.

PR #1490, opened 2025-08-26
fix(button): add missing name and value attributes

maxinteger:
### Description

Add missing name and value attributes to Buttonsimple via the new NameAndValueMixin


packages/components/src/components/buttonsimple/buttonsimple.component.ts
```diff
@@ -93,6 +93,18 @@ class Buttonsimple extends AutoFocusMixin(TabIndexMixin(DisabledMixin(Component)
   @property({ reflect: true })
   type: ButtonType = DEFAULTS.TYPE;
 
+  /**
+   * The name of the button, submitted as a pair with the button's value as part of the form data,
+   * when that button is used to submit the form.
+   */
+  @property({ type: String, reflect: true }) name?: undefined | string = undefined;
```
cscheffauer, line 100: nitpick: the `?` is already making it undefined, so undefined as a type is not necessary.
Also setting the default value to undefined is not necessary either.
same below

cscheffauer (dismissed):

cscheffauer (approved):

PR #1491, opened 2025-08-26
fix(button-group): buttong group height fixed

maxinteger:
### Description

Button group secondary variant got extra border and its height was 2px bigger then a button with the same size.


cscheffauer (approved):
lgtm!

PR #1500, opened 2025-08-29
feat: created static chip component

supminn:
### Description

Created a new component mdc-staticchip which is a static representation of label chip.
It is not an interactive element. Hence, it doesn't have disabled state too.

MOMENTUM-784


packages/components/src/components/staticchip/staticchip.feature
```diff
@@ -0,0 +1,30 @@
+Feature: StaticChip Accessibility, Appearance, and User Interaction
```
2020Deeya, line 1: suggestion: Better to replace `StaticChip` with the actual component name, `staticchip`

packages/components/src/components/staticchip/staticchip.component.ts
```diff
@@ -0,0 +1,79 @@
...
+import type { ColorType } from './staticchip.types';
+import styles from './staticchip.styles';
+
+/**
+ * mdc-staticchip is an static element that can be used to represent a chip. It supports a leading icon along with label.
+ *
+ * It is recommended to keep the label text for the chip component concise and compact.
+ * For best results, we recommend limiting the <b>maximum length of the label text to 20 characters</b>, including empty spaces to split words.
+ *
+ * This component is built by extending Buttonsimple.
```
2020Deeya, line 18: Not needed for this, it's not extended from `Buttonsimple`

2020Deeya (approved):

PR #1502, opened 2025-08-29
fix(listitem): ensure trailing content doesn’t overlap when width shrinks

2020Deeya:
### Description
When ListItem width was constrained, both leading and trailing parts were forced to width: 100%, causing trailing content to overlap and preventing proper shrink/ellipsis.

#### Fix:

- Updated layout so the leading section flexes with available space while the trailing section sizes to its content, preventing overlap.

### DIFF
--- a/packages/components/src/components/listitem/listitem.styles.ts
+++ b/packages/components/src/components/listitem/listitem.styles.ts
@@ -79,9 +79,13 @@ const styles = css`
     align-items: center;
     column-gap: var(--mdc-listitem-column-gap);
     display: flex;
-    width: 100%;
+  }
+  :host::part(leading) {
+    flex: 1;
+    min-width: 0;
   }
   :host::part(trailing) {
+    flex: 0 0 auto;
     justify-content: flex-end;
   }
   :host::part(leading-text-secondary-label),


packages/components/src/components/listitem/listitem.styles.ts
```diff
@@ -96,12 +97,24 @@ const styles = css`
...
+
+  :host(:dir(ltr))::part(trailing-text-side-header),
+  :host(:dir(ltr))::part(trailing-text-subline) {
+    margin-left: var(--mdc-listitem-column-gap);
+  }
+
+  :host(:dir(rtl))::part(trailing-text-side-header),
+  :host(:dir(rtl))::part(trailing-text-subline) {
+    margin-right: var(--mdc-listitem-column-gap);
+  }
```
maxinteger, line 109: Please use [logical margin](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-start) to handle rtl/ltr:


```suggestion
  :host()::part(trailing-text-side-header),
  :host()::part(trailing-text-subline) {
    margin-inline-start: var(--mdc-listitem-column-gap);
  }
```

packages/components/src/components/option/option.styles.ts
```diff
@@ -2,29 +2,32 @@ import { css } from 'lit';
...
   }
-  :host::part(leading-text) {
-    flex: 1;
-    /** 2x of leading and trailing icon width + 2x of column gap on both sides of the label text */
-    width: calc(100% - (2 * var(--mdc-option-icon-width)) - (2 * var(--mdc-listitem-column-gap)));
+  :host(:dir(ltr))::part(trailing-icon) {
+    margin-left: var(--mdc-listitem-column-gap);
+  }
+  :host(:dir(rtl))::part(trailing-icon) {
+    margin-right: var(--mdc-listitem-column-gap);
```
maxinteger, line 30: Same as above

```css
  :host()::part(leading-icon) {
    margin-inline-end: var(--mdc-listitem-column-gap);
  }
  :host()::part(trailing-icon) {
    margin-inline-start: var(--mdc-listitem-column-gap);
  }
```

maxinteger (changes requested):

maxinteger (approved):

PR #1503, opened 2025-08-31
fix(popover): fix multi popover aria attributes

maxinteger:
### Description

- remove disable-aria-haspopup attribute from popover
- control aria-expanded and aria-haspopup with disable-aria-expanded
- do not remove aria-expanded and aria-haspopup when disable-aria-expanded become false to keep previous value which probably registered by another popover.


packages/components/src/components/popover/popover.e2e-test.ts
```diff
@@ -658,7 +676,10 @@ const userStoriesTestCases = async (componentsPage: ComponentsPage) => {
         'disable-aria-expanded': 'true',
         visible: 'true',
       });
-      await expect(triggerButton).not.toHaveAttribute('aria-expanded');
+      // disable-aria-expanded changed after first update
+      // it will not remove already existing aria-expanded adn aria-haspopup attributes!
```
supminn, line 680:
```suggestion
      // it will not remove already existing aria-expanded and aria-haspopup attributes!
```

cscheffauer (changes requested):

packages/components/src/components/popover/popover.e2e-test.ts
```diff
@@ -1054,6 +1075,13 @@ test('mdc-popover', async ({ componentsPage }) => {
     await attributeTestCases(componentsPage);
   });
 
+  /**
+   * disabled-aria-expanded ATTRIBUTE
+   */
+  await test.step('disabled-aria-expanded attribute for popover component', async () => {
```
cscheffauer, line 1081: question: why is this a separate test step and not part of the existing attribute test cases

maxinteger: My understanding is, `componentsPage.setAttributes` runs after the `firstUpdate`, so if I want to test a popup with `disabled-aria-expanded=true` properly then I have to set the attribute with the setup function.

packages/components/src/components/popover/popover.e2e-test.ts
```diff
@@ -276,7 +276,7 @@ const attributeTestCases = async (componentsPage: ComponentsPage) => {
     await expect(popover).toHaveAttribute('aria-labelledby', 'popover-label');
     await expect(popover).toHaveAttribute('aria-describedby', 'popover-description');
     await expect(popover).toHaveAttribute('role', DEFAULTS.ROLE);
-    await expect(triggerButton).not.toHaveAttribute('aria-expanded');
+    await expect(triggerButton).toHaveAttribute('aria-expanded');
```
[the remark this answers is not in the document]
maxinteger: This change is necessary and correct.

we relaxed the effect of the `disable-aria-expanded` attr change. It does not remove the `aria-expanded` and `aria-haspopup` attrs anymore even if it change from `false` to `true`.

`componentsPage.setAttributes` changes the attributes after the `firstUpdate`.

So in this case when we rendered the popover `disable-aria-expanded` was false and it set the `aria-expanded` and `aria-haspopup` attrs on the trigger.

Later `componentsPage.setAttributes` changed `disable-aria-expanded` to `true`, but it does not have effect on the 2 ares attrs anymore.

cscheffauer (dismissed):

cscheffauer (approved):

PR #1506, opened 2025-09-03
fix: popover max-width handling and container width issue

2020Deeya:
### Description
- Popover max-width was not applied correctly in some cases.
- When a fixed width was applied to MenuPopover, long menu items could still overflow beyond the popover boundary.

#### Fix
- Applied proper `max-width` handling in Popover to prevent overlap and ensure `--mdc-popover-max-width` takes effect.
- Fixed width handling in Popover to ensure that when a fixed width is set, menu items wrap/truncate as expected and do not overlap or overflow outside the component.

### DIFF
--- a/packages/components/src/components/menupopover/menupopover.e2e-test.ts
+++ b/packages/components/src/components/menupopover/menupopover.e2e-test.ts
@@ -198,9 +198,6 @@ test('mdc-menupopover', async ({ componentsPage }) => {
 
       const trigger = componentsPage.page.locator('#trigger-btn');
       const menupopover = componentsPage.page.locator('mdc-menupopover[triggerid="trigger-btn"]');
-      await trigger.click();
-      await expect(menupopover).toBeVisible();
-
       // Override CSS variable to force small width for truncation
       await componentsPage.page.evaluate(() => {
         const popover = document.querySelector('mdc-menupopover');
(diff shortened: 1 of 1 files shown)


maxinteger (changes requested):
Looks good, but it cause some issues in the menus, maybe because #1502 not merge yet

1. the arrow should has it is own column in the menu same way as the checmark in the mdc-option. Look like the arrow is in part of the width calculation for some reason

Image

VS

Image

Notice the arrow shifted a bit on the second pic.

2. The submenu arrow overflows when the menu width set. It is not Ok here because the text could shrike more. Probably related to the first problem.

Image

same happens with the checkmark

Image

2020Deeya:
> Looks good, but it cause some issues in the menus, maybe because #1502 not merge yet
> 
> 1. the arrow should has it is own column in the menu same way as the checmark in the mdc-option. Look like the arrow is in part of the width calculation for some reason
> 
> Image
> VS
> 
> Image
> Notice the arrow shifted a bit on the second pic.
> 
> 2. The submenu arrow overflows when the menu width set. It is not Ok here because the text could shrike more. Probably related to the first problem.
> 
> Image
> same happens with the checkmark
> 
> Image

It will get resolved once this PR is merged. 
https://github.com/momentum-design/momentum-design/pull/1502

maxinteger (approved):

PR #1524, opened 2025-09-15
chore: add list navigation mixin on select component

arjunkesava:
### Description

- Add ListNavigationMixin and CaptureDestroyEventForChildElement mixin on Select component.
- Remove keyboard navigation on select component.
- Create new utils file and move helper methods of select on select.utils.ts file.
- No UI / Functionality changes on this PR.

### BREAKING CHANGE
- This PR removes the updateState() method call.
- From hereonwards, when the selected attribute of an option changes or when the value of select attribute is changed, then the component automatically gets updated, without the need to call `updateState()` method.


packages/components/src/components/select/select.utils.ts
```diff
@@ -0,0 +1,22 @@
+import { TAG_NAME as OPTION_TAG_NAME } from '../option/option.constants';
```
cscheffauer, line 1: issue: this is not necessary. The whole point of the itemStore is to be able to get all that information from there, instead of using the querySelector approach.

cscheffauer (changes requested):

packages/components/src/components/select/select.component.ts
```diff
@@ -79,7 +90,13 @@ class Select
   @property({ type: Boolean }) readonly = false;
 
   /**
-   * The placeholder text which will be shown on the text if provided.
+   * The placement of the popover within Combobox.
+   * This defines the position of the popover relative to the combobox input field.
```
cscheffauer, line 94: this is not a combobox, but a select

packages/components/src/components/select/select.component.ts
```diff
@@ -1,18 +1,22 @@
 import type { PropertyValueMap, PropertyValues } from 'lit';
```
cscheffauer, line 1: suggestion: we currently have the "updateState()" function in the select component - that can be removed and replaced by using the modify event from the elements in the store (see mixins / controllers for reference). The story "SelectWithChangingSelectedAfterMount" should then still work without calling the updateState() method. Would be great to do this while we are here.

cscheffauer (dismissed):

packages/components/src/components/select/select.component.ts
```diff
@@ -155,31 +169,91 @@ class Select
...
+    });
   }
 
-  private getFirstValidOption(): Option | null {
-    return this.slottedListboxes[0]?.querySelector(`${OPTION_TAG_NAME}:not([disabled])`);
+  override connectedCallback(): void {
+    super.connectedCallback();
+
+    this.loop = 'false';
+    this.initialFocus = 0;
```
maxinteger, line 188: `connectedCallback` called every time when the the component mounted, do we really want to reset both value every time?

It should be Ok to initialise them in the constructor

packages/components/src/components/select/select.component.ts
```diff
@@ -155,31 +169,91 @@ class Select
...
+   *
+   * @internal
+   */
+  private handleDestroyEvent = (event: Event) => {
+    const destroyedElement = event.target as HTMLElement;
+    if (!this.isValidItem(destroyedElement) || destroyedElement.tabIndex !== 0) {
+      return;
+    }
+
+    const destroyedItemIndex = this.navItems.findIndex(node => node === destroyedElement);
```
maxinteger, line 237: `findIndex` is for search based on generic predicate, but here you just compare references and `indexOf` is enough for that:

```suggestion
const destroyedItemIndex = this.navItems.indexOf(destroyedElement);
```

packages/components/src/components/select/select.component.ts
```diff
@@ -155,31 +169,91 @@ class Select
...
+    if (destroyedItemIndex === -1) {
+      return;
+    }
+
+    let newIndex = destroyedItemIndex + 1;
+    if (newIndex >= this.navItems.length) {
+      newIndex = destroyedItemIndex - 1;
+    }
+
+    this.resetTabIndexes(newIndex);
```
maxinteger, line 247: If you have only one element in the list `destroyedItemIndex` will `0` and the `newIndex` will be `-1` at this point.

Please handle the case when the last item removed. 

Since we can not move the focus to other option we should close the select dropdown and move the focus back to the select.

maxinteger (changes requested):

packages/components/src/components/select/select.component.ts
maxinteger: Thanks for the explanation

packages/components/src/components/select/select.component.ts
maxinteger: @supminn You are right `if (newIndex >= this.navItems.length)` will prevent to set the `newIndex` to -1, but we still call `resetTabIndexes` with the wrong value, if you remove the last item there is no valid newIndex left in the select.

Instead we should do something like this:

```ts
    if (newIndex >= this.navItems.length) 
      newIndex = destroyedItemIndex - 1;
    } else {
      return this.hideDropdownAndMoveFocusToTheDropdownTrigger()
    }
```

PR #1531, opened 2025-09-17
fix(virtualizedlist): improve virtualizedlist

dominiccarrington:
(Still WIP)

### Description

Rewrite the virtualizedlist to:
- add keyboard navigation
- keep active (focused/selected) element available in the DOM
- add scroll anchoring when items are added/removed from the list
- add the correct aria properties to the list items so screen reader reads the list correctly

This PR also fixes a bug where if the last element in a list is the active element and is removed from the DOM, the focus is always moved to the new last element.

**Breaking Changes**
The API for the virtualizedlist component has been completely updated. You still need to pass the virtualizerProps but the virtual items are now provided by an event instead of a callback. 
(Following the standard DOM modal: https://lit.dev/docs/composition/component-composition/#passing-data-up-and-down-the-tree)

### DIFF
--- a/packages/components/src/components/listitem/listitem.component.ts
+++ b/packages/components/src/components/listitem/listitem.component.ts
@@ -125,6 +125,9 @@ class ListItem extends DisabledMixin(TabIndexMixin(LifeCycleMixin(Component))) {
   @property({ type: Boolean, reflect: true, attribute: 'soft-disabled' })
   softDisabled?: boolean;
 
+  @property({ type: Number, reflect: true, attribute: 'data-index' })
+  dataIndex?: number;
+
   constructor() {
     super();
 
@@ -224,6 +227,10 @@ class ListItem extends DisabledMixin(TabIndexMixin(LifeCycleMixin(Component))) {
 
     if (changedProperties.has('softDisabled')) {
       this.disableSlottedChildren(this.softDisabled);
+    }
+
+    if (changedProperties.has('dataIndex')) {
+      this.ariaPosInSet = `${this.dataIndex !== undefined ? this.dataIndex + 1 : ''}`;
     }
   }
 
--- a/packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
+++ b/packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
@@ -74,6 +74,8 @@ class VirtualizedList extends DataAriaLabelMixin(List) {
 
   private hiddenIndexes: number[] = [];
 
+  private isAtBottom: boolean = false;
+
   /**
    * Create the virtualizer controller and the virtualizer instance when the component is first connected to the DOM.
    */
@@ -140,6 +142,12 @@ class VirtualizedList extends DataAriaLabelMixin(List) {
       this.selectedIndex = Math.max(0, Math.min(this.virtualizer.options.count - 1, newSelectedIndex));
       this.requestUpdate();
       await this.updateComplete;
+
+      if (this.isAtBottom) {
+        this.scrollElementRef.value.scrollTop = this.scrollElementRef.value.scrollHeight;
+
+        return;
+      }
 
       const scrollDifference = this.scrollElementRef.value!.scrollHeight - previousScrollHeight;
       const shouldAdjustScroll =
@@ -329,6 +337,18 @@ class VirtualizedList extends DataAriaLabelMixin(List) {
    * Refires the scroll event from the internal scroll container to the host element
    */
   private handleScroll(event: Event): void {
+    const scrollElement = this.scrollElementRef.value;
+    if (!scrollElement) {
+      // We really shouldn't get here
+      return;
+    }
+
+    if (this.virtualizer) {
+      const lastItemSize = this.virtualizer.options.estimateSize(this.virtualizer.options.count - 1);
+      this.isAtBottom =
+        scrollElement.scrollHeight - scrollElement.scrollTop <= scrollElement.clientHeight + lastItemSize;
+    }
+
     const EventConstructor = event.constructor as typeof Event;
     this.dispatchEvent(new EventConstructor(event.type, event));
   }
--- a/packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
+++ b/packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
@@ -40,13 +40,17 @@ class VirtualizedList extends DataAriaLabelMixin(List) {
 
   /**
    * Object that sets and updates the virtualizer with any relevant props.
-   * There are two required object props in order to get virtualization to work properly.
-   * count - The length of your list that you are virtualizing.
+   * There are three required object props in order to get virtualization to work properly.
+   *
+   * **count** - The length of your list that you are virtualizing.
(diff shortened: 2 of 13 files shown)


packages/components/src/components/listitem/listitem.component.ts
```diff
@@ -130,6 +130,17 @@ class ListItem extends DisabledMixin(TabIndexMixin(LifeCycleMixin(Component))) {
   @property({ type: Boolean, reflect: true, attribute: 'soft-disabled' })
   softDisabled?: boolean;
 
+  /**
+   * Data attribute to define the index of the list item in a list.
+   * This is used to set the `aria-posinset` attribute for accessibility purposes.
+   *
+   * This will be set by the virtualized list automatically.
```
cscheffauer: this is a public attribute now, so it might be very confusing for consumers what to do with this. In a normal list aria-posinset doesn't need to be set on the listitem, so as a consumer i would be very confused on when i have to set this and when not. Please make this more clear or at least remove this dataIndex from storybook...

packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
```diff
@@ -1,180 +1,673 @@
...
+  /**
+   * The virtualizer instance created by the VirtualizerController.
+   */
+  public virtualizer: Virtualizer | null = null;
+
+  /**
+   * A ref to the scrollable element within the list.
+   */
+  @query('[part="scroll"]', true)
+  public $scroll!: HTMLElement;
```
cscheffauer: this is introducing a new property naming convention with the `$` - can we instead name it scrollRef os something similar?
also is there a reason all of those queries are public? does a consumer really need them?

packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
```diff
@@ -1,180 +1,673 @@
...
+  @query('[part="wrapper"]', true)
+  public $wrapper!: HTMLElement;
+
+  @query('[part="container"]', true)
+  public $container!: HTMLElement;
+
+  /**
+   * The current virtual items being rendered.
+   */
+  get items(): VirtualItem[] {
```
cscheffauer: same as above, does this need to public? please make everything private unless its really necessary for consumers.

packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
```diff
@@ -1,180 +1,673 @@
...
+    return defaultIndexes;
   }
 
-  public override connectedCallback(): void {
-    this.virtualizerController = new VirtualizerController(this, {
-      count: this.virtualizerProps.count!,
-      estimateSize: this.virtualizerProps?.estimateSize!,
-      getScrollElement: () => this.scrollElementRef.value || null,
-      ...this.virtualizerProps,
+  updateHiddenItemsPosition() {
```
cscheffauer: issue: this method has no private nor @internal comment - has to be properly written.

packages/components/src/components/virtualizedlist/virtualizedlist.e2e-test.ts
```diff
@@ -1,52 +1,349 @@
...
+
+      // Check that the virtualized list is present
+      await expect(virtualizedList).toBeVisible();
+
+      // Check that only a subset of items are rendered (virtualization working)
+      const listItems = virtualizedList.locator('mdc-listitem').filter({ visible: true });
+      const itemCount = await listItems.count();
+
+      // Should be less than total count (100) due to virtualization
+      expect(itemCount).toBeLessThan(100);
```
cscheffauer: suggestion: this would be better to be more accurate, like this gives a lot of room for error

cscheffauer (changes requested):

packages/components/src/components/virtualizedlist/virtualizedlist.stories.ts
```diff
@@ -19,19 +24,16 @@ const meta: Meta = {
...
-      about all possible props.`,
-      control: 'object',
-    },
-    setlistdata: {
-      description: `A function that is passed in that when called, will udpate the state of the parent component.
-      This is necessary so that updates inside of virtualizedlist also 
-      rerender the parent with any appropriate updates.`,
-      type: 'function',
-    },
+    ...hideControls([
```
cscheffauer: issue: due to not making certain properties private they show up on storybook for no reason.
1. Cleanup the component itself to make only public what consumers need.
2. If something needs to be public, make sure the control for it is either disabled (if a consumer shouldn't change it) or change it to the proper storybook control

Image

packages/components/src/components/virtualizedlist/virtualizedlist.stories.ts
```diff
@@ -41,6 +43,42 @@ export default meta;
...
+
+export const Dynamic: StoryObj = {
+  render: () => html` <mdc-virtualizeddynamiclist></mdc-virtualizeddynamiclist>`,
+};
+
+export const Chat: StoryObj = {
+  render: () => html` <mdc-virtualizedlist-chat-example></mdc-virtualizedlist-chat-example>`,
+};
+
+export const DynamicContent: StoryObj = {
```
cscheffauer: issue: this example is messed up for me, if u press get next message and u scroll up in the list, it jumps around and messes up the list. There is something wrong which needs fixing.
https://github.com/user-attachments/assets/6e726fda-11bd-4452-9f8a-d7c273bc8c4c

packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
maxinteger: Yes, otherwise can consumer can not access to the virtualised items without listening on the change event

packages/components/src/components/virtualizedlist/virtualizedlist.stories.ts
Coread: this works for me now

packages/components/src/components/virtualizedlist/helpers/virtualizedDynamicList.stories.utils.ts
```diff
@@ -0,0 +1,134 @@
+import { html, TemplateResult } from 'lit';
```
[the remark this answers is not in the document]
Coread: this is fixed

packages/components/src/components/virtualizedlist/virtualizedlist.constants.ts
```diff
@@ -1,13 +1,14 @@
 import utils from '../../utils/tag-name';
 
-const TAG_NAME = utils.constructTagName('virtualizedlist');
-const VIRTUALIZED_WRAPPER_TAG_NAME = utils.constructTagName('virtualizedwrapper');
-
-const DEFAULTS = {
+export const TAG_NAME = utils.constructTagName('virtualizedlist');
+export const VIRTUALIZED_WRAPPER_TAG_NAME = utils.constructTagName('virtualizedwrapper');
+export const IS_AT_BOTTOM_THRESHOLD = 16;
```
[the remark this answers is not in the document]
Coread: this now has a default

Coread (approved):

PR #1557, opened 2025-09-26
fix: tooltip open while hovering on content

supminn:
### Description

* Allow tooltips to be open when the mouse is hovered upon the popover content (moving away the hover from mouse trigger)

MOMENTUM-824

### DIFF
--- a/packages/components/config/storybook/preview.jsx
+++ b/packages/components/config/storybook/preview.jsx
@@ -48,10 +48,18 @@ function refactorCustomElements(customElements) {
           name: `Event Name: "${event.name}"`,
         }));
 
+      const mappedSlots = declaration.slots?.map(slot =>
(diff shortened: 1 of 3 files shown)


packages/components/src/components/popover/popover.component.ts
```diff
@@ -980,7 +980,7 @@ class Popover extends BackdropMixin(PreventScrollMixin(FocusTrapMixin(Component)
       if (middlewareData.arrow && this.arrowElement) {
         this.utils.updateArrowStyle(middlewareData.arrow, placement);
       }
-      if (this.trigger.includes('mouseenter') && this.interactive) {
+      if (this.trigger.includes('mouseenter')) {
```
cscheffauer, line 983: suggestion: we would want a e2e test for the hover bridge

cscheffauer (changes requested):

cscheffauer (approved):

PR #1558, opened 2025-09-26
chore: rm text controls for css properties in all components

arjunkesava:
### Description

- This PR removes text control for all css properties for all components in storybook.
- The problem with current approach is that a css property is defined in component.styles.ts and mentioned on tsdoc as @cssproperty on the component.ts file for each component.
- By adding a text control for each component.stories.ts, we need to update in two places (at tsdoc section and at stories argTypes section), which may not be ideal.
- The goal is to maintain the single source of truth, by placing them on tsdocs.
- With the help of argTypeEnhancers in https://github.com/momentum-design/momentum-design/pull/1544, we can globally set text controls for all cssproperties on all components.

### DIFF
--- a/packages/components/src/components/accordiongroup/accordiongroup.stories.ts
+++ b/packages/components/src/components/accordiongroup/accordiongroup.stories.ts
@@ -8,7 +8,6 @@ import '../avatarbutton';
 import '../chip';
 
 import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
-import { textControls } from '../../../config/storybook/utils';
 
 import { DEFAULTS, SIZE, VARIANT } from './accordiongroup.constants';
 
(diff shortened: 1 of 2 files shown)


packages/components/src/components/accordiongroup/accordiongroup.stories.ts
```diff
@@ -52,7 +52,7 @@ const meta: Meta = {
       control: 'select',
       options: Object.values(VARIANT),
     },
-    ...textControls(['default', '--mdc-accordiongroup-border-color']),
+    ...textControls(['default']),
```
supminn, line 55: why do we still need default here?

supminn (commented):
If we are not making use of `textControls` anymore, could we also get it removed from storybook utils?

PR #1565, opened 2025-09-30
refactor: badge component adjustments

arjunkesava:
#### Problem:
- The current badge component uses parts to set the background and foreground colors for icon types

#### Solution:
- This PR removes the unnessary CSS parts and set the styles with type attribute.
- There are no UI, UX changes on the badge component with this PR.

### BREAKING CHANGE: 
- This PR removes the below css parts from badge component:
```
@csspart badge-icon__error - The error icon badge.
@csspart badge-icon__primary - The primary icon badge.
@csspart badge-icon__secondary - The secondary icon badge.
@csspart badge-icon__success - The success icon badge.
@csspart badge-icon__warning - The warning icon badge.
```


packages/components/src/components/badge/badge.component.ts
```diff
@@ -49,11 +49,6 @@ import type { BadgeType, IconVariant } from './badge.types';
  *
  * @csspart badge-dot - The dot notification badge.
  * @csspart badge-icon - The icon badge.
- * @csspart badge-icon__error - The error icon badge.
```
cscheffauer, line 52: issue: removing css parts here is a breaking change for consumers who are using those. We have to point this out in the PR description!

cscheffauer (changes requested):

packages/components/src/components/badge/badge.component.ts
arjunkesava: Done. added breaking change on PR desc/

PR #1575, opened 2025-10-03
fix: store update error on select default value

arjunkesava:
### Description

- The select component when renders for the first time with a default selected value, then the `onStoreUpdate` would be called where all options are removed from store before they are even added.
- At that time, the `itemStore.items` will be undefined (as no options are added) and trying to get the length of items results in an error.
image

- This PR fixes this issue by checking the truthiness value of `itemStore.items` before removing the state item on onStoreUpdate.

### Links

- MOMENTUM-828

### DIFF
--- a/packages/components/src/components/select/select.component.ts
+++ b/packages/components/src/components/select/select.component.ts
@@ -259,7 +259,7 @@ class Select
         option.setAttribute('tabindex', '-1');
         break;
       case 'removed': {
-        if (index === -1 || option.tabIndex !== 0) {
+        if (!this.navItems || index === -1 || option.tabIndex !== 0) {
           return;
         }
 


packages/components/src/components/select/select.component.ts
```diff
@@ -259,7 +259,7 @@ class Select
         option.setAttribute('tabindex', '-1');
         break;
       case 'removed': {
-        if (index === -1 || option.tabIndex !== 0) {
```
cscheffauer, line 262: question: is there a way we can cover this in a e2e test potentially?

cscheffauer (changes requested):

arjunkesava:
Closing this PR w.r.t to. https://github.com/momentum-design/momentum-design/pull/1577

PR #1576, opened 2025-10-03
refactor: updated css properties on button and button simple component

supminn:
### Description

* This PR simplifies the css properties used on button and buttonsimple component
* Update ts-docs to support the cssproperties. Thus, they are also available on storybook.

### DIFF
--- a/packages/components/src/components/popover/popover.stories.ts
+++ b/packages/components/src/components/popover/popover.stories.ts
@@ -21,6 +21,7 @@ import type Popover from '.';
 import type Dialog from '../dialog';
 
 import { COLOR, DEFAULTS, POPOVER_PLACEMENT } from './popover.constants';
+import { VALID_TEXT_TAGS } from '../text/text.constants';
 
 const createPopover = (args: Args, content: TemplateResult) => html`
   <mdc-popover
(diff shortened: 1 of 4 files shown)


packages/components/src/components/button/button.component.ts
```diff
@@ -50,6 +50,17 @@ import { getIconNameWithoutStyle } from './button.utils';
...
+ * @cssproperty --mdc-button-height - Height for button size
+ * @cssproperty --mdc-button-background-color - Background color of the button
+ * @cssproperty --mdc-button-border-color - Borer color of the button
+ * @cssproperty --mdc-button-text-color - Text color of the button
+ * @cssproperty --mdc-button-prefix-icon-size - Size of the prefix icon
+ * @cssproperty --mdc-button-postfix-icon-size - Size of the postfix icon
+ * @cssproperty --mdc-button-line-height-size-40 - Line height of the button text for size 40
+ * @cssproperty --mdc-button-line-height-size-32 - Line height of the button text for size 32
+ * @cssproperty --mdc-button-line-height-size-28 - Line height of the button text for size 28
+ * @cssproperty --mdc-button-line-height-size-24 - Line height of the button text for size 24
```
maxinteger, line 63: If we can set the button height and the button is always single line, then we do we need line height?

packages/components/src/components/button/button.styles.ts
```diff
@@ -2,44 +2,21 @@ import { css } from 'lit';
...
     --mdc-button-line-height-size-40: var(--mds-font-lineheight-body-large);
     --mdc-button-line-height-size-32: var(--mds-font-lineheight-body-large);
     --mdc-button-line-height-size-28: var(--mds-font-lineheight-body-midsize);
     --mdc-button-line-height-size-24: var(--mds-font-lineheight-body-small);
 
-    --mdc-button-prefix-icon-size: 1rem;
-    --mdc-button-postfix-icon-size: 1rem;
+    background: var(
+      --mdc-button-background-color
+    ); /* Using background instead of background-color since gradients are not supported in background-color */
```
maxinteger, line 14: I would recommend to rename `--mdc-button-background-color` to `--mdc-button-background`. Then you will not need the this comment and the consumer will be not confused by the `color` postfix

maxinteger (changes requested):

PR #1577, opened 2025-10-03
fix(utils): element store handle host disconnection properly

maxinteger:
### Description

Element store do not wait for destroy events and do not call `onStoreUpdate` after the host disconnected

### DIFF
--- a/packages/components/src/utils/controllers/ElementStore.ts
+++ b/packages/components/src/utils/controllers/ElementStore.ts
@@ -6,7 +6,7 @@ import type { Component } from '../../models';
 
 export type ElementStoreChangeTypes = 'added' | 'removed';
 
-interface ElementStoreOptions<TItem extends HTMLElement = HTMLElement> {
+interface ElementStoreOptions<TItem extends HTMLElement> {
   /**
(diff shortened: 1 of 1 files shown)


packages/components/src/utils/controllers/ElementStore.ts
```diff
@@ -117,19 +117,26 @@ export class ElementStore<TItem extends HTMLElement> implements ReactiveControll
...
+  hostConnected() {
     this.host.addEventListener(LIFE_CYCLE_EVENTS.CREATED, this.itemCreationHandler);
     this.host.addEventListener(LIFE_CYCLE_EVENTS.DESTROYED, this.itemDestroyHandler);
   }
 
-  hostConnected() {}
-
-  hostDisconnected() {}
+  hostDisconnected() {
+    this.host.addEventListener(LIFE_CYCLE_EVENTS.CREATED, this.itemCreationHandler);
```
cscheffauer, line 133: i don't understand why this is done? when the host is disconnected event listeners should be cleaned up no? we are adding them here again?

arjunkesava (approved):
LGTM

PR #1609, opened 2025-10-14
refactor: adding unselected error validation state on checkbox

supminn:
### Description

* mdc-checkbox component supports 'help-text-type' with the values 'default' and 'error'.
* The error state can be used while working with form validation along with help-text Validation.
MOMENTUM-827


packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -90,11 +91,8 @@ class Checkbox
    */
   @property({ type: Boolean, reflect: true }) override autofocus = false;
 
-  override connectedCallback(): void {
-    super.connectedCallback();
-    // Checkbox does not contain helpTextType property.
-    this.helpTextType = undefined as unknown as ValidationType;
-  }
+  @property({ type: String, reflect: true, attribute: 'help-text-type' })
```
arjunkesava: Please add a comment explaing that this attribute only supports "error" and "default", as of now it's showing that it will support all types of help text.

PR #1611, opened 2025-10-15
fix(controlType): add controlType option to menuitemradio

lismith2-cisco:
### Description

add controlType option to menuitemradio


packages/components/src/components/menuitemradio/menuitemradio.e2e-test.ts
```diff
@@ -80,34 +119,51 @@ test('mdc-menuitemradio', async ({ componentsPage }) => {
...
       const radio = await setup({ componentsPage, softDisabled: true });
-      await expect(radio).toHaveAttribute('aria-disabled', 'true');
-      await expect(radio).toHaveAttribute('soft-disabled', '');
+      await expectSoftDisabled(radio);
+
+      // Click should change state when soft disabled
+      const changeEventFiredPromiseFunction = await getChangeEventFiredPromiseFunction(componentsPage, radio);
+      await radio.click({ force: true });
+      await expectChangeEventFired(changeEventFiredPromiseFunction);
+      await expectChecked(radio);
```
maxinteger: This is incorrect. Soft-disabled element is the same as disabled except the user can focus on it.

More details in the docs:
https://momentum.design/storybook-static/iframe.html?viewMode=docs&id=attributes--docs&globals=#soft-disabled

maxinteger (changes requested):
Could you please add the `controlled` attr and a short description to the `packages/components/src/docs/Attributes.mdx` docs about what is it and why do we need it?

packages/components/src/components/menuitemradio/menuitemradio.e2e-test.ts
maxinteger: I created the issue to fix it

maxinteger (approved):

PR #1614, opened 2025-10-15
fix(accordion): add aria-label

nataliadelmar:
### Description

Adding aria-label to the accordion component trigger button to make it accessible. Also deleting title prop as it is not accessible.


packages/components/src/components/accordion/accordion.component.ts
```diff
@@ -67,6 +67,18 @@ class Accordion extends AccordionButton {
   @queryAssignedElements({ slot: 'trailing-controls' })
   trailingControlsSlot!: Array<HTMLElement>;
 
+  /**
+   * Aria-label attribute for the trigger button when accordion is collapsed.
+   */
+  @property({ type: String, attribute: 'open-button-aria-label' })
+  openButtonAriaLabel?: string;
```
cscheffauer: suggestion: there are not 2 separate properties necessary, since the this.expanded controls the state of it anyways. Its similar to the `showHideButtonAriaLabel` in password. Check that how its done there.

maxinteger: @nataliadelmar I agree with Christoph, if you use both aria label then VO will announce:
- "Open accordion, collapsed" and
- "Hide accordion, expanded"

It is more confusion then useful

cscheffauer (approved):

PR #1628, opened 2025-10-23
fix: soft-disabled support for listitem, menuitemcheckbox and menuitemradio

2020Deeya:
### Description

This PR fixes an issue where menuitemcheckbox and menuitemradio with the soft-disabled attribute were still clickable. 

#### Fixes
MenuItem (including checkbox and radio variants) now correctly prevent interactions when soft-disabled, while remaining keyboard accessible for focus navigation.


maxinteger (changes requested):
After testing I found a problem. When the menu item (checkbox and radio as well) soft-disabled and I click on it the menu closes. My expectation is nothing happens so the menu does not close either. Let me know if you have different opinion.

PR #1629, opened 2025-10-23
chore: button group styles and documentation

arjunkesava:
### Description

- Add documentation for button group component.
- Update styles to limit the overflow of button boundaries inside the button group.
- Remove 1px from the padding-left to adjust the first button width (to include 1px border left) inside the button group for all supported sizes and in the horizontal orientation and secondary variant. 

### Links

- MOMENTUM-841 | MOMENTUM-750


packages/components/src/components/buttongroup/buttongroup.stories.ts
```diff
@@ -4,9 +4,11 @@ import { html } from 'lit';
 
 import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
 import { disableControls } from '../../../config/storybook/utils';
+import { AnimationNames } from '../animation/animation.types';
```
2020Deeya:
```suggestion
import type { AnimationNames } from '../animation/animation.types';
```

2020Deeya:
What has changes in the snapshots?

arjunkesava:
> What has changes in the snapshots?

There is a slight < 1px difference in the space between the horizontal buttons. This change was happened with the effect of styles.

PR #1641, opened 2025-10-29
feat(sidenav): sidenavigation and related component fixes

2020Deeya:
#### Breaking Changes
- [SideNavigation] - Removed 'parent-nav-tooltip-text' attribute. Instead please set 'is-active-parent-tooltip-text' on individual navmenuitems which act as a parent in a nested sidenav. This allows for more fine grained control on the tooltip text.
- [SideNavigation] - Changed default Expanded Width of SideNavigation from 240px to 232px (use css properties to override)

#### New Features

- [SideNavigation] - Added a new variant 'flexible-on-hover' to enable grabber visibility on hover or keyboard focus

#### Functional Fixes
- [SideNavigation] - Added new property 'hideFixedSectionDivider' to control visibility of section divider.
- [NavMenuItem] - Changed icon switch mechanism (between icon-name and filled if toggling active) for better performance.
- [NavMenuItem] - Added new properties 'cannot-activate', 'is-active-parent-tooltip-text' and various tooltip properties for more flexibility.
- [Menubar] - Menubar component was unnecessarily wrapping children in role group, removed the group role there.
- [Popover] - Fixed an issue where attached Popovers / Tooltips to Link, ButtonLink, etc where not opening on focus. 

#### Styling Fixes
- [SideNavigation] - If footer-text is not provided, styling is adjusted to have less bottom padding.
- [SideNavigation] - Various margin and padding changes to follow latest design specs.
- [SideNavigation] - Fixed an issue where trailing arrow icon was hidden behind scrollbar in nested side navigation case.
- [SideNavigation] - Fixed an issue where horizontal scrollbar was shown in scrollable section.
- [MenuItem] - Fixed the trailing and leading arrow icon size to follow latest design specs, in a case where submenus are used.
- [MenuItem] - Fixed the trailing and leading arrow icon size to follow latest design specs, in a case where submenus are used.


packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -132,8 +133,87 @@ class SideNavigation extends Provider<SideNavigationContext> {
...
+    this.isHovered = false;
+    if (!this.isFocused) {
+      this.hideGrabberButton();
+    }
+  }
+
+  /** @internal */
+  private handleFocusIn(e: FocusEvent): void {
+    // Only set focus if the sidenavigation itself or a focus-visible element inside it is focused
+    if (e.target === this || (e.target as HTMLElement)?.matches(':focus-visible')) {
```
maxinteger: the comment is not matching with the code. The comment says the focus-visible element is inside the sidenavigation element:


```suggestion
    if (e.target === this || (this.contains(e.target) && e.target as HTMLElement)?.matches(':focus-visible')) {
```

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -132,8 +133,87 @@ class SideNavigation extends Provider<SideNavigationContext> {
...
+  /**
+   * Removes event listeners for flexible-on-hover variant.
+   * Safe to call regardless of current variant.
+   * @internal
+   */
+  private removeFlexibleOnHoverListeners = (): void => {
+    this.removeEventListener('mouseenter', this.handleMouseEnter.bind(this));
+    this.removeEventListener('mouseleave', this.handleMouseLeave.bind(this));
+    this.removeEventListener('focusin', this.handleFocusIn.bind(this));
+    this.removeEventListener('focusout', this.handleFocusOut.bind(this));
```
maxinteger: Unfortunately this will not remove any listeners and leads to memory leak.

It happens because `bind` returns with new function every time when you call it, even if you bind to the same object:

```js
function fn () { return this }

fn.bind(window) === fn.bind(window) // false
```

You can solve this multiple ways:
1. use arrow function for the handlers like: `private handleMouseEnter = (e) => {}`
2. store the binded function and use that for add/remove listener: 
   ```js
      private handleMouseEnter() {...}
      private handleMouseEnterBind = handleMouseEnter.bind(this)
   ```
   
Usually we do the first option

packages/components/src/components/sidenavigation/sidenavigation.styles.ts
```diff
@@ -70,6 +71,16 @@ const styles = css`
...
     z-index: var(--mdc-sidenavigation-vertical-divider-button-z-index);
   }
+
+  :host([variant='flexible-on-hover'])::part(vertical-divider),
+  :host([variant='flexible-on-hover'])::part(vertical-divider-button) {
+    opacity: 0;
+  }
+
+  :host([data-grabber-visible][variant='flexible-on-hover'])::part(vertical-divider),
+  :host([data-grabber-visible][variant='flexible-on-hover'])::part(vertical-divider-button) {
```
maxinteger: it is really just nitpicking but for better readability you should order the selectors from the more generic to the more specific


```suggestion
  :host([variant='flexible-on-hover'][data-grabber-visible])::part(vertical-divider),
  :host([variant='flexible-on-hover'][data-grabber-visible])::part(vertical-divider-button) {
```

packages/components/src/components/navmenuitem/navmenuitem.styles.ts
```diff
@@ -138,18 +153,30 @@ const styles = [
...
+      right: -0.25rem;
+      top: 0.875rem;
+    }
+
+    :host(:dir(rtl))::part(arrow) {
+      position: absolute;
+      left: -0.25rem;
+      top: 0.875rem;
+    }
+
```
maxinteger: Please use logical values instead:

```suggestion
    :host::part(arrow) {
      position: absolute;
      inset-inline-end: -0.25rem;
      top: 0.875rem;
    }
```

packages/components/src/components/navmenuitem/navmenuitem.component.ts
```diff
@@ -60,16 +61,35 @@ import styles from './navmenuitem.styles';
  * @csspart text-container - The container of the text.
  * @csspart trailing-arrow - The trailing arrow of the navmenuitem.
  */
-class NavMenuItem extends IconNameMixin(MenuItem) {
+class NavMenuItem extends MenuItem {
   /**
    * The navmenuitem's active state indicates whether it is currently toggled on (active) or off (inactive).
-   * When the active state is true, the navmenuitem is considered to be in an active state, meaning it is toggled on.
-   * Conversely, when the active state is false, the navmenuitem is in an inactive state, indicating it is toggled off.
+   * If cannotActivate is set to true, do not use this property.
```
maxinteger: This line is confusing. What happen if I use `active` when `cannotActivate` is `true`? If it cause undefined behaviour then it should be fixed. If the value ignored then the docs should say: "This property ignore when `cannotActivate` set to `true`."

packages/components/src/components/navmenuitem/navmenuitem.component.ts
```diff
@@ -165,48 +225,55 @@ class NavMenuItem extends IconNameMixin(MenuItem) {
...
+      changedProperties.has('tooltipText') ||
+      changedProperties.has('showLabel') ||
+      changedProperties.has('hasActiveChild')
+    ) {
+      this.renderDynamicTooltip();
+    }
+
+    if (changedProperties.has('showLabel')) {
+      // If collapsed and aria-label is not set, use visible label
+      if (!this.showLabel && !this.getAttribute('aria-label')?.length && this.label) {
```
maxinteger: `?.length` is not necessary. both `null` and empty string `true` when you negate it.

maxinteger (changes requested):
This PR contains way to many changes and some are completely independent from each other. Next time please create separate PRs.

I added few minor comments as well

packages/components/src/components/navmenuitem/navmenuitem.styles.ts
```diff
@@ -103,13 +125,23 @@ const styles = [
...
+      left: calc(-1 * var(--mdc-navmenuitem-in-sidenav-collapsed-margin-left));
       border-radius: 0 0.25rem 0.25rem 0;
     }
 
-    :host(:dir(rtl))::before {
-      right: -1rem;
+    :host(:not([show-label]):dir(rtl))::before {
+      right: calc(-1 * var(--mdc-navmenuitem-in-sidenav-collapsed-margin-left));
       border-radius: 0.25rem 0 0 0.25rem;
     }
```
maxinteger:
```suggestion
    :host(:not([show-label]):dir(ltr))::before {
      inset-inline-start: calc(-1 * var(--mdc-navmenuitem-in-sidenav-collapsed-margin-left));
      border-radius: 0;
      border-start-end-radius: 0.25rem;
      border-end-end-radius: 0.25rem;
   }
```

packages/components/src/components/sidenavigation/sidenavigation.styles.ts
```diff
@@ -90,12 +97,20 @@ const styles = css`
...
+    padding: 0.5rem calc(var(--mdc-sidenavigation-expanded-left-padding) + 0.5rem) 0.5rem 0rem;
+  }
+
+  :host(:not([expanded]):dir(ltr))::part(brand-logo-container) {
+    padding: 0.5rem 0rem 0.5rem calc(var(--mdc-sidenavigation-collapsed-left-padding) + 0.5rem);
+  }
+
+  :host(:not([expanded]):dir(rtl))::part(brand-logo-container) {
+    padding: 0.5rem calc(var(--mdc-sidenavigation-collapsed-left-padding) + 0.5rem) 0.5rem 0rem;
   }
```
maxinteger:
```suggestion
  :host(:not([expanded]))::part(brand-logo-container) {
    padding-block: 0.5rem;
    padding-inline: 0rem calc(var(--mdc-sidenavigation-collapsed-left-padding) + 0.5rem);
 }
```

PR #1642, opened 2025-10-29
chore: add new virtualized story for listbox fixed height

arjunkesava:
### Description

- Update listbox story with virtualized list of 1000 options.

### Links

- MOMENTUM-863


cscheffauer (changes requested):
disagree with this, putting 1000 items in the DOM should not max out the CPU to 100% for sure.... please investigate properly

packages/components/src/utils/controllers/ElementStore.ts
```diff
@@ -180,6 +180,33 @@ export class ElementStore<TItem extends HTMLElement> implements ReactiveControll
...
   }
 
+  /**
+   * Implements binary search on the cache, to get the index of the nearest newItem position.
+   *
+   * @param cache - The array which contains the list of dom node elements.
+   * @param newItem - The new node element which is needed to be added in the list item.
+   * @returns
+   *  - The index of the existing element which is the nearest one to the new item.
+   *  - If the index is -1, then the item is added at the end of cache.
```
maxinteger: this is not accurate

```suggestion
   * @returns either:
   *  - the index where the new item need to be inserted or
   *  - -1 when the item must be appended to the end of the list
```

packages/components/src/utils/controllers/ElementStore.ts
```diff
@@ -180,6 +180,33 @@ export class ElementStore<TItem extends HTMLElement> implements ReactiveControll
...
+   */
+  private getIndexToInsertInCache(newItem: TItem): number {
+    // If the new item is located at the last place, then we can check the last -1 cache item and return if valid
+    if (this.cache.length && !isAfter(this.cache.at(-1)!, newItem)) {
```
maxinteger: we can save few more instructions if:
```suggestion
    if (!this.cache.length || !isAfter(this.cache.at(-1)!, newItem)) {
```

maxinteger (dismissed):

PR #1644, opened 2025-10-29
fix(virtualizedlist): miscellaneous bug fixes

dominiccarrington:
### Description

**Bug Fixes:**
- When using `revert-list` sync the UI when the list is resized to keep the listitems at the bottom of the new height
- Fix the positioning of popovers when a popover is defined inside of the listitem

**E2E test updates:**
- Add visual regression test for the list header slot
- Fix potential flake in webkit e2e tests

### DIFF
--- a/packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
+++ b/packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
@@ -293,14 +293,12 @@ class VirtualizedList extends DataAriaLabelMixin(List) {
    * Resize observer to monitor size changes of the component.
    * @internal
    */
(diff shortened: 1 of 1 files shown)


packages/components/src/components/virtualizedlist/virtualizedlist.component.ts
```diff
@@ -323,6 +333,8 @@ class VirtualizedList extends DataAriaLabelMixin(List) {
     this.clearScrollToBottomTimer();
     this.virtualizerController = null;
     this.virtualizer = null;
+
+    this.resizeObserver.disconnect();
```
cscheffauer: question: do we also need to remove the object? like this.resizeObserver = null; ?

PR #1645, opened 2025-10-29
fix(dialog): set max-height

maxinteger:
### Description

set the maximum height of the dialog and allow scrolling inside it to prevent content clipping

### DIFF
--- a/packages/components/src/components/dialog/dialog.component.ts
+++ b/packages/components/src/components/dialog/dialog.component.ts
@@ -34,9 +34,6 @@ import styles from './dialog.styles';
  * Use the `onClose` event to handle the close action of the dialog (fired when Close button is clicked
  * or Escape is pressed).
  *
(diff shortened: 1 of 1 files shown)


packages/components/src/components/dialog/dialog.styles.ts
```diff
@@ -76,12 +76,15 @@ const styles = css`
...
     border: none;
     border-radius: 0;
+    max-height: 100vh;
   }
 
   :host {
     width: var(--mdc-dialog-width);
     height: var(--mdc-dialog-height);
     max-width: 100%;
+    max-height: calc(100vh - 2rem);
```
cscheffauer: issue: this max height calculation will be broken if the `--mdc-dialog-padding` css property will be overridden. I would recommend to adjust the calc to be agnostic to that instead of hard coding the 2rem.

PR #1647, opened 2025-10-30
fix(textarea): width issues in Textarea component

2020Deeya:
### Description

This PR fixes width issues in Textarea component.

#### Fixes:
- Textarea by default now properly fills the full width of its container instead of being constrained to a fixed width.
- The default cols attribute is no longer set, allowing the component to adapt responsively to its container width.

#### Breaking change:
- The removal of the default cols value may impact layouts that previously relied on it for fixed sizing. Consumers can still override the width by providing a custom value if needed.


maxinteger:
@2020Deeya this should be 2 separated PR one for the stepper and another one for the textarea

2020Deeya:
> @2020Deeya this should be 2 separated PR one for the stepper and another one for the textarea

Sure, we could create separate PRs. PR title and description have been updated. Let's ignore the branch name.

PR #1653, opened 2025-11-03
chore(textarea): Add custom resize button with keyboard and mouse support

2020Deeya:
### Description
This PR adds a custom resize button to `mdc-textarea` that allows users to adjust mdc-textarea height using keyboard (Arrow Up/Down) or mouse drag.

#### Changes
- Focusable resize button appears inside the textarea when `resizable` is enabled.
- Keyboard resize support:
  - Press Arrow Up to decrease height by 1 row.
  - Press Arrow Down to increase height by 1 row.
- Mouse resize support: Drag the resize button to adjust height.
- The resize button is non-focusable and non-interactable in readonly and disabled states.
- RTL support.

#### Breaking Changes
None - fully backward compatible


packages/components/src/components/textarea/textarea.component.ts
```diff
@@ -410,6 +488,20 @@ class Textarea extends AutoFocusOnMountMixin(FormInternalsMixin(DataAriaLabelMix
           announcement="${ifDefined(this.ariaLiveAnnouncer)}"
           data-aria-live="polite"
         ></mdc-screenreaderannouncer>
+        ${
+          !this.disabled &&
+          !this.readonly &&
```
maxinteger: I think the main purpose of the resize button is to make the content more readable, so we should allow it in both disabled and readonly modes as well.

The native textaria can be resized when it is disabled or readonly.

packages/components/src/components/textarea/textarea.styles.ts
```diff
@@ -103,6 +104,27 @@ const styles = [
       opacity: 0;
       pointer-events: none;
     }
+
+    .resize-button {
+      position: absolute;
+      bottom: 0.25rem;
+      right: 0.25rem;
```
maxinteger: please use `inset-inline-end` instead, then you can remove `left: 0.25rem; right: unset;` from the ` :host(:dir(rtl)) .resize-button`

maxinteger (dismissed):

PR #1663, opened 2025-11-07
fix(textarea): Fix VoiceOver announcement issue

2020Deeya:
### Description
This PR Fixes two accessibility issues in the `mdc-textarea` component when using VoiceOver with character limit announcements.

- Interrupted Character Limit Announcements(MOMENTUM-873): When typing, the character limit announcement would be interrupted by VoiceOver reading the last inserted word, preventing users from hearing the full character count message.
- Character count announcements for mdc-textarea were not being read by VoiceOver when the textarea was placed inside mdc-dialog.

#### Fixes
- Fixed Interrupted Character Limit Announcements
- Fixed VoiceOver announcements for textarea in modal dialogs

### DIFF
--- a/packages/components/src/components/announcementdialog/announcementdialog.component.ts
+++ b/packages/components/src/components/announcementdialog/announcementdialog.component.ts
@@ -90,7 +90,21 @@ class AnnouncementDialog extends Dialog {
    * @default medium
    */
   @property({ type: String, reflect: true })
-  override size: AnnouncementDialogSize = DEFAULTS.SIZE;
+  override get size(): AnnouncementDialogSize {
+    return this.responsiveSettingsContext?.value?.forceFullscreenDialog
+      ? 'fullscreen'
+      : (this.internalSize as AnnouncementDialogSize);
+  }
+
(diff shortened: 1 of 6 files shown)


arjunkesava (changes requested):
I've tested this PR locally, and noticed the the textarea is not announcing it on the dialog is sitll not working. MOMENTUM-872.

packages/components/src/components/textarea/textarea.component.ts
```diff
@@ -527,6 +536,7 @@ class Textarea extends AutoFocusOnMountMixin(FormInternalsMixin(DataAriaLabelMix
           identity="${this.inputId}"
           announcement="${ifDefined(this.ariaLiveAnnouncer)}"
           data-aria-live="polite"
+          debounce-time="1500"
```
supminn:
```suggestion

```

2020Deeya:
Note: The issue related to duplicate announcements — where typed text is announced twice (e.g., “hello”) — isn’t in our control, as it also occurs in the native textarea.

PR #1668, opened 2025-11-10
fix: popover backdrop mixin zindex layout issue

arjunkesava:
### Description

- [Popover] This PR will fix the popover default backdrop z-index layout issue.
- There are no UI/UX changes with this PR.

### Breaking Changes:
- The backdrop element zindex logic is updated now. Here is an example:
```html
<mdc-button id="trigger-element"></mdc-button> // <-- the trigger element zindex will now be 1002.
<mdc-popover triggerid="trigger-element"><mdc-popover> // <-- the popover zindex will be 1003 (default one)
<div class="popover-backdrop></div> // <-- the backdrop layer with the zindex value of 1001.
```
- With this change, the trigger element will be clickable and the popover will be set on top of it.
- The nested popover will effectively have a new z-index of multiples of 3 which is 1003, 1006, 1009, 1012 etc.

### Links

- MOMENTUM-767


arjunkesava:
When any popover with the backdrop attribute enabled, the popover trigger element will be set with the popover zindex (i.e. 1000), due to which when an internal popover is open, we can still see the parent popover trigger element button.

https://github.com/user-attachments/assets/ba578b97-268e-4f32-8c74-afdb1b2335e9

An easy hack could be to the place the trigger element on the top of the popover always, so that in the DOM order, the trigger element is painted first on the screen and it will not be visible when the children popover are open.

But ideally, we should not depend on DOM order as the consumer may place the trigger element on top or bottom.

A possible fix is to set the default backdrop mixin to the popover zindex - 1 (to set it to the back of the parent popover) so that when children are open then the trigger will stay on the back of it.

cscheffauer (changes requested):

packages/components/src/utils/mixins/BackdropMixin.ts
```diff
@@ -120,7 +120,7 @@ export const BackdropMixin = <T extends Constructor<LitElement>>(superClass: T)
       };
 
       // Set the z-index and position to ensure the element is above the backdrop
-      element.style.zIndex = `${this.zIndex}`;
+      element.style.zIndex = `${this.zIndex - 1}`;
```
cscheffauer: so the backdrop is -2 and the element is -1? why is that necessary?

packages/components/src/components/popover/popover.component.ts
```diff
@@ -441,6 +450,12 @@ class Popover extends BackdropMixin(PreventScrollMixin(FocusTrapMixin(Component)
   /** @internal */
   protected shouldSuppressOpening: boolean = false;
 
+  /** @internal */
+  private internalZIndex?: number = DEFAULTS.Z_INDEX;
```
maxinteger: this should be `undefined` by default otherwise the `zIndex` calculation (line will never happen

maxinteger (approved):

PR #1673, opened 2025-11-11
fix: Judge the background color of the buttons below the dialog box

zekyu11:
### Description

This PR fixes an issue where the FooterMixin was overriding the color attribute of footer buttons even when a color was already set.  The logic now respects existing color attributes and only applies the default or promotional colors when none are defined.


maxinteger (changes requested):
@zekyu11 Please fill the description section of the PR, also remove non-public links

PR #1877, opened 2026-04-08
chore(tab): add size 28 tab

NotNestor:
### Description

Adds explicit public `size` support to `mdc-tab`, aligned with the existing button-style API, while limiting valid values to `32` and `28`. The new compact size only changes the tab height; typography and the rest of the tab styling remain unchanged. This also updates the tab and tablist Storybook examples to expose the size option and adds tab e2e coverage for supported sizes plus invalid-size fallback behavior.

### DIFF
--- a/packages/components/src/components/tab/tab.e2e-test.ts
+++ b/packages/components/src/components/tab/tab.e2e-test.ts
@@ -4,8 +4,8 @@ import { ComponentsPage, test } from '../../../config/playwright/setup';
 import StickerSheet from '../../../config/playwright/setup/utils/Stickersheet';
 import type { IconNames } from '../icon/icon.types';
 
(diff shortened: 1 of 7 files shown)


packages/components/src/components/tab/tab.e2e-test.ts
```diff
@@ -140,15 +143,30 @@ test('mdc-tab', async ({ componentsPage }) => {
         await componentsPage.setAttributes(tab, { tabindex: '0' });
       });
 
+      // size
+      await Object.values(TAB_SIZES).reduce(async (previous, tabSize) => {
```
cscheffauer, line 147: suggestion: we need to add the new size also to the tab screenshots to confirm visual integrity in the future.

cscheffauer (approved):
approved - please add the new tabsize to e2e screenshot

PR #1880, opened 2026-04-09
feat: new icon (call-accept)

clairelikesemojis:
New Icon added: 
- Call-accept

No visual updates:
- Block-quote (light)
- Indent-decrease


clairelikesemojis:
Will re-push

PR #1882, opened 2026-04-09
fix(fonts): apply ss02 only to the Momentum font-face

NotNestor:
### Description

Moves the `ss02` font feature setting off the `mdc-themeprovider` host styles and onto the `Momentum` `@font-face` definition instead.

This ensures the stylistic set is only applied when the Inter-backed `Momentum` font is actually used, and does not affect fallback fonts such as Arial, where the setting produces poor results.

Also removes the now-stale `themeprovider` css property documentation for `--mdc-themeprovider-font-feature-settings`.

Consumers should bump both package versions at the same time when taking this change:
`@momentum-design/fonts` and `@momentum-design/components`.
These changes need to ship together so the font-face update and the theme provider update stay in sync.

### Links

N/A


packages/components/package.json
```diff
@@ -59,7 +59,7 @@
     "copy:icons": "node scripts/copyIcons.js",
     "copy:illustrations": "node scripts/copyIllustrations.js",
     "copy:brandvisuals": "node scripts/copyBrandvisuals.js",
-    "copy:fonts": "node scripts/copyFonts.js",
```
cscheffauer, line 62: @NotNestor why is this line necessary? in CI thats already done and locally we are working in a way where scripts affecting other packages are executed manually.

cscheffauer (changes requested):

packages/components/src/components/calendar/__screenshots__/firefox/mdc-calendar-minmax-ltr.png
supminn, line 1: The 0 in `2025` looks different on this snapshot

supminn: Looks like this is an issue on Firefox

supminn (commented):
The font update is slightly different on Firefox snapshots. Is this as expected or something that we'll need to fix?

packages/components/src/components/calendar/__screenshots__/firefox/mdc-calendar-minmax-ltr.png
NotNestor: the updated is correct the 0 with the line through it means ss02 is turned on. it is for disambiguity so that the letter O and the number 0 do not get confused

supminn (approved):

PR #1888, opened 2026-04-20
fix: flakiness in visual regression tests

2020Deeya:
### Description

This PR fixes flakiness in visual regression tests caused by icon loading race condition. [MOMENTUM-1029]


packages/components/src/components/input/input.e2e-test.ts
```diff
@@ -86,6 +88,10 @@ const setup = async (args: SetupOptions, isForm = false) => {
 
 test.use({ viewport: { width: 800, height: 1500 } });
 test('mdc-input', async ({ componentsPage, browserName }) => {
+  // Many page remounts across attributes, interactions, form validation, and visual regression
+  // can exceed the default 30s timeout in slower browsers (webkit, msedge, firefox)
+  test.setTimeout(60000);
```
maxinteger: If we add more test or run a slower machine then it can cause flakiness again.

Can we fix this other way? For example add explicit `waitFor` calls after the mount, or split the test case so the 30s limit applied for both.

maxinteger (changes requested):

maxinteger (approved):

PR #1892, opened 2026-04-24
chore(tokens): add core motion tokens 

2020Deeya:
### Description

This PR adds the Core motion tokens to the design token pipeline of the Motion Provider architecture (Core → Animation → Component).


packages/tools/token-builder/src/schemas/json/core/motion.json
maxinteger, line 1: I think this schema is too restrictive.

Basically this schema does not allow any modification on the core token json.

We should allow any category name (if they are valid identifiers, contains only letters and numbers and starts with letter)

Also we do not need to restrict which type of value can goes into which category.
basically any type can goes into any category.

With the committed motion.json we can give a baseline to the user, but then it is up to them how they change it.

maxinteger (changes requested):

packages/tools/token-builder/src/schemas/json/core/motion.json
```diff
@@ -0,0 +1,61 @@
...
+        },
+        "description": {
+          "type": "string"
+        }
+      },
+      "required": [
+        "value",
+        "type"
+      ],
+      "additionalProperties": false,
```
2020Deeya, line 56: `additionalProperties: false` on MotionToken means designers cannot add extra metadata fields (like deprecated, alias, group).

maxinteger: yes, this is the correct behaviour

packages/tools/token-builder/src/schemas/json/core/motion.json
```diff
@@ -0,0 +1,61 @@
...
+    "MotionToken": {
+      "type": "object",
+      "properties": {
+        "value": {
+          "type": "string",
+          "minLength": 1
+        },
+        "type": {
+          "type": "string",
+          "minLength": 1
```
maxinteger: We have a well defined list of values for the `type` field, let's use that:

```suggestion
          "type": "string",
          "enum": ["duration", "bazierCurve"]
```

packages/tools/token-builder/src/schemas/json/core/motion.json
```diff
@@ -0,0 +1,61 @@
...
+      },
+      "additionalProperties": false,
+      "title": "TokenCategory"
+    },
+    "MotionToken": {
+      "type": "object",
+      "properties": {
+        "value": {
+          "type": "string",
+          "minLength": 1
```
maxinteger: the actual format of the value is depend on the `type`
- for duration it should match "100ms"
- for bazierCurve it should match "cubic-bezier(0..44, 0, 0, 1)"

You can do [condition validation](https://json-schema.org/understanding-json-schema/reference/conditionals#ifthenelse) in JSON schema.
You can ask AI to fix it up for you

maxinteger (changes requested):

maxinteger (approved):

supminn (approved):

PR #1893, opened 2026-04-27
refactor: side navigation with dropdown support in expanded mode

supminn:
### Description

* Side navigation component supports a dropdown type variant in the expanded mode. 
* This capability has been added through the PR.
* It follows the same keyboard interactions as the flyout menu approach, with the only visual difference being that it is presented as an accordion-style dropdown.
* Refactoring `Provider` component from an abstract class to a regular class - This change is to support SideNavigation class to be able to consume the `KeyToActionMixin` along with having a Provider and Context.
* Introduced new css variable ` --mdc-navmenuitem-notch-color` to update the active navmenuitem's notch colour.

### Snapshot updates
* Navmenuitem snapshots have an addition of navmenuitem with label only and no icon.
* Sidenavigation snapshots have an addition of accordion style dropdown in expanded mode
* Exisiting Sidenavigation snapshots have been updated for children navmenuitems not to have icons but label only. This is a recommendation from momentum-design team. It has been added to the documentation as well.

### Breaking Change

* None.
* This is an add-on feature and doesn't hamper any of the existing functionality of the side navigation component.


packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -323,6 +377,191 @@ class SideNavigation extends Provider<SideNavigationContext> {
...
 
+  /**
+   * Closes all open dropdown containers and resets their parent navmenuitems.
+   * @internal
+   */
+  private closeAllDropdowns(): void {
+    const dropdownParents = this.navMenuItems.filter(item => this.context.value.isDropDownParent(item));
+    dropdownParents.forEach(item => {
+      item.closeDropdown();
+    });
```
maxinteger: nitpicking:

```suggestion
    this.navMenuItems
       .filter(item => this.context.value.isDropDownParent(item))
       .forEach(item => item.closeDropdown());
```

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -323,6 +377,191 @@ class SideNavigation extends Provider<SideNavigationContext> {
...
+   * @internal
+   */
+  private hideAllDropdownContainers(): void {
+    const containers = this.querySelectorAll<HTMLElement>('div[data-trigger]');
+    containers.forEach(container => {
+      Object.assign(container.style, {
+        display: 'none',
+        flexDirection: 'column',
+        gap: '0.25rem',
+      });
```
maxinteger, line 363: it would be nice to use class or data attribute to style the hidden dropdown container

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -323,6 +377,191 @@ class SideNavigation extends Provider<SideNavigationContext> {
     return Array.from(this.querySelectorAll(`${NAVMENUITEM_TAGNAME}:not([disabled])`));
   }
 
+  /**
+   * Closes all open dropdown containers and resets their parent navmenuitems.
+   * @internal
+   */
+  private closeAllDropdowns(): void {
+    const dropdownParents = this.navMenuItems.filter(item => this.context.value.isDropDownParent(item));
```
maxinteger: Other places you use the `div[data-trigger]` selector, why not here?
```ts
this.querySelectorAll<HTMLElement>('div[data-trigger]').map(d => d.parentElement)
```

packages/components/src/components/navmenuitem/navmenuitem.component.ts
```diff
@@ -273,12 +285,42 @@ class NavMenuItem extends MenuItem {
...
   private renderDynamicTooltip(): void {
     this.removeTooltip();
-
+    if (this.disabled) {
+      return;
+    }
+    // Do not show the tooltip when the dropdown is open and the navmenuitem has active child.
+    if (this.dropdownOpen) {
+      return;
+    }
```
maxinteger: 1. the 2 `if` can be merged into one
2. the comment above the second `if` is not in sync with the if's condition

packages/components/src/components/navmenuitem/navmenuitem.component.ts
```diff
@@ -337,6 +379,21 @@ class NavMenuItem extends MenuItem {
...
   }
 
+  /**
+   * Check whether the navmenuitem is inside a dropdown container (div[data-trigger]).
+   * @internal
+   */
+  private isInsideDropdownContainer(): boolean {
+    let parent = this.parentElement;
+    while (parent) {
+      if (parent.tagName.toLowerCase() === 'div' && parent.hasAttribute('data-trigger')) {
```
maxinteger: you can use `matches` method, it is more readable


```suggestion
      if (parent.matches('div[data-trigger]')) {
```

maxinteger (changes requested):
Found few issues during manual testing:

1. When I press down when the focus is on the last item in the dropdown, the focus does not move to the next side nav menu item
2. Visually hard to see which items part of the dropdown and which are the main items: Image
3. Voice Over incorrectly announce the group for the dropdown: Image
4. I guess it was a UX decision, but in dropdown mode it would be nice to put the small Image marker as well next to the selected item. The bold icon and text is not enough highlight.
5. In collapsed mode when you press Right on a item with submenu, the menu opens, but when you press Right again the focus jumps to the next main item and the previous menu does not close Image

packages/components/src/components/sidenavigation/sidenavigation.component.ts
supminn: The `container` is placed within the slots in the light DOM, and we have to add styling through the component's shadow DOM.
For this reason, I have appended the styles in this manner.

The alternative approach copilot suggests is by adding injected styles on the DOM.

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -323,6 +378,208 @@ class SideNavigation extends Provider<SideNavigationContext> {
...
+        const currentIndex = children.indexOf(targetItem);
+        if (currentIndex === -1) return;
+
+        if (event.key === KEYS.ARROW_DOWN) {
+          const nextIndex = currentIndex + 1;
+          if (nextIndex < children.length) {
+            event.preventDefault();
+            children[nextIndex].focus();
+          }
+          // At last child: do nothing, focus stays on the last item
```
2020Deeya: issue: The logic here explicitly keeps focus on the first child when `currentIndex === 0` and and the e2e test also asserts the same.

packages/components/src/components/sidenavigation/sidenavigation.context.ts
```diff
@@ -61,10 +81,19 @@ class SideNavigationContext {
     return parents;
   }
 
+  private getTriggerElement(container: Element, attribute: string): NavMenuItem | null {
+    const triggerId = container.getAttribute(attribute);
+    if (!triggerId) return null;
+    return document.getElementById(triggerId) as NavMenuItem | null;
```
2020Deeya, line 89: issue: `document.getElementById` searches the entire document. If two instances of `mdc-sidenavigation `exist on the same page, this will silently return the wrong element. This should be scoped to the component.
```suggestion
    return this.hostElement?.querySelector(`#${CSS.escape(triggerId)}`) as NavMenuItem | null;
```

packages/components/src/components/menupopover/menupopover.component.ts
```diff
@@ -429,7 +429,7 @@ class MenuPopover extends Popover {
       case ACTIONS.RIGHT: {
         // If there is a submenu, open it.
         const subMenu = this.getSubMenuPopoverOfTarget(target);
-        if (subMenu) {
+        if (subMenu && !target.hasAttribute('soft-disabled')) {
```
2020Deeya, line 400:
```suggestion
       if (subMenu && !target.hasAttribute('disabled') && !target.hasAttribute('soft-disabled')) {
```

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -323,6 +378,208 @@ class SideNavigation extends Provider<SideNavigationContext> {
...
+      const triggerItem = this.querySelector(`${NAVMENUITEM_TAGNAME}#${CSS.escape(triggerId)}`) as NavMenuItem | null;
+      if (triggerItem) {
+        triggerItem.closeDropdown();
+        triggerItem.focus();
+        event.stopPropagation();
+      }
+      return;
+    }
+
+    if (event.key === KEYS.TAB) {
```
2020Deeya: question: Why do we need to override the Tab key behavior? As per my understanding on https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/, Tab should follow natural DOM order, not be overridden.

suggestion: Consider removing the Tab override and relying on natural DOM ordering (placing div[data-trigger] immediately after its trigger in the DOM, as is already the case).

packages/components/src/components/navmenuitem/navmenuitem.component.ts
```diff
@@ -351,10 +404,58 @@ class NavMenuItem extends MenuItem {
...
+    } else {
+      this.removeAttribute('aria-expanded');
+    }
+  }
+
+  /**
+   * Closes the dropdown if it is open.
+   * @internal
+   */
+  public closeDropdown(): void {
```
2020Deeya, line 448: issue: public + @internal is contradictory. If it must be callable from SideNavigation, it's effectively part of the internal API between these two sibling components. Consider making it protected with a @internal note, or firing a custom event instead.

packages/components/src/components/sidenavigation/sidenavigation.e2e-test.ts
```diff
@@ -523,3 +523,325 @@ test.describe.parallel('SideNavigation (Nested, all scenarios, all variants)', (
...
+
+      // Children should be back in the div[data-trigger]
+      const childrenInContainer = dropdownContainer1.locator('mdc-navmenuitem');
+      await expect(childrenInContainer).toHaveCount(3);
+      const trigger = sidenav.locator('mdc-navmenuitem#dropdown-trigger-1');
+      await trigger.click();
+      await expect(dropdownContainer1).toHaveCSS('display', 'flex');
+      await expect(dropdownContainer1.locator('mdc-navmenuitem').first()).toBeVisible();
+
+      await componentsPage.visualRegression.takeScreenshot('sidenavigation-dropdown-re-expanded');
```
2020Deeya, line 866: suggestion: There is no checkForA11yViolations call here. The re-expanded state should also verify no accessibility violations.

packages/components/src/components/menupopover/menupopover.component.ts
supminn: When the navmenuitem is disabled, it does not allow keyboard focus. However, when soft-disabled is set to true, keyboard focus is possible. Therefore, we only need to check for soft-disabled in this scenario.

packages/components/src/components/sidenavigation/sidenavigation.context.ts
supminn: Since the `id` needs to be unique to a particular element in the entire document, this shouldn't pose a problem.

packages/components/src/models/provider/provider.component.ts
```diff
@@ -65,7 +65,7 @@ abstract class Provider<C> extends Component {
    * re-render. If the `render()` method is overwritten, this call must be made
    * manually.
    */
-  protected abstract updateContext(): void;
+  protected updateContext(): void {}
```
maxinteger: this should throw an "Not implemented" error

packages/components/src/components/menubar/menubar.component.ts
```diff
@@ -278,9 +272,9 @@ class MenuBar extends KeyDownHandledMixin(KeyToActionMixin(Component)) {
   }
 
   private async closeAllMenuPopovers() {
-    const popovers = this.depthManager.popUntil(item => this.contains(item));
-
-    await Promise.all(popovers.map(popover => popover.updateComplete));
+    const openPopovers = this.getAllPopovers().filter(p => (p as MenuPopover).visible) as MenuPopover[];
+    openPopovers.forEach(p => p.hide());
+    await Promise.all(openPopovers.map(p => p.updateComplete));
```
maxinteger: Why did you changed this?

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -137,13 +153,22 @@ class SideNavigation extends Provider<SideNavigationContext> {
   @property({ type: Boolean, reflect: true, attribute: 'hide-fixed-section-divider' })
   hideFixedSectionDivider: boolean = false;
 
+  /**
+   * When `is-dropdown` is enabled, a parent `mdc-navmenuitem` can render its submenu inside a sibling
+   * `div[data-trigger="<navmenuitem-id>"]` while the sidenavigation is expanded. In collapsed mode,
+   * dropdown submenus are never rendered, even when `is-dropdown` is enabled.
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'is-dropdown' })
```
maxinteger: nitpicking: `<mdc-sidenavigation is-dropdown>` suggest the sidenav become a dropdown which is not true. `submenu-type="dropdown"` or `submenu-mode="dropdown"` would be more readable for the consumer. The default value would be "menu" or "fly-out"

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -1,48 +1,63 @@
...
+ *
+ * 4. **Action-based navmenuitem (no submenu)**:
+ *    - Performs an action such as navigation or alert trigger or expand/collapse (in case of dropdown) when clicked
+ *    - To maintain active state without navigation behavior, set `disable-aria-current="true"` on the navmenuitem. This allows the item to visually appear active while preventing screen readers from announcing it as the current page, which is ideal for items that trigger actions rather than navigation.
  *
  * ### Recommendations:
  * - Use `mdc-text` for section headers
  * - Use `mdc-divider` with the attribute `variant="gradient"` for section dividers
  * - For the brand logo, use an informative icon. Refer to `Momentum Informative Icons`
+ * - It is recommended to use icons for parent navmenuitems only. For submenu items (in both flyouts and dropdowns), it is recommended to use text labels without icons to ensure readability and usability.
```
maxinteger: :(

maxinteger (changes requested):

packages/components/src/components/menubar/menubar.component.ts
supminn: The depth manager is not recognising popover in the stack within menubar. Thus, it was causing an issue of keeping the menupopover open even when the adjacent parent's popover was open.

packages/components/src/components/sidenavigation/__screenshots__/chrome/sidenavigation-dropdown-expanded-open-rtl.png
maxinteger, line 1: on the LTR version the "Parent Item" had background color, but not here. 
All browsers has the same problem except Webkit

packages/components/src/components/sidenavigation/__screenshots__/chrome/sidenavigation-dropdown-re-expanded-rtl.png
maxinteger, line 1: on the LTR version the "Parent Item" had background color, but not here. 
All browsers has the same problem except Webkit

packages/components/src/components/sidenavigation/sidenavigation.context.ts
```diff
@@ -34,22 +42,39 @@ class SideNavigationContext {
...
+  public isDropDownParent(navMenuItem: NavMenuItem | undefined) {
+    const id = navMenuItem?.getAttribute('id');
+    if (!id) return false;
+
+    const siblings = Array.from(navMenuItem?.parentElement?.children ?? []);
+    return siblings.some(
+      sibling =>
+        sibling !== navMenuItem &&
+        sibling.tagName.toLowerCase() === 'div' &&
+        sibling.getAttribute('data-trigger') === id,
```
maxinteger: nitpicking:

```suggestion
        sibling !== navMenuItem && sibling.matches(`div[data-trigger="${id}"]`),
```

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -323,6 +385,231 @@ class SideNavigation extends Provider<SideNavigationContext> {
...
+    // Also hide all dropdown containers
+    this.hideAllDropdownContainers();
+  }
+
+  /**
+   * Hides all div[data-trigger] elements within the sidenavigation.
+   * @internal
+   */
+  private hideAllDropdownContainers(): void {
+    const containers = this.querySelectorAll<HTMLElement>('div[data-trigger]');
```
maxinteger, line 357: could not we use `:slotted`?

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -323,6 +385,231 @@ class SideNavigation extends Provider<SideNavigationContext> {
...
+          const triggerId = dropdownContainer.getAttribute('data-trigger');
+          const triggerIndex = outerItems.findIndex(item => item.id === triggerId);
+          if (triggerIndex !== -1 && triggerIndex + 1 < outerItems.length) {
+            const nextItem = outerItems[triggerIndex + 1];
+            nextItem.focus();
+            if (context.isDropDownParent(nextItem)) {
+              nextItem.openDropdown();
+              const nextDropdownContainer = nextItem.parentElement?.querySelector(
+                `div[data-trigger="${nextItem.id}"]`,
+              ) as HTMLElement | null;
```
maxinteger: please create a utility function for this selector to encapsulate the static part and type casting.

It can be defined inside the `handleDropdownKeyDown` method as well

packages/components/src/components/sidenavigation/sidenavigation.component.ts
```diff
@@ -323,6 +385,231 @@ class SideNavigation extends Provider<SideNavigationContext> {
...
+              if (nextDropdownContainer) {
+                const firstChild = nextDropdownContainer.querySelector(
+                  `${NAVMENUITEM_TAGNAME}:not([disabled])`,
+                ) as NavMenuItem | null;
```
maxinteger: This selector also used 3 places, please create a utility function fto encapsulate the static part and type casting.

It can be defined inside the `handleDropdownKeyDown` method as well

maxinteger (changes requested):
few minor comments

maxinteger (approved):

PR #1894, opened 2026-04-27
fix(toast): display Show More/Less button when the title goes beyond 2 lines

dominiccarrington:
### Description

- [toast] Fixes a bug where truncation of the header-text didn't work where the header-tag-name was not span or small in Chrome and Firefox
- [toast] Show the "Show More/Less" button when the toast title overflows the 2 line limit.

**Breaking Changes:** 
Makes Chrome/Firefox respect the line-clamp already in the style sheet. This could cause the header-text to be be clamped to 2 lines where it wasn't being clamped before.


packages/components/src/components/toast/toast.component.ts
```diff
@@ -144,9 +165,12 @@ class Toast extends FooterMixin(Component) {
         : '';
   }
 
-  protected override firstUpdated(changedProperties: PropertyValues): void {
+  protected override async firstUpdated(changedProperties: PropertyValues): Promise<void> {
     super.firstUpdated(changedProperties);
     this.updateDetailedSlotPresence();
+
+    await this.updateComplete;
+    this.hasOverflowingHeaderText = this.headerTextElement.isHeightOverflowing();
```
cscheffauer: its not a given that the headerTextElement has this function as a property (when consumer uses the slots / part), so we need to first check if the function is available and then call it - if its not available set the boolean to false.

cscheffauer (changes requested):

cscheffauer (approved):

PR #1898, opened 2026-05-01
fix(timepicker): open popover from blank trigger area

NotNestor:
### Description

This change updates mdc-timepicker so clicking the blank, non-editable area between the spinbutton controls and dropdown arrow toggles the popover, matching the arrow button behavior. Clicking directly on spinbutton inputs still focuses/edit them and does not open the dropdown.

### Links

N/A


packages/components/src/components/timepicker/timepicker.e2e-test.ts
```diff
@@ -186,6 +188,43 @@ test.describe('mdc-timepicker', () => {
...
+        componentsPage,
+        label: 'Start time',
+        value: '08:30',
+        width: '12rem',
+      });
+
+      const spinbuttonGroup = timepicker.locator('[part="spinbutton-group"]');
+      const box = await spinbuttonGroup.boundingBox();
+      if (!box) {
+        throw new Error('Expected spinbutton group bounding box');
```
supminn:
```suggestion
       await expect(box).not.toBeUndefined() 
```

I believe `expect` would be the preferred way to check instead of throwing an error from the test file.
Not a blocker though.

supminn (approved):

PR #1899, opened 2026-05-04
docs: added docs to the repo for adrs and project briefs

tricarro:
### Description

Created a new root folder: .docs. This folder will contain documentation for key decision, projects, or other related materials that we would like to store for our collaborative efforts. 

There are two subfolders within the .docs folder: "decisions" and "projects". The decisions folder will contain Architectural Decision Records. These are key decisions we make as a team related to the products that we build.

The project folder will contain project briefs for specific projects or initiatives that we undertake. 

There are proposed templates for each document type.

### DIFF
--- a/.docs/projects/_OVERVIEW.md
+++ b/.docs/projects/_OVERVIEW.md
@@ -1,0 +1,40 @@
+# Projects Overview
+
+This document contains design briefs for key projects that Momentum Design System team members are working on. These documents will be used to align on project goals and define project success.
+
+Here is the standard format of a Project Design Brief:
+
+# Project Title
+
+## Overview
+General idea or summary of the project.
+
+## Problem
+The primary (and secondary) problems that this work will solve.
+
+## Background
+Background information on this work. 
+
+## Goals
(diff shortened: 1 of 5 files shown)


maxinteger (changes requested):
Couple of thins.
1. Why do we have `decisions` and `projects` folder?. We can have a global and one for each packages if the ADRs are too specific, but we should not have 2 ADRs on the top level.
2. Decisions and ASDs should be in the same file, for example ".docs/projects/2026-5-3_knowledge-base.md", does not have a "Decision" section, but it should
3. Kind of nit picking but usually files are sorted alphabetically, so please use `0` prefixed dates in the file names eg.: `2026-05-03`

.docs/projects/_OVERVIEW.md
```diff
@@ -0,0 +1,40 @@
+# Projects Overview
```
maxinteger, line 1: I can not see "Proposed solution(s)" and "Decision" sections

maxinteger (changes requested):

PR #1900, opened 2026-05-05
fix(tab): icon on selected tab not rendered when -filled icon is used

jor-row:
### Description

Support using a filled icon for the unselected state. Currently if a filled icon is used for the regular icon, the filled icon is undefined.

Before

https://github.com/user-attachments/assets/af6efac3-d2dd-4d3f-a4b5-ceb0b77b185d

After

https://github.com/user-attachments/assets/fb88b506-32ce-4455-bed6-b27870faaaa1


maxinteger (approved):
the code looks good and works as expected. 

please check the failing CI

PR #1902, opened 2026-05-06
chore(animation): enable animation tokens for web

2020Deeya:
### Description

This PR adds the Animation motion tokens to the design token pipeline of the Motion Provider architecture (Core → Animation → Component).

#### Note: 
- This currently supports web only.
- ATM, Sample animations have been added to streamline the process.


packages/tools/token-builder/src/schemas/json/motion/animation.json
```diff
@@ -0,0 +1,181 @@
...
+        "max-width",
+        "opacity",
+        "outline",
+        "rotate",
+        "scale",
+        "transform",
+        "translate",
+        "visibility",
+        "width"
+      ]
```
maxinteger, line 38: Please check these values again, some of the are not CSS properties but CSS values, for example: `scale`, `rotate` and `translate` are values of `transform`

packages/assets/tokens/scripts/build-animation-css.js
```diff
@@ -0,0 +1,136 @@
...
+    if (current == null || typeof current !== 'object') {
+      throw new Error(`Unresolved token reference: "${ref}" — path segment "${part}" not found.`);
+    }
+    current = current[part];
+  }
+  if (!current || !current.value) {
+    throw new Error(`Unresolved token reference: "${ref}" — no value found.`);
+  }
+  return current.value;
+}
```
maxinteger: please use lodash (or similar) instead

lodash [`get`](https://lodash.com/docs/4.17.23#get) should do the trick:

packages/assets/tokens/scripts/build-animation-css.js
```diff
@@ -0,0 +1,136 @@
...
+    ...token,
+    duration: resolveRef(token.duration, resolvedTokens),
+    easing: resolveRef(token.easing, resolvedTokens),
+    delay: resolveRef(token.delay, resolvedTokens),
+  };
+}
+
+function toKebab(camelName) {
+  return camelName.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
+}
```
maxinteger: lodash [`kebabCase`](https://lodash.com/docs/4.17.23#kebabCase)

packages/assets/tokens/scripts/build-animation-css.js
```diff
@@ -0,0 +1,136 @@
...
+
+    if (token.type === 'transition') {
+      variableLines.push(`  --mds-transition-${kebab}: ${buildTransitionValue(token)};`);
+
+    } else if (token.type === 'keyframe') {
+      const kfName = `mds-animation-${kebab}`;
+      keyframeBlocks.push(buildKeyframeBlock(kfName, token));
+      const iteration = token.iterationCount ? ` ${token.iterationCount}` : '';
+      variableLines.push(
+        `  --mds-animation-${kebab}: ${kfName} ${token.duration} ${token.easing} ${token.delay}${iteration};`,
```
maxinteger: in the [CSS animation](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/animation) shortcut the animation-name must be the last item, please check the order for the reset of the parameters

packages/assets/tokens/scripts/build-animation-motion-css.js
```diff
@@ -0,0 +1,136 @@
...
+function buildKeyframeBlock(kfName, token) {
+  const fromProps = {};
+  const toProps = {};
+  token.keyframes.forEach(({ propertyName, from, to }) => {
+    fromProps[propertyName] = from;
+    toProps[propertyName] = to;
+  });
+  const fromStr = Object.entries(fromProps).map(([p, v]) => `${p}: ${v}`).join('; ');
+  const toStr = Object.entries(toProps).map(([p, v]) => `${p}: ${v}`).join('; ');
+  return `@keyframes ${kfName} {\n  from { ${fromStr}; }\n  to { ${toStr}; }\n}`;
```
maxinteger, line 58: we are validating (limiting) the css property names in the schema, but we do not check the values.

We can do it here or during validation after the scheme passed. For example [cssType](https://www.npmjs.com/package/csstype) should do the trick (IF we stick to the css values)

We do not need this validation right now, but please raise a tech debpt issue!

packages/assets/tokens/scripts/build-animation-css.js
```diff
@@ -0,0 +1,136 @@
...
+        return ref ? buildTransitionValue(ref).split(', ') : [];
+      });
+      if (parts.length) {
+        variableLines.push(`  --mds-transition-${kebab}: ${parts.join(', ')};`);
+      }
+
+    } else if (token.type === 'compoundKeyframes') {
+      const parts = token.animations.map((refName) => {
+        const ref = resolved[refName];
+        if (!ref) return null;
```
maxinteger: throw invalid ref error here please

packages/tools/token-builder/src/schemas/json/motion/animation.json
```diff
@@ -0,0 +1,181 @@
...
+            "properties": {
+              "properties": false,
+              "animations": false,
+              "composition": false
+            }
+          }
+        },
+        {
+          "if": {
+            "properties": { "type": { "const": "compoundTransitions" } },
```
maxinteger: minor, but we should call this `transitionCompaund`

packages/tools/token-builder/src/schemas/json/motion/animation.json
```diff
@@ -0,0 +1,181 @@
...
+              "properties": false,
+              "keyframes": false,
+              "iterationCount": false,
+              "fillMode": false
+            }
+          }
+        },
+        {
+          "if": {
+            "properties": { "type": { "const": "compoundKeyframes" } },
```
maxinteger: minor, but we should call this `keyframeCompaund`

packages/assets/tokens/scripts/build-animation-motion-css.js
maxinteger, line 1: post test is fine you can keep it, but we should have proper unit test against the generator

maxinteger, line 1: The generator ignores the `composition` property on the compound animations.

It is fine, we do not have actual use case for that right now, but please raise a new story about it

maxinteger (changes requested):

packages/tools/token-builder/src/schemas/json/motion/animation.json
```diff
@@ -0,0 +1,181 @@
...
+      "additionalProperties": false,
+      "properties": {
+        "description": { "type": "string" },
+        "type": {
+          "type": "string",
+          "enum": ["transition", "keyframe", "compoundTransitions", "compoundKeyframes"]
+        },
+        "duration": { "$ref": "#/definitions/TokenRef" },
+        "easing": { "$ref": "#/definitions/TokenRef" },
+        "delay": { "$ref": "#/definitions/TokenRef" },
```
maxinteger, line 76: these should be available only for `transition` and `keyframe` animations and not for `compound`

maxinteger (changes requested):

packages/tools/token-builder/src/schemas/json/motion/animation.json
```diff
@@ -0,0 +1,181 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$ref": "#/definitions/AnimationFile",
+  "definitions": {
+    "TokenRef": {
+      "type": "string",
+      "pattern": "^\\{[a-zA-Z_][a-zA-Z0-9_]*(\\.[a-zA-Z_][a-zA-Z0-9_]*)+\\}$"
+    },
```
maxinteger, line 9: please add `description` for each type, for example:


```suggestion
    "TokenRef": {
      "type": "string",
      "pattern": "^\\{[a-zA-Z_][a-zA-Z0-9_]*(\\.[a-zA-Z_][a-zA-Z0-9_]*)+\\}$"
      "description": "Path to a Momentum design token"
    },
```

maxinteger (changes requested):

packages/tools/token-builder/src/schemas/json/motion/animation.json
maxinteger: Sorry, I meant all definitions should have `description`, like: `TokenIdentifier`, `CssAnimatableProperty`, `AnimationFile` etc.

maxinteger (dismissed):

maxinteger (approved):

PR #1904, opened 2026-05-07
chore: add custom state to checkbox component

2020Deeya:
### Description

This PR adds a custom 'checked' property to `mdc-checkbox`  so the checkbox's checked status is accessible via :state(checked).


packages/components/src/components/button/button.component.ts
```diff
@@ -71,11 +71,13 @@ import { getIconNameWithoutStyle } from './button.utils';
...
  * @cssproperty --mdc-button-height - Height for button size
  * @cssproperty --mdc-button-background - Background of the button
- * @cssproperty --mdc-button-border-color - Borer color of the button
+ * @cssproperty --mdc-button-border-color - Border color of the button
  * @cssproperty --mdc-button-text-color - Text color of the button
  * @cssproperty --mdc-button-prefix-icon-size - Size of the prefix icon
  * @cssproperty --mdc-button-postfix-icon-size - Size of the postfix icon
  * @cssproperty --mdc-button-line-height - Line height of the button text
+ *
+ * @cssstate loading - Active when `loading` is `true`. Use `mdc-button:state(loading)` to style externally.
```
maxinteger: " Use `mdc-button:state(loading)` to style externally." part is unnecessary. If we really want to document this, then we should put it into one of the top level .md file but not for every `@cssstate`

packages/components/src/components/button/button.component.ts
```diff
@@ -107,6 +109,32 @@ class Button extends OverflowMixin(ButtonComponentMixin(Buttonsimple)) {
   @property({ type: Number })
   override size: PillButtonSize | IconButtonSize = DEFAULTS.SIZE;
 
+  /**
+   * Indicates the button is in a loading state. When set:
+   * - Suppresses interaction
+   * - Sets `aria-busy="true"` to notify screen readers that the action is in progress.
+   * - Exposes `:state(loading)` via CSS Custom States API (`ElementInternals.states`).
```
maxinteger:
```suggestion
   * - Set CSS `:state(loading)` .
```

maxinteger (changes requested):
Few issues: 

## 1 
The component is spinning in storybook.

Image

It would be better to spin an icon.

## 2

I can not see "css states" section in button's control in Storybook.

2020Deeya:
> Few issues:
> 
> ## 1
> The component is spinning in storybook.
> 
> Image
> It would be better to spin an icon.
> 
> ## 2
> I can not see "css states" section in button's control in Storybook.

1] okay.
2] It's there in the bottom.
image

packages/components/src/components/button/button.styles.ts
```diff
@@ -286,11 +286,25 @@ const styles = css`
...
+      transform: rotate(0deg);
+    }
+    to {
+      transform: rotate(360deg);
+    }
+  }
+
+  :host(:state(loading))::part(prefix-icon) {
+    animation: mdc-icon-spin 600ms linear infinite;
+  }
```
maxinteger: We do not know what kind of animation the consumer want to apply so please move this animation back to the story

maxinteger (changes requested):

maxinteger (approved):

PR #1906, opened 2026-05-08
refactor(input): add character counter support

NotNestor:
### Description

Add character counter to mdc-input, matching the existing mdc-textarea implementation.
When `max-character-limit` is set, a counter (e.g. "12/75") is displayed in the input footer alongside the help text.

- Add `max-character-limit` and `character-limit-announcement` properties
- Add `limitexceeded` event fired when the character limit is crossed or restored
- Add screen reader announcements via mdc-screenreaderannouncer
- Block form submission when limit is exceeded with error help-text-type
- Render input footer with help text and character counter side by side
- Add character counter styles with error and disabled state variants
- Add InputWithCharacterCounter story
- Add e2e tests for character counter attributes, interactions, form
  validation, and visual regression

### Links

N/A


packages/components/src/components/input/input.component.ts
```diff
@@ -64,6 +66,7 @@ import styles from './input.styles';
  * @cssproperty --mdc-input-support-text-color - Text color for the help text
  * @cssproperty --mdc-input-selection-text-color - Text color for the selected text
  * @cssproperty --mdc-input-selection-background-color - Background color for the selected text
+ * @cssproperty --mdc-input-text-secondary-normal - Text color for the character counter
```
supminn: Do we want to have a new variable for this? Can we not reuse `--mdc-input-support-text-color` for both help text as well as character counter?

supminn (changes requested):
1. I believe you've copied over the code from textarea component to the input component as well, right? I'd suggest to create a new mixin and share between the 2 components - input and textarea?
2. The snapshots for inputchip is incorrectly updated. Please get them removed.

packages/components/src/utils/mixins/CharacterLimitMixin.ts
```diff
@@ -0,0 +1,152 @@
...
+          composed: true,
+        }),
+      );
+    }
+
+    /**
+     * Handles the character overflow state change.
+     * Dispatches the character overflow state change event if the character limit is exceeded or restored.
+     */
+    handleCharacterOverflowStateChange() {
```
cscheffauer: issue: all those functions here are public, thus will be public on the component which extends this mixin. We need to make those protected, given the component shouldn't expose those functions as a public method. Same applies for the Typescript Interface at the top of the file.

cscheffauer (dismissed):

packages/components/config/playwright/public/assets/animations/animation.json
supminn, line 1: Why was this file removed?

NotNestor: it wasn't removed it looks like it got minified by the github action to update snapshots for some reason

supminn (approved):

PR #1907, opened 2026-05-08
chore(component): enable component tokens for web

2020Deeya:
### Description

This PR adds the Component motion tokens to the design token pipeline of the Motion Provider architecture (Core → Animation → Component).

#### Note:
This currently supports web only.
ATM, Sample component tokens have been added to streamline the process.


packages/tools/token-builder/src/schemas/json/motion/component.json
```diff
@@ -0,0 +1,67 @@
...
+      "type": "object",
+      "description": "Binds an animation token to a component part and optional state(s)",
+      "required": ["part", "animation"],
+      "additionalProperties": false,
+      "properties": {
+        "part": { "$ref": "#/definitions/PartName" },
+        "state": {
+          "type": "array",
+          "items": { "$ref": "#/definitions/StateName" },
+          "minItems": 1,
```
maxinteger, line 60: we need to represent default state which is no state.

We either use empty array or a specific name. Since we already use specific name for the component itself in the `part` we should do the same here.

We can use `:noState` or `:default` so the `minItems` can remain

packages/tools/token-builder/src/schemas/json/motion/component.json
```diff
@@ -0,0 +1,67 @@
...
+          "uniqueItems": true,
+          "description": "One or more states that must be active for this rule to apply"
```
maxinteger: Please update 
```suggestion
          "description": "Animation assigned to the component's part when the all the listed states are active on the component itself (not the part)"
```

packages/tools/token-builder/src/schemas/json/motion/component.json
```diff
@@ -0,0 +1,67 @@
...
+          "items": { "$ref": "#/definitions/ComponentEntry" },
+          "minItems": 1
+        }
+      },
+      "additionalProperties": false,
+      "title": "ComponentMap"
+    },
+    "ComponentEntry": {
+      "type": "object",
+      "description": "Binds an animation token to a component part and optional state(s)",
```
maxinteger:
```suggestion
      "description": "Binds the animation token to the component's part when the state(s) are active on the component",
```

packages/tools/token-builder/src/schemas/json/motion/component.json
```diff
@@ -0,0 +1,67 @@
...
+  "definitions": {
+    "AnimationIdentifier": {
+      "type": "string",
+      "pattern": "^[a-zA-Z][a-zA-Z0-9]*$",
+      "description": "A camelCase identifier referencing an animation token by name"
+    },
+    "StateName": {
+      "type": "string",
+      "pattern": "^[a-z][a-z0-9-]*$",
+      "description": "A CSS pseudo-class name (e.g. 'hover', 'focus-visible') or CustomStateSet name (e.g. 'loading')"
```
maxinteger:
```suggestion
     "oneOf": [
        { "const": ":noState" },
        { "type": "string", "pattern": "^[a-z][a-z0-9-]*$" }
      ]
      "description": "component state eg.: hover, enable, checked, etc. `:noState` is a special state to represent default/no state of the component"
```

packages/tools/token-builder/src/schemas/json/motion/component.json
```diff
@@ -0,0 +1,67 @@
...
+        "component": {
+          "$ref": "#/definitions/ComponentMap"
+        }
+      },
+      "required": ["component"],
+      "title": "ComponentFile"
+    },
+    "ComponentMap": {
+      "type": "object",
+      "description": "A map of component element names (e.g. 'mdc-button') to their animation entries",
```
maxinteger:
```suggestion
      "description": "A map of component element names (e.g. 'button') to their animation entries",
```

packages/assets/tokens/scripts/build-component-motion-css.js
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Maps a single state name to its CSS pseudo-class string.
+ * Native pseudo-classes use `:name`; custom states use `:state(name)`.
+ *
+ * @param {string} state
+ * @returns {string}
+ */
+function stateToSelector(state) {
+  return NATIVE_PSEUDO_CLASSES.has(state) ? `:${state}` : `:state(${state})`;
```
maxinteger: Unfortunately we can not do this way.

We need to do it the other way around.

1. If `state === ':noState'` then return empty string 
1. otherwise if the state is in the components-manifest.json for the specified component, use `:state(${state})`;
2. otherwise if state is in `NATIVE_PSEUDO_CLASSES` use`:${state}`
3. otherwise throw an error

maxinteger (changes requested):

packages/tools/token-builder/src/schemas/json/motion/component.json
```diff
@@ -0,0 +1,70 @@
...
+  "definitions": {
+    "AnimationIdentifier": {
+      "type": "string",
+      "pattern": "^[a-zA-Z][a-zA-Z0-9]*$",
+      "description": "A camelCase identifier referencing an animation token by name"
+    },
+    "StateName": {
+      "oneOf": [
+        { "const": ":noState" },
+        { "type": "string", "pattern": "^:state\\([a-z][a-z0-9-]+\\)$" },
```
maxinteger: please remove this pattern

packages/tools/token-builder/src/schemas/json/motion/component.json
```diff
@@ -0,0 +1,70 @@
...
+        { "type": "string", "pattern": "^[a-z][a-z0-9-]*$" }
+      ],
+      "description": "component state. Use ':noState' for the default/no-state rule, ':state(name)' for custom component states (CSS custom state API), or a plain name for native pseudo-classes (e.g. 'hover', 'disabled')."
```
maxinteger: Sorry, maybe I was not clear about this, but component.json should be ass platform agnostic as possible. This mean
- We should not mention CSS anywhere in the schema
- We should not use logic that is CSS specific


```suggestion
      "description": "component state. Use ':noState' for the default/no-state, or any state specified by the component (hover, enabled, focus, active, checked, etc.)"
```

packages/assets/tokens/scripts/build-component-motion-css.js
```diff
@@ -0,0 +1,151 @@
...
+// and will pass through as-is in step 2. Unknown states will throw an error.
+const NATIVE_PSEUDO_CLASSES = new Set([
+  // Interaction states — browser sets these on any element
+  'hover',
+  'active',
+  'focus',
+  'focus-visible',
+  'focus-within',
+  // Form states — browser sets these on form-associated custom elements
+  // via ElementInternals (static formAssociated = true)
```
maxinteger: These states are set for native form elements as well. This comment is confusing please remove it.

packages/assets/tokens/scripts/build-component-motion-css.js
```diff
@@ -0,0 +1,151 @@
...
+const SRC_COMPONENT = path.join(__dirname, '../src/motion/component.json');
+const SRC_ANIMATION = path.join(__dirname, '../src/motion/animation.json');
+const OUTPUT_DIR = path.join(__dirname, '../dist/css/motion');
+const OUTPUT_FILE = path.join(OUTPUT_DIR, 'component.css');
+
+const KEYFRAME_TYPES = new Set(['keyframe', 'keyframeCompound']);
+const TRANSITION_TYPES = new Set(['transition', 'transitionCompound']);
+
+// Native CSS pseudo-classes that map directly to `:name` syntax (step 3 of stateToSelector).
+// Custom component states must be written explicitly as ":state(name)" in component.json
```
maxinteger: remove this line please

packages/assets/tokens/scripts/build-component-motion-css.js
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * @param {string} state
+ * @returns {string}
+ */
+function stateToSelector(state) {
+  if (state === ':noState') return '';
+
+  if (/^:state\([a-z][a-z0-9-]+\)$/.test(state)) {
+    return state;
+  }
```
maxinteger: this is wrong, custom state information should came form the `packages/components/dist/custom-elements.json` and not from the `components.json`

packages/assets/tokens/src/motion/component.json
```diff
@@ -0,0 +1,39 @@
+{
+  "component": {
+    "mdc-button": [
```
maxinteger, line 3: Do not change it for now, but we need to talk about the `mdc-`prefix.

It is convenient for use to just use the component name for selection, but on other platforms probably they do not have this prefix

We could make the different between the platforms here for example use different prefix like:
- `button` (default)
- `web-button` (web specifi)
- `android-button` (android specific)
- `qt-button` (Qt specific)
- etc.

maxinteger (changes requested):
Sorry, probably I referred to the wrong json when I requested the changes in the `stateToSelector` function.

packages/assets/tokens/scripts/build-component-motion-css.js
2020Deeya: Please check my comment in webex chat.\

maxinteger (approved):

packages/assets/tokens/src/motion/component.json
```diff
@@ -0,0 +1,39 @@
...
+      {
+        "part": ":host",
+        "state": ["hover"],
```
maxinteger, line 6: `"state": ["hover", "hover"]` will generate `mdc-button:hover:hover {`

make sure every state appears only once in the css selector

maxinteger (changes requested):
I am getting error during the build

Image

2020Deeya:
> I am getting error during the build
> 
> Image

Build is running successfully, The affected schemas are the pre-existing dark.json theme schemas (ex: #/properties/contrast, #/properties/positive).  They're not errors, these warnings existed before our branch, they don't fail the build.
We should have a separate PR to target them.

package.json
```diff
@@ -15,7 +15,7 @@
     "analyze:config": "eslint \"./config/**\"",
     "analyze:root": "yarn analyze:config && yarn analyze:dependencies",
     "analyze": "yarn workspaces foreach --topological-dev -p run analyze",
-    "build": "yarn workspaces foreach --topological-dev -p run build",
+    "build": "yarn workspaces foreach --topological-dev -p run build && yarn tokens build:component-motion",
```
cscheffauer, line 18: issue: the top level build should always run the build of the packages below. This violates the idea of yarn workspaces. Instead the build:component-motion should be part of the tokens package build.

packages/assets/tokens/package.json
```diff
@@ -15,11 +15,12 @@
     "build:elevation": "md-token-builder --config ./config/tokens/elevation.json --input ./src --output ./dist",
     "build:motion": "md-token-builder --config ./config/tokens/motion.json --input ./src --output ./dist",
     "build:animation": "node ./scripts/build-animation-motion-css.js",
+    "build:component-motion": "node ./scripts/build-component-motion-css.js",
     "build:additionalthemes": "md-token-builder --config ./config/tokens/additional-themes.json --input ./src --output ./dist",
     "build:components": "node ./scripts/build-components-css.js",
     "clean": "yarn clean:dist",
     "clean:dist": "rimraf ./dist",
-    "prepublishOnly": "yarn clean && yarn build",
+    "prepublishOnly": "yarn clean && yarn build && yarn build:component-motion",
```
cscheffauer, line 23: q: why not doing it as part of the build script? when we run build in tokens package, we want everything to be built, thus it needs to be there.

packages/assets/tokens/package.json
```diff
@@ -15,11 +15,12 @@
     "build:elevation": "md-token-builder --config ./config/tokens/elevation.json --input ./src --output ./dist",
     "build:motion": "md-token-builder --config ./config/tokens/motion.json --input ./src --output ./dist",
     "build:animation": "node ./scripts/build-animation-motion-css.js",
+    "build:component-motion": "node ./scripts/build-component-motion-css.js",
```
cscheffauer, line 18: why inventing a new script, not using the md-token-builder for it?

packages/assets/tokens/scripts/build-component-motion-css.js
```diff
@@ -0,0 +1,160 @@
+const fs = require('fs');
```
cscheffauer, line 1: issue: having a separate script for this is not ideal. The token-builder should do this.

cscheffauer (changes requested):

PR #1908, opened 2026-05-08
feat(animation): accept lottie json src url in addition to known name

gabrielchl:
### Description
for the custom reactions feature
currently the animation component only accepts a known animation name
this PR adds a src property which can be a link pointing to a lottie json file

### Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-810566


maxinteger (changes requested):
Looks good and works as expected.

My only concern is the external lotti file in the E2E tests.
- The file can be removed which can cause test failure
- Test can fail if the external host of the lotti file is slow
 
Can we use our existing lotti file instead served from localhost?

I think `/playwright-temp/assets/animations/animation.json` should do the trick. You can add a different animation as well in `packages/components/scripts/copyAnimations.js`

maxinteger (approved):

PR #1911, opened 2026-05-13
fix(popover): prevent premature close on mouseleave within trigger shadow DOM

NotNestor:
### Description

When a trigger element contains shadow DOM children (e.g., an mdc-button
with a prefix icon), mouseleave events fire as the cursor moves between
internal shadow DOM elements like SVGs. This caused the popover to close
unexpectedly even though the mouse was still within the trigger.

Add an `isDescendantOfTrigger` helper that walks up the DOM tree across
shadow root boundaries to determine if the mouseleave relatedTarget is
still inside the trigger. If it is, the close is suppressed.

Add an e2e test that verifies the popover stays open when mouseleave
fires with a shadow DOM child as the relatedTarget, and still closes
when the mouse truly leaves the trigger.

### Links

N/A


packages/components/src/components/popover/popover.component.ts
```diff
@@ -1210,6 +1221,27 @@ class Popover
...
 
+  /**
+   * Checks if a node is a descendant of the trigger element, crossing shadow DOM boundaries.
+   * This is needed because `Element.contains()` does not traverse into shadow roots.
+   * @param node - The node to check.
+   * @param trigger - The trigger element.
+   * @returns True if the node is inside the trigger element.
+   * @internal
+   */
+  private isDescendantOfTrigger(node: Node, trigger: HTMLElement): boolean {
```
cscheffauer: suggestion: we do have DOM utils for this already, see here:
https://github.com/momentum-design/momentum-design/blob/995b19e9d1eef5d75d620391903fd8f3256c06b8/packages/components/src/utils/dom.ts#L27
i would suggest to make use of them.

cscheffauer (changes requested):

cscheffauer (approved):

PR #1916, opened 2026-05-20
fix: tooltip-default-text-style

cscheffauer:
### Description

- Fix default text style of Tooltip (to be body-midsize-regular)
- Updated E2E Test Screenshots


cscheffauer:
not necessary

PR #1921, opened 2026-05-27
fix(tooltip): prevent infinite render loop when id prop is unset

NotNestor:
### Description

The previous implementation generated the tooltip's auto-id from inside
`update()` and then `await`ed `this.updateComplete`, which scheduled a
follow-up render cycle for every id assignment. Under the @lit/react
wrapper - which re-passes props on every React render and reset `id`
back to "" each time - this produced a non-terminating chain of
updates, locking the host app's main thread.

Move the id derivation to `willUpdate()`, where the assignment is
folded into the in-flight cycle (no extra render), and cache the
generated id so a repeated reset to an empty value restores the same
value instead of minting a fresh UUID. Trigger aria-attribute sync is
performed in the same hook.

Behavior preserved:
- Auto-generates an id when none is provided.
- Re-populates the id if the consumer clears it after mount.
- Explicit consumer-set ids still win and invalidate the cache so a
  later clear can regenerate.
- aria-describedby / aria-labelledby on the trigger continue to track
  the current id and tooltipType.

### Links

N/A


cscheffauer (approved):
great catch! thanks for the contribution

PR #1924, opened 2026-05-28
fix(popover): show on hover when trigger lives inside shadow DOM

NotNestor:
### Description

Hover triggers used document-level `mouseenter`/`mouseleave` listeners. Those
events are spec'd as `composed: false` and do not bubble, so when the trigger
lives inside a shadow root (e.g. wrapped in `mdc-iconprovider`/`mdc-themeprovider`)
the event path never reaches the document listener and the popover never opens.
Delegate hover via `mouseover`/`mouseout` instead. They are `composed: true` and
bubble, so they cross shadow boundaries and reach the document-level listener,
which lets us keep event delegation — no direct reference to the trigger element
and no dependency on its mount/unmount lifecycle. The handlers reuse the existing
`isEventFromTrigger` (composedPath) check and a subtree guard on `relatedTarget`
to recreate enter/leave semantics and ignore movements within the trigger.

### Links

N/A

### DIFF
--- a/packages/components/src/components/popover/popover.component.ts
+++ b/packages/components/src/components/popover/popover.component.ts
@@ -950,14 +950,33 @@ class Popover
   }
 
   /**
+   * Determines whether a delegated `mouseover`/`mouseout` event is movement that stays within the
+   * trigger's subtree rather than a genuine enter/leave of the trigger.
+   *
+   * Both events fire repeatedly while the pointer moves between elements inside the trigger
+   * (e.g. an icon with internal SVG elements). `relatedTarget` is the element on the other side of
+   * the boundary (the element being left for `mouseover`, or entered for `mouseout`); if it
+   * resolves to the trigger or one of its shadow hosts, the pointer has not crossed the trigger
+   * boundary and the event should be ignored.
+   * @internal
+   */
+  private isHoverWithinTrigger = (event: Event): boolean => {
+    const { triggerElement } = this;
+    const { relatedTarget } = event as MouseEvent;
+    if (triggerElement && relatedTarget instanceof Element) {
+      return getHostComposePath(relatedTarget).includes(triggerElement);
+    }
+    return false;
+  };
+
+  /**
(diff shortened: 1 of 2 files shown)


maxinteger (changes requested):
Thank you for the contribution.

Originally all types of trigger events registered directly on the trigger component. But because the popover does not have any control  over the trigger component (it does not know when the component mounted/un-mounted). We switched to delegated events to not remove dependency on the trigger component.

Unfortunately your change brings back the direct reference to the trigger component, but only for `mouseenter` case which also add inconsistency to the codebase.

I did some investigation and looks like `mouseover`/`mouseout` are composed and can bubble.  

The only issue with them is they can trigger on every elements of the trigger element's sub-tree. This snippet can avoid this problem:
```
const { target, relatedTarget } = event;

    if (relatedTarget && target.contains(relatedTarget)) {
      return; // Ignore deep-nested movements within the same element tree
    }
```
With this approach we can keep the event delegation.

packages/components/src/components/popover/popover.component.ts
```diff
@@ -954,17 +964,17 @@ class Popover
...
    */
   private handleMouseLeave = (event: Event) => {
     if (!this.isEventFromTrigger(event)) return;
 
-    // When the trigger contains shadow DOM children (e.g. an icon with internal SVG elements),
-    // mouseleave fires on internal elements as the mouse moves between them.
-    // Only close if the mouse has actually left the trigger element.
+    // `mouseout` also fires while the pointer moves between elements inside the trigger
+    // (e.g. an icon with internal SVG elements). Only close if the pointer has actually left the
+    // trigger's subtree — i.e. the element being entered is not the trigger or one of its hosts.
```
maxinteger: `mouseover` also fires multiple times, we need the same safeguard as `mouseout`

maxinteger (changes requested):

maxinteger (approved):

PR #1928, opened 2026-05-31
fix(spatialnavigation): increase data-spatial-dir attribute priority

maxinteger:
### Description

Prevent all other navigation when any of the `data-spatial-{left|right|up|down}` attribute defined on the active element.

This helps to handle edge cases where default navigation behaviour is incorrect, or it is not user-friendly

Breaking change: none

### DIFF
--- a/packages/components/src/components/spatialnavigationprovider/spatialnavigationprovider.component.ts
+++ b/packages/components/src/components/spatialnavigationprovider/spatialnavigationprovider.component.ts
@@ -18,24 +18,43 @@ import { SpatialNavigationEvent } from './spatialnavigationprovider.events';
 
 // AI-Assisted
 /**
- * This component manages focus using spatial navigation and provides context for child components.
- *
- * Place it at the root of the application.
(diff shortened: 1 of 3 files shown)


packages/components/src/components/spatialnavigationprovider/spatialnavigationprovider.component.ts
```diff
@@ -293,6 +329,7 @@ class SpatialNavigationProvider extends Provider<SpatialNavigationContextValue>
   override connectedCallback() {
     super.connectedCallback();
 
+    document.addEventListener('keydown', this.handleKeyDownBefore, { capture: true });
```
cscheffauer, line 332: issue: this needs to be cleanedup no? like the this.handleKeyDownBefore function is bound to the keydown, but not cleaned up in disconnectedCallback

cscheffauer (changes requested):

cscheffauer (approved):

PR #1951, opened 2026-06-17
feat(components): add status message component

NotNestor:
### Description

Add mdc-statusmessage as a reusable inline severity/status message primitive with default severity icons, custom icon support, message fallback, slots, CSS parts, and CSS variables.

Refactor formfieldwrapper helper text rendering to compose mdc-statusmessage while preserving existing help-text APIs, helper icon behavior, CSS parts, and aria-describedby wiring.

Add Storybook stories, knowledge-base docs, and e2e coverage for severity icons, custom icons, message fallback, slotted content, wrapping text, and formfieldwrapper compatibility.

### Links

N/A


packages/components/src/components/formfieldwrapper/formfieldwrapper.component.ts
```diff
@@ -226,10 +226,14 @@ class FormfieldWrapper extends DisabledMixin(Component) {
     if (!this.helpText) {
       return nothing;
     }
-    return html`<div part="help-text-container">
-      <slot name="help-icon">${this.renderHelpTextIcon()}</slot>
+    return html`<mdc-statusmessage
```
cscheffauer, line 229: the stepperitem component implements its own status message help text - can we update that one too possibly?

packages/components/src/components/statusmessage/statusmessage.component.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import { Component } from '../../models';
+
+import { DEFAULTS } from './statusmessage.constants';
+import styles from './statusmessage.styles';
+import type { StatusMessageSeverity } from './statusmessage.types';
+import { getStatusMessageIcon } from './statusmessage.utils';
+
+/**
+ * @remarks
```
cscheffauer: remarks is not a tag we use - remove it since this is covered in the knowledge base. In the new system the TSDoc will be automatically enhanced on build, so every description / summary / remark just goes directly into the knowledge-base md now.

packages/components/src/components/statusmessage/statusmessage.component.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  @state()
+  private hasDefaultSlotContent = false;
+
+  protected override firstUpdated(): void {
+    this.updateDefaultSlotContent();
+  }
+
+  private updateDefaultSlotContent() {
+    this.hasDefaultSlotContent = this.defaultSlotNodes.some(node => Boolean(node.textContent?.trim()));
```
cscheffauer: i think this trimming is unnecessary - we can leave it up to the consumer to provide trimmed text in the slot anyways. I wouldn't modify the slot (unnecessary bloat imho)

packages/components/src/components/statusmessage/statusmessage.component.ts
```diff
@@ -0,0 +1,99 @@
...
+      <mdc-text part="text" tagname="${DEFAULTS.TEXT_TAGNAME}" type="${DEFAULTS.TEXT_TYPE}">${this.message}</mdc-text>
+    `;
+  }
+
+  public override render() {
+    return html`
+      <div part="container">
+        <slot name="icon" part="icon">${this.renderIcon()}</slot>
+        <slot part="text" @slotchange="${this.updateDefaultSlotContent}"></slot>
+        ${this.renderMessage()}
```
cscheffauer: instead of updateDefaultSlotContent in line 89, we can put this renderMessage as the fallback of the slot, so whenever a slot is provided the message will not be rendered, and if the slot is not provided the message will take precedence.

packages/components/src/components/statusmessage/statusmessage.utils.ts
```diff
@@ -0,0 +1,18 @@
+// AI-Assisted
+import { STATUSMESSAGE_ICON_NAMES, STATUSMESSAGE_SEVERITY } from './statusmessage.constants';
+import type { StatusMessageIcon, StatusMessageSeverity } from './statusmessage.types';
+
+const getStatusMessageIcon = (severity: StatusMessageSeverity): StatusMessageIcon => {
```
cscheffauer: nitpick: we can put this map into the component itself or also constants, like having this function here as a util is not really necessary

cscheffauer (changes requested):

NotNestor:
build

cscheffauer (approved):

PR #1954, opened 2026-06-20
fix(listbox): keep an option focusable when the focused option is removed

jor-row:
### Description

When a consumer filters an `mdc-listbox` by removing `mdc-option` elements from the DOM (e.g. a search field that narrows the list), the option that currently holds the roving `tabindex="0"` can be the one that gets removed. Nothing re-assigned `tabindex="0"` to a surviving option, so **every** option was left at `tabindex="-1"` and the listbox became unreachable by keyboard (Tab skipped it entirely).

This wires the existing `ElementStore` `onStoreUpdate` hook in `ListBox`:

- When an option is `removed` from the store and it held `tabindex="0"`, a microtask (deferred so it runs after the store cache is spliced) re-assigns `tabindex="0"` to the option that takes its place, clamped to the new bounds, via `resetTabIndexes(newIndex, false)`.
- Focus is **not** moved (`focusElement = false`), which matters because filtering typically happens while focus is in an external input (e.g. a `Searchfield`).
- It is idempotent: if a surviving option already holds `tabindex="0"` (e.g. several options removed at once), it no-ops.

#### Testing

- Added an e2e step in `listbox.e2e-test.ts` — `keeps a focusable option when the focused option is removed` — that tabs into the listbox, removes the focusable option, asserts a remaining option receives `tabindex="0"`, and confirms the list is still reachable via Tab. Passing locally (`--project=chrome`).
- Manually validated in a downstream consumer 

### Links

_N/A_


packages/components/src/components/listbox/listbox.component.ts
```diff
@@ -121,6 +122,35 @@ class ListBox extends ListNavigationMixin(CaptureDestroyEventForChildElement(Com
...
+   * option would be focusable, making the listbox unreachable by keyboard. When
+   * that happens we re-assign `tabindex="0"` to the option that took the removed
+   * one's place (clamped to the new bounds) without moving focus, so a consumer
+   * filtering via an external input keeps focus where it is.
+   *
+   * The callback runs before the item is spliced out of the store, so the
+   * re-assignment is deferred to a microtask to act on the updated item list.
+   * @internal
+   */
+  private handleStoreUpdate(item: Option, changeType: ElementStoreChangeTypes, index: number): void {
```
cscheffauer: issue: this is introducing a new logic, instead we should do the exact same as in the List component:
https://github.com/momentum-design/momentum-design/blob/main/packages/components/src/components/list/list.component.ts#L106
even worth considering creating a Mixin, since the VirtualizedList has something in that regard too.

cscheffauer (changes requested):

packages/components/src/components/listbox/listbox.component.ts
jor-row: Good call — done in af1b87d7784. I dropped the bespoke microtask logic and reused the same synchronous [onElementStoreUpdate](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) pattern as List/TabList: on removed with [item.tabIndex === 0](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), reassign to a neighbour ([index + 1](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), falling back to [index - 1](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) for the last option) via [resetTabIndexes(newIndex, false)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). Acting pre-splice means no queueMicrotask is needed, and the existing e2e step still passes.

On the Mixin: agreed it's worth doing given List, TabList, VerticalTabList and VirtualizedList all carry variants of this. I'd prefer to keep this listbox bugfix scoped and extract the shared mixin in a dedicated follow-up so the List/TabList/VirtualizedList refactor gets its own review — happy to raise the ticket. WDYT?

cscheffauer (approved):

PR #1964, opened 2026-07-03
fix: datepicker range placeholder

NotNestor:
### Description

Updates range and week date pickers to display a locale-aware range placeholder instead of a single-date placeholder. For example, dd / mm / yyyy now displays as dd / mm / yyyy - dd / mm / yyyy. Single-date and manual-entry placeholders remain unchanged. Includes E2E coverage and updated visual regression snapshots.

### Breaking Change

N/A


maxinteger (changes requested):
The code looks Ok, but I found some issues during testing:
1. When I open the [DatePicker Exampele](http://localhost:6006/?path=/story/components-datepicker--example) and change the `selection-mode` to `week` nothing happens. I still can select only days. If I select `range` first and then `week` it works.
2. E2E screenshots naming is very confusing. Looks like the ones which has `default` in its name has a `range` selection, but not all represents the date in ranges for example "mdc-datepicker-default-variant-value-rtl.png" has a "Date range" label but shows a single date picker. Please clean up the screenshots and reflect the selection mode in the filename
3. Nitpick, In "[Localized German](http://localhost:6006/?path=/story/components-datepicker--localized-german) the date format is reverse, it should be `yyyy.mm.dd.`.

NotNestor:
> The code looks Ok, but I found some issues during testing:
> 
> 1. When I open the [DatePicker Exampele](http://localhost:6006/?path=/story/components-datepicker--example) and change the `selection-mode` to `week` nothing happens. I still can select only days. If I select `range` first and then `week` it works.
> 2. E2E screenshots naming is very confusing. Looks like the ones which has `default` in its name has a `range` selection, but not all represents the date in ranges for example "mdc-datepicker-default-variant-value-rtl.png" has a "Date range" label but shows a single date picker. Please clean up the screenshots and reflect the selection mode in the filename
> 3. Nitpick, In "[Localized German](http://localhost:6006/?path=/story/components-datepicker--localized-german) the date format is reverse, it should be `yyyy.mm.dd.`.

Fixed issue 1 and 2.

point 3 is incorrect German local should be dd.mm.yyyy

maxinteger (approved):

PR #1967, opened 2026-07-06
fix(CI): replace cache actions with upload/download artifact actions

maxinteger:
### Description

It is necessary because:
- upload/download is the right action to share artifacts between jobs
- GitHub made cache read-only in some cases which block our CI in PRs https://github.blog/changelog/2026-06-26-read-only-actions-cache-for-untrusted-triggers/

Breaking change: none

### DIFF
--- a/.github/workflows/deploy-package.yml
+++ b/.github/workflows/deploy-package.yml
@@ -72,11 +72,12 @@ jobs:
       - name: Build Automation
         run: yarn telemetry build && yarn automation build:cli
 
-      - name: Cache Automation
-        uses: actions/cache@v5
+      - name: Upload Automation
+        uses: actions/upload-artifact@v4
         with:
+          name: automation-dist-${{ env.rid }}
           path: '**/automation/dist'
-          key: automation-dist-${{ env.rid }}
+          retention-days: 1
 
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-package.yml
```diff
@@ -102,22 +103,23 @@ jobs:
...
-          path: '**/automation/dist'
-          key: automation-dist-${{ env.rid }}
+          name: automation-dist-${{ env.rid }}
+          path: .
 
-      - name: Build Automation (cache miss fallback)
-        if: steps.uncache-automation.outputs.cache-hit != 'true'
+      - name: Build Automation (artifact miss fallback)
+        if: steps.download-automation.outcome == 'failure'
         run: yarn telemetry build && yarn automation build:cli
```
maxinteger: @cscheffauer do we need this? The artifact must exists except when automation build failed, but then no point to re run it here

cscheffauer (dismissed):

.github/workflows/deploy-package.yml
cscheffauer: automation package needs to exist before other steps run. like its necessary to get the package-matrix et.

PR #1974, opened 2026-07-08
docs: add engineer guide for knowledge base onboarding

morgan1bathe:
## Summary
- Adds `knowledge-base/engineer-guide.md`, a Tier 1 onboarding guide for engineers covering KB tiers, folder layout, index lookup, contributing workflow, and CI enforcement.
- Updates `knowledge-base/index.root.json` so the new topic is indexed (required for `yarn knowledge-base:validate`).

## Test plan
- [x] `yarn knowledge-base:validate` passes locally
- [ ] Review engineer guide content for accuracy
- [ ] Confirm links resolve on GitHub after merge


knowledge-base/engineer-guide.md
```diff
@@ -0,0 +1,129 @@
+---
```
cscheffauer: issue: this is not customer / user facing, but for contributors of the repo itself.
we should not put this here

cscheffauer (changes requested):

knowledge-base/engineer-guide.md
cscheffauer: we should convert this into a KNOWLEDGE_BASE.md file at the top of the repo or similar, instead of putting it into the knowledge-base/ folder. The knowledge-base folder is meant to be processed by tooling and all the documents in those are going to be outputted to the Website and/or MCP. This knowledge should not be on the website nor the MCP, since the scope of those output formats is not meant to be for contributing engineers.
Contributing engineers can use agentic skills we add and Markdown files in the repo itself (outside of knowledge-base/ folders)

PR #1985, opened 2026-07-21
feat(component): add focus trap

maxinteger:
### Description

Implement Focus trap component

FocusTrap is a container component that manages keyboard focus within a specified region. It prevents focus from moving outside the container via Tab/Shift+Tab and optionally restores focus to the previously focused element when the trap is deactivated. Commonly used in modals, dialogs, popovers, and other overlay patterns to improve accessibility and user experience. 
This component is in addition to the existing dialog, popover components, which have the focus trap already built in.

breaking change: none

### DIFF
--- a/packages/components/src/components/focustrap/focustrap.component.ts
+++ b/packages/components/src/components/focustrap/focustrap.component.ts
@@ -23,28 +23,28 @@ class FocusTrap extends FocusTrapMixin(Component) {
    *
    * @default false
    */
-  @property({ type: Boolean, reflect: true })
-  disabled: boolean = DEFAULTS.DISABLED;
+  @property({ type: Boolean, reflect: true, attribute: 'trap-disabled' })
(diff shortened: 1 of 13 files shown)


packages/components/src/components/focustrap/focustrap.component.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+class FocusTrap extends FocusTrapMixin(Component) {
+  /**
+   * When `true`, disables keyboard focus trapping inside this component.
+   * When `false` (default), focus will be trapped.
+   *
+   * @default false
+   */
+  @property({ type: Boolean, reflect: true })
+  disable: boolean = DEFAULTS.DISABLE;
```
cscheffauer: it should be disabled, following all other components naming

cscheffauer (changes requested):

cscheffauer (approved):

PR #1990, opened 2026-07-24
fix(presence): presence component update

tricarro:
### Description

Locked sizing of avatar presence icon to rem to fix rendering issue in some products. 

I noticed that in a product consuming Momentum, the presence icon was being rendered incorrectly at certain sizes. I tracked the bug to be an alignment of the size of the icon to "em" instead of "rem". 

I locked the size at the component level to "rem" so it wouldn't inherit an "em" value from its parent accidentally.


tricarro:
I've added a test to the .e2e-test.ts file. I don't have enough experience to understand what that is doing or how to write a good test. Would appreciate extra care reviewing that.

PR #1991, opened 2026-07-25
feat: Add Momentum Design MCP server

jenet1410:
## Description

This PR adds the initial Momentum Design MCP server under the new `@momentum-design/components-mcp` workspace.

The MCP server helps AI agents discover and use Momentum Design information from the public monorepo sources. It generates static registries from component knowledge-base files, component source metadata, public asset packages, and color token files.

Covered capabilities include:

- Listing and searching Momentum Design components
- Reading component knowledge-base documentation
- Inspecting component APIs, including properties, attributes, defaults, allowed values, slots, events, CSS parts, and accessibility rules
- Generating component usage examples for Web Component, Angular, and React
- Validating component usage against documented API and accessibility rules
- Listing and searching public icons, illustrations, animations, and brand visuals
- Looking up color tokens, CSS variables, and resolved color values
- Suggesting migration paths from Momentum UI component/icon usage to Momentum Design

The package also includes usage documentation with example prompts and framework-specific guidance.

## Package Structure

Momentum MCP package structure

## How the flow works:

Momentum MCP list components source flow

## Build and Configure Locally

From the Momentum repo root:

```sh
cd /path/to/momentum-design
yarn install
yarn components-mcp build
```

After the build, the MCP server entry file should be:

```txt
/path/to/momentum-design/packages/tools/components-mcp/dist/module/index.js
```

### VS Code / Cursor MCP Configuration

Add this MCP server configuration:

```json
{
  "servers": {
    "momentum-design": {
      "command": "/path/to/node",
      "args": [
        "/path/to/momentum-design/packages/tools/components-mcp/dist/module/index.js"
      ]
    }
  }
}
```

If `node` is available in the editor environment, `"command": "node"` can also be used.

## Example Usage 1:
List all Momentum Design components.

https://github.com/user-attachments/assets/3572e82d-9a08-4f22-80bc-8f99afdb40d8

## Example Usage 2:
Show Button component API with all details.

https://github.com/user-attachments/assets/a8014f77-c255-44b0-9d23-835b55b544bf

## Example Usage 3:
Generate React code for a primary Button with size 32, prefix icon download-bold, and text Download.

https://github.com/user-attachments/assets/82d4fdaf-7172-42ac-9620-a4ee60c57fe2

## Example Usage 4:
Search Momentum icons for success and show me how to use the best matching icon.

https://github.com/user-attachments/assets/c5d7294b-aabd-48bb-9a66-877f41cae763

## Example Usage 5:
What is the right Momentum color token for #0070D2?

https://github.com/user-attachments/assets/2d475745-5abe-427a-bc80-6a1526be1a4e

## To know more about usage:
For more details about supported workflows, example prompts, and framework-specific usage, see `USAGE.md`.

## Validation

- Ran `yarn components-mcp test`
- Ran `yarn components-mcp build`
- Ran `yarn components-mcp analyze`
- Ran `git diff --check`
- Verified local MCP startup through VS Code/MCP configuration
- Verified `npm pack --dry-run`
- Checked authored files for private URLs, tokens, local paths, or internal-only documentation

## Breaking Change

No breaking changes.


packages/tools/components-mcp/scripts/generate-asset-registry.js
```diff
@@ -0,0 +1,144 @@
+const fs = require('node:fs');
+const path = require('node:path');
+
+const OUTPUT_PATH = path.resolve(__dirname, '../src/data/assets.registry.ts');
+const REPO_ROOT = path.resolve(__dirname, '../../../..');
+
+const ASSET_SOURCES = [
+  {
+    kind: 'icon',
```
cscheffauer, line 9: issue: we do have a manifest.json file in each asset package dist folder - there is no need to go through all assets, instead just consume the existing json files. This needs to be checked if we want to generate a registry json file for each package before this PR. Then in the MCP we would just get the data directly from the node_modules (instead of ingesting the data during build time).

packages/tools/components-mcp/scripts/generate-color-registry.js
```diff
@@ -0,0 +1,190 @@
+const fs = require('node:fs');
+const path = require('node:path');
+
+const REPO_ROOT = path.resolve(__dirname, '../../../..');
+const TOKENS_ROOT = path.resolve(__dirname, '../../../assets/tokens/src');
+const OUTPUT_PATH = path.resolve(__dirname, '../src/data/colors.registry.ts');
+
+const STATIC_TOKEN_SOURCES = [
```
cscheffauer, line 8: issue: this gets the data from the src folder. We have to point the output (including MCP) to the dist folders, not the src.
check the different output formats.

packages/tools/components-mcp/src/common/constants.ts
```diff
@@ -0,0 +1,9 @@
+const CONSTANTS = {
+  PACKAGE_NAME: '@momentum-design/components-mcp',
+  SERVER_NAME: 'momentum-design-components-mcp',
+  SERVER_VERSION: '0.1.0',
```
cscheffauer, line 4: for what is this used and why? also why is there a server version hard coded? this needs to be automatically updated as part of our release history if we want this to be used

packages/tools/components-mcp/scripts/generate-asset-registry.js
```diff
@@ -0,0 +1,144 @@
+const fs = require('node:fs');
+const path = require('node:path');
+
+const OUTPUT_PATH = path.resolve(__dirname, '../src/data/assets.registry.ts');
+const REPO_ROOT = path.resolve(__dirname, '../../../..');
```
cscheffauer, line 5: issue: paths should not be relative, but use `@momentum-design/...` imports. Add the peerDependencies in the package.json, same as we do it in other packages.

packages/tools/components-mcp/src/tools/getComponentApi.ts
```diff
@@ -0,0 +1,63 @@
+import { requireComponent } from '../utils/componentRegistry';
+import { formatProperty, formatReferenceList } from '../utils/formatters';
+import { textResponse } from '../utils/response';
+
+import type { ComponentsMcpTool } from '../utils/toolTypes';
+import { getComponentNameArg } from '../utils/toolUtils';
+
+const getComponentApiTool: ComponentsMcpTool = {
```
cscheffauer, line 8: instead of just saving the data in the "data" folder, use the MCP "Resources" feature:
https://modelcontextprotocol.io/specification/2025-11-25/server/resources

packages/tools/components-mcp/src/tools/getComponentExamples.ts
```diff
@@ -0,0 +1,87 @@
...
+        `## ${item} Example`,
+        '',
+        ...result.setupNotes.map(note => `- ${note}`),
+        '',
+        `\`\`\`${item === 'react' ? 'tsx' : 'html'}`,
+        result.code,
+        '```',
+      ].join('\n');
+    });
+    const documentationExamples = getDocumentationExamples(framework, component.examples);
```
cscheffauer, line 76: where is this setup? we do not have this atm... also this should be maintainable.

packages/tools/components-mcp/USAGE.md
```diff
@@ -0,0 +1,447 @@
...
+- Find public Momentum Design icons, illustrations, animations, and brand
+  visuals.
+- Look up color tokens, CSS variables, and resolved color values.
+- Suggest migration paths from Momentum UI usage to Momentum Design usage.
+
+## Current Coverage
+
+The generated registry currently covers:
+
+- 96 Momentum Design components.
```
cscheffauer, line 26: issue: how is this updated? like these are hard-coded numbers, if we add components / assets etc. this Readme gets out of date. Either automate the generation of this or remove the hard coded data.

cscheffauer (changes requested):
first initial review

packages/tools/components-mcp/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@momentum-design/components-mcp",
```
cscheffauer, line 2: issue: the folder name, package, etc. should be: "@momentum-design/mcp"

cscheffauer (changes requested):

packages/tools/components-mcp/scripts/generate-asset-registry.js
```diff
@@ -0,0 +1,144 @@
+const fs = require('node:fs');
```
cscheffauer, line 1: issue: the MCP server should consider locally installed versions of all packages (in the consumers package.json). This approach by ingesting the data into the MCP will work for the latest version, but not if locally installed versions differ

cscheffauer (changes requested):

package.json
```diff
@@ -12,6 +12,7 @@
   ],
   "scripts": {
     "analyze:dependencies": "yarn syncpack lint",
+    "components-mcp": "yarn workspace @momentum-design/components-mcp",
```
cscheffauer, line 15: we would want to order this alongside with the other workspaces at the bottom of the scripts

packages/tools/components-mcp/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "type": "commonjs",
+  "main": "./dist/module/index.js",
+  "types": "./dist/types/index.d.ts",
+  "files": [
+    "./dist",
+    "./USAGE.md"
+  ],
+  "bin": {
+    "components-mcp": "./dist/module/index.js",
+    "momentum-components-mcp": "./dist/module/index.js"
```
maxinteger, line 15: Do we need 2 names for the same mcp?

Also, it covers all tokens and assets not just components so the name should be 
```suggestion
    "momentum-mcp": "./dist/module/index.js",
```

packages/tools/components-mcp/package.json
maxinteger: for better consistency the `pakages/tools/components-mcp` should be renamed to `packages/tools/mcp`

packages/tools/components-mcp/USAGE.md
maxinteger: Please remove the hardcoded numbers, they do not have real values in this documentation

packages/tools/components-mcp/scripts/generate-asset-registry.js
maxinteger: @cscheffauer what do you mean? Using `node:` prefix is the preferred way to import code modules and it is supported since node 16.

packages/tools/components-mcp/scripts/generate-asset-registry.js
```diff
@@ -0,0 +1,144 @@
...
+    entry.format,
+    entry.packageName,
+    path.basename(entry.sourcePath),
+  ]
+    .join(' ')
+    .toLowerCase()
+    .split(/[^a-z0-9]+/)
+    .filter(Boolean)
+    .filter((keyword, index, all) => all.indexOf(keyword) === index)
+    .sort();
```
maxinteger, line 82: This looks a bit over engineered


```suggestion
  return [
    entry.name,
    entry.kind,
    entry.category,
    entry.format,
    entry.packageName,
    path.basename(entry.sourcePath),
  ])
    .filter(Boolean)
    // Lower case and replace spaces
    .map(kw => kw.toLowerCase().replaceAll().replaceAll(/\s+/g, '-'))
    // filter out duplicates
    .filter((keyword, index, all) => all.indexOf(keyword) === index)
    .sort();
```

packages/tools/components-mcp/scripts/generate-asset-registry.js
```diff
@@ -0,0 +1,144 @@
...
+  const entries = ASSET_SOURCES.flatMap(sourceConfig => (
+    walkFiles(sourceConfig.rootPath)
+      .filter(filePath => sourceConfig.extensions.includes(path.extname(filePath)))
+      .map(filePath => buildEntry(sourceConfig, filePath))
+  )).sort((a, b) => (
+    a.kind.localeCompare(b.kind)
+    || a.category.localeCompare(b.category)
+    || a.name.localeCompare(b.name)
+  ));
+  const jsonChunks = JSON.stringify(entries, null, 2)
```
maxinteger, line 122: I think, no point to format the JSON here

```suggestion
  const jsonChunks = JSON.stringify(entries)
```

packages/tools/components-mcp/scripts/generate-asset-registry.js
```diff
@@ -0,0 +1,144 @@
...
+    .match(/[\s\S]{1,16000}/g)
+    .map(chunk => JSON.stringify(chunk))
+    .join(',\n  ');
```
maxinteger, line 125: Is this really necessary? Is 16000 character long line a NodeJS limitation?

packages/tools/components-mcp/scripts/generate-asset-registry.js
```diff
@@ -0,0 +1,144 @@
...
+
+  const output = `// Generated by scripts/generate-asset-registry.js. Do not edit manually.
+import type { AssetRegistryEntry } from './registry.types';
+
+const assetsRegistryJson: string = [
+  ${jsonChunks},
```
maxinteger, line 131: Interestingly `generate-component-registry.js` just inject the JSON formatted structure directly into the code.

Please make it consistent and use only one way to inject data.

I prefer the way you do it for components registry:
`...= ${JSON.stringify(entries, null, 2)};`

Why did you do the `jsonChunks` instead of the direct injection?

packages/tools/components-mcp/scripts/generate-component-registry.js
```diff
@@ -0,0 +1,445 @@
...
+}
+
+function toPascalCase(value) {
+  return value
+    .replace(/([a-z0-9])([A-Z])/g, '$1 $2')
+    .split(/[\s-_]+/)
+    .filter(Boolean)
+    .map(part => `${part.charAt(0).toUpperCase()}${part.slice(1)}`)
+    .join('');
+}
```
maxinteger, line 27: It is Ok, to add utility packages like `lodash` to use well known function instead of re-implement them

packages/tools/components-mcp/src/common/types.ts
```diff
@@ -0,0 +1,118 @@
...
+  componentPath: string;
+  knowledgeBasePath: string;
+}
+
+export interface ComponentRegistryEntry {
+  name: string;
+  title: string;
+  tagName: string;
+  summary: string;
+  tier?: number;
```
maxinteger, line 76: `tier` is for knowledge base, it is always `3` for components. We is it part of the entry?

packages/tools/components-mcp/src/common/types.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export interface ComponentProperty {
+  name: string;
+  attribute?: string;
+  type?: string;
+  default?: RegistryValue | string;
+  values?: RegistryValue[];
+  description?: string;
+  reflected?: boolean;
+  source?: 'component' | 'storybook';
```
maxinteger, line 49: All properties must be defined by the component and it should be the only source of truth. Do you have an example where the `storybook` has to be used?

packages/tools/components-mcp/scripts/generate-component-registry.js
```diff
@@ -0,0 +1,445 @@
+const fs = require('node:fs');
+const path = require('node:path');
+
+const COMPONENTS_ROOT = path.resolve(__dirname, '../../../components/src/components');
+const OUTPUT_PATH = path.resolve(__dirname, '../src/data/components.registry.ts');
+
+const SOURCE_EXTENSION_PATTERN = /\.(component|constants|types|stories)\.ts$/;
```
maxinteger, line 7: issue:

Why do you need all these sources?

We already have all the informations in the `packages/components/dist/custom-elements.json` It resolves all the inherited props and it also contains the public methods.

Probably the only missing bit is the `dependencies`, but you can get is from the component's `jsDoc` field

packages/tools/components-mcp/src/common/types.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export interface ComponentApi {
```
maxinteger, line 52: issue: please add public methods as well

packages/tools/components-mcp/src/common/types.ts
```diff
@@ -0,0 +1,118 @@
...
+  tagName: string;
+  summary: string;
+  tier?: number;
+  source: ComponentSourceReference;
+  docs: string;
+  sections: Record<string, string>;
+  dependencies: string[];
+  api: ComponentApi;
+  examples: ComponentExample[];
+  reactImports: string[];
```
maxinteger, line 83: nitpick: this is not scalable, do this instead:

```suggestion
  imports: Record<string, string[]>;
```

out put:
```js
{
...
imports: {
   react: ['....']
}
...
```

packages/tools/components-mcp/src/utils/validation.ts
maxinteger, line 1: issue: What are we validating here?

Looks like there are a lot of hardcoded component constants, but I think it is pointless to check them one by one.

If you want to test the output of the MDC tools, then you should validate agains the source which you used, for example in the component registry case: `packages/components/dist/custom-elements.json`. Also the validation/test should focus more on the part which was added by the tool, which is basically a mapping between the previously generated JSON files and the JSON which coming out from MCP.

This validation is very fragile because the hard coded values and it fails even when nothing changed in the MDC output structure.

maxinteger (changes requested):
issue: Why are the generated files under `/src/data/` committed?

- None MDC developers should use NPM 
- MDC developers can run yarn script to start MDC and it can auto generate these files


I stopped reviewing because the PR contains a lot of tooling to parse source code instead of use the already parsed data as input, like 
- packages/components/dist/custom-elements.json
- packages/assets/animations/dist/manifest.json
- packages/assets/tokens/dist/json/core/complete.json
- etc.

Please use these as a base and add more metadata if you need.

PR #1992, opened 2026-07-27
fix: staticchip add prefix slot for avatar support

NotNestor:
### Description

Allow slotted leading content (e.g. mdc-avatar) in static chips,
matching the prefix slot pattern already established in mdc-inputchip.
Slot content takes precedence over icon-name when both are provided.

### Breaking Change

None


packages/components/src/components/staticchip/knowledge-base/staticchip.component.md
```diff
@@ -40,6 +40,14 @@ Minimal markup example:
 <mdc-staticchip label="Done" color="mint"></mdc-staticchip>
 ```
 
+With an avatar slotted as the prefix (takes precedence over `icon-name`). Use `size="24"` on the avatar so the initials text scales appropriately for the chip height:
```
cscheffauer, line 48: suggestion: i think this section would be better suited in the Content guidance

cscheffauer (dismissed):
1 comment, which would be great to address before merge

cscheffauer (approved):

PR #1996, opened 2026-07-31
fix(iconProvider): updating default sizing to rem instead of em

tricarro:
### Description

Investigating an icon rendering issue in a product consuming Momentum. Found that the default sizing of the icon provider is set to "em". This breaks our sizing conventions and architecture in Momentum, everything is based on "rem".

### Breaking Change

This is a breaking change.

### DIFF
--- a/packages/components/src/components/iconprovider/iconprovider.constants.ts
+++ b/packages/components/src/components/iconprovider/iconprovider.constants.ts
@@ -14,7 +14,7 @@ const LENGTH_UNIT_SIZE = {
 const DEFAULTS = {
   FILE_EXTENSION: 'svg',
   LENGTH_UNIT: 'rem',
-  SIZE: LENGTH_UNIT_SIZE.em,
+  SIZE: LENGTH_UNIT_SIZE.rem,
   ICON_SET: 'momentum-icons',
 } as const;
 
--- a/packages/components/src/components/iconprovider/iconprovider.constants.ts
+++ b/packages/components/src/components/iconprovider/iconprovider.constants.ts
@@ -13,7 +13,7 @@ const LENGTH_UNIT_SIZE = {
 
 const DEFAULTS = {
   FILE_EXTENSION: 'svg',
-  LENGTH_UNIT: 'em',
+  LENGTH_UNIT: 'rem',
   SIZE: LENGTH_UNIT_SIZE.em,
   ICON_SET: 'momentum-icons',
 } as const;
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,6 @@ figma-webcomponent.config.json
 
 # copilot plan mode
 plan.md
+
+# Context Documentation
+.docs/*
--- a/packages/components/src/components/iconprovider/iconprovider.stories.ts
+++ b/packages/components/src/components/iconprovider/iconprovider.stories.ts
@@ -62,7 +62,7 @@ export const Example: StoryObj = {
     'icon-set': 'momentum-icons',
     url: './icons/svg',
     'file-extension': 'svg',
-    'length-unit': 'em',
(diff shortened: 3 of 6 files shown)


tricarro:
Screenshot 2026-07-31 at 10 50 43 AM
Example of icon rendering issue within the presence component.

Screenshot 2026-07-31 at 10 51 20 AM
Example of the icon rendering after this fix.

jor-row:
Some failing tests:

```mdc-icon```, ```attributes``` test checks for ```em``` explicitly so that needs updated


This change has changed the size of icons inside of all buttons as well, is that intentional?
Screenshot 2026-07-31 at 22 39 14

```mdc-iconprovider``` snapshot tests (standalone and nested) explicitly test for ```em``` as well, so those need updated

File: packages/components/src/components/iconprovider/iconprovider.constants.ts [hunk 1]
```diff
@@ -13,7 +13,7 @@ const LENGTH_UNIT_SIZE = {
 
 const DEFAULTS = {
   FILE_EXTENSION: 'svg',
-  LENGTH_UNIT: 'em',
+  LENGTH_UNIT: 'rem',
   SIZE: LENGTH_UNIT_SIZE.em,
```
maxinteger: Please update the `size` property as well.
```suggestion
  SIZE: LENGTH_UNIT_SIZE.rem,
```

maxinteger (changes requested):
the `em` value mentioned couple of other places, please update all:

- packages/components/src/components/iconprovider/iconprovider.component.ts line 74
- packages/components/src/components/icon/knowledge-base/icon.component.md line 64
- packages/components/src/components/iconprovider/knowledge-base/iconprovider.component.md line 72

Also, progress spinner set the icon length unit to `em` here: packages/components/src/components/progressspinner/progressspinner.constants.ts line 13. That should be change do `rem` as well.

jor-row:
Please note in the PR description that this is a breaking change

tricarro:
> Some failing tests:
> 
> `mdc-icon`, `attributes` test checks for `em` explicitly so that needs updated
> 
> This change has changed the size of icons inside of all buttons as well, is that intentional? Screenshot 2026-07-31 at 22 39 14
> 
> `mdc-iconprovider` snapshot tests (standalone and nested) explicitly test for `em` as well, so those need updated

@jor-row, that's not intentional. And, it's making me wonder if I need to override the button's size to "em" to align them with the type size, or just update the size of the presence icon to "rem" instead of the iconProvider?

tricarro:
**EM Components**
The only components in Momentum using "em" as the default spacing are IconProvider and ProgressSprinner. After investigating further, I'm pretty sure that this default sizing is a mistake and is out of alignment with how sizing is built for all the other components in our library.

Because of that, I've made a few changes that I've made to this PR:
• Updated the default size of the Progress Spinner component. 
• Updated knowledge base file for IconProvider.
• Run local testing for icon impacts. I was unable to find a discernable difference between the old and new components.


Actual
Expected

Screenshot 2026-08-17 at 3 23 53 PM

jor-row:
The snapshots are still failing on the buttons with icons on the 2 smallest sizes. We are expecting much smaller icons in those buttons those in this pr
Screenshot 2026-08-18 at 22 21 46


in iconprovider.e2e-test.ts, we need to update the snapshots because the text has been updated from em to rem

the attributes tests in icon.e2e-test.ts need fixed, as they currently check for 
```await expect(icon).toHaveAttribute('style', '--computed-icon-size: 1em;');```

jor-row:
in the storybook, comparing the buttons with icons, i see no difference. im not sure why the e2e test snapshots have small icons

we just need to update the snapshots (i think there is a github action to do that?)

PR #1999, opened 2026-08-03
docs: adding fonts guideline

tricarro:
### Description

Adding a guidelines document for fonts.

### Breaking Change

This is not a breaking change.


packages/assets/fonts/knowledge-base/fonts.md
```diff
@@ -0,0 +1,95 @@
...
+
+Pin the version to match the rest of your Momentum dependencies. See [CONSUMPTION.md](../../../../CONSUMPTION.md) for general npm consumption patterns.
+
+## Loading the font
+
+Import the provided stylesheet so the `@font-face` rule is applied before Momentum typography renders:
+
+```javascript
+import '@momentum-design/fonts/dist/css/fonts.css';
+```
```
maxinteger: As I mentioned in your typography pr (#2000) you also need a bundler to make the JS import work.

Again I recommend to use `<link />` tag:
```html
<link rel="stylesheet" href="https://unpkg.com/@momentum-design/fonts/dist/css/fonts.css">
```

and probably mention the npm package as well.

Also there are more then 2 Node package manager on the market, so better to mention the package name only and the consumer will pass it to the right command

packages/assets/fonts/knowledge-base/fonts.md
```diff
@@ -0,0 +1,95 @@
...
+
+Example manifest:
+
+```json
+{
+  "fonts": "./css/fonts.css",
+  "Inter": "./inter/Inter.var.woff2",
+  "manifest": "./manifest.json"
+}
+```
```
maxinteger: I think this is unnecessary and hard to maintain section of the doc

Instead just add a link to the NPM page code section which list the content of the package:

https://www.npmjs.com/package/@momentum-design/fonts?activeTab=code

packages/assets/fonts/knowledge-base/fonts.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Licensing
+
+The `@momentum-design/fonts` package is released under the [MIT License](../LICENSE.md). Inter is bundled as a third-party typeface; retain license notices shipped with the package when redistributing font files.
```
maxinteger: Do we need this in the knowledge base. If yes, then we should mention in the root  only and not in every single page.

maxinteger (changes requested):

packages/assets/fonts/knowledge-base/fonts.md
tricarro: I don't know. This was one of the things that AI wrote as part of Christoph's skill. But I don't see it in other knowledge base documents. I'll remove it.

packages/assets/fonts/knowledge-base/fonts.md
tricarro: I'll remove it.

packages/assets/fonts/knowledge-base/fonts.md
tricarro: Ok. For the package manager, I've changed the instruction a bit, but kept the npm and yarn examples for AI.

I've updated the Loading the font section to use a Link element.

maxinteger: AI should find out which packages manager must be used based on the project config (package.json)

The package name should be sufficient for both AI and human.

Since it is not a show stopper I approve it and leave you the decision :)

maxinteger (dismissed):

PR #2007, opened 2026-08-06
refactor(components): migrate Figma Code Connect from parsers to template files

morgan1bathe:
### Description

Migrate all Code Connect files from the deprecated framework-specific parser format (React/HTML) to the new framework-agnostic template file format, ahead of Figma dropping parser support on 2026-08-17.

- Convert 14 *.react.figma.tsx (React parser) and 14 *.webcomponent.figma.ts (HTML parser) into template files exporting default { id, imports, example }
- Preserve <FIGMA_*_URL> placeholders and documentUrlSubstitutions so the .env-based URL mapping keeps working unchanged
- Drop parser/importPaths/paths from the config factories; add language
- Update plop scaffolding templates to emit template files
- Fix code-connect file path in getFigmaComponentAttributes.js
- Bump @figma/code-connect 1.3.4 -> 1.5.2 (adds migrate command / template API)

### Breaking Change

None


packages/components/src/components/radio/code-connect/radio.webcomponent.figma.ts
```diff
@@ -1,38 +1,11 @@
...
-    }),
-  },
-  example: () => html`<mdc-radio></mdc-radio>`,
+import figma from "figma"
+
+export default {
+  id: "radio",
   imports: ["import '@momentum-design/components/components/radio';"],
-});
+  example: figma.code`<mdc-radio></mdc-radio>`,
```
maxinteger, line 9: question: Why did you kept the props for the react version but not for the webcomponent?

morgan1bathe: This was an already existing gap before working on this migration. Those props were declared but never referenced in the snippet, so the published Web Component snippet has always rendered a bare <mdc-radio></mdc-radio> regardless of the Figma properties. My migration just removed those now-dead props/__props declarations. The rendered output is the same as before.

I was hoping for this PR to include no behavior changes and tackle the wiring of props in a later PR.

I'm also fumbling around a bit with the process so I appreciate your feedback!

PR #2010, opened 2026-08-07
docs(tokens): add motion knowledge-base guideline

morgan1bathe:
### Description

Tier 2 guideline for the motion tokens: the shipped core scales (duration, easing, delay, stagger), the animation recipe layer and its four types, principles and rules for applying motion, and usage guidance. Shipped token values are treated as the source of truth.

### Breaking Change

None


packages/assets/tokens/knowledge-base/motion.md
```diff
@@ -0,0 +1,147 @@
...
+
+How long an animation runs. Match the step to the interaction's scale.
+
+| Token | Custom property | Value | Use for |
+| --- | --- | --- | --- |
+| `motion.duration.instant` | `--mds-motion-duration-instant` | 100ms | Immediate change |
+| `motion.duration.fast` | `--mds-motion-duration-fast` | 200ms | Micro-interactions, hover states, opacity fades, small scale changes, toggles |
+| `motion.duration.normal` | `--mds-motion-duration-normal` | 300ms | Medium transitions — modals, accordions, drawers |
+| `motion.duration.relaxed` | `--mds-motion-duration-relaxed` | 450ms | Large / emphatic — full-screen modals, sidebars, page transitions |
+| `motion.duration.slow` | `--mds-motion-duration-slow` | 600ms | Deliberate attention — onboarding, celebrations, progress completion, complex entrances |
```
maxinteger: Please do not use actual tokens and its values in this documentation because they can change and can lead to discrepancies.

Focus on the concept instead, for example

```suggestion
| Token  | Use for |
| ---  | --- |
| instant | Immediate change |
| fast | Micro-interactions, hover states, opacity fades, small scale changes, toggles |
| normal | Medium transitions — modals, accordions, drawers |
| relaxed | Large / emphatic — full-screen modals, sidebars, page transitions |
| slow | Deliberate attention — onboarding, celebrations, progress completion, complex entrances |
```

please do the same for the other categories

packages/assets/tokens/knowledge-base/motion.md
```diff
@@ -0,0 +1,147 @@
...
+
+| Token | Custom property | Value |
+| --- | --- | --- |
+| `motion.stagger.tight` | `--mds-motion-stagger-tight` | 30ms |
+| `motion.stagger.normal` | `--mds-motion-stagger-normal` | 60ms |
+| `motion.stagger.relaxed` | `--mds-motion-stagger-relaxed` | 100ms |
+
+Authoritative source: `packages/assets/tokens/src/core/motion.json`.
+
+## Animation recipes
```
maxinteger: These are actual "Animations" and not recipes.

Consumer can access to them and use then in there on app just by referring to the name of the animation

maxinteger (changes requested):

packages/assets/tokens/knowledge-base/motion.md
morgan1bathe: Updated in latest commit

packages/assets/tokens/knowledge-base/motion.md
morgan1bathe: Updated in latest commit

PR #2014, opened 2026-08-10
docs: add Tier 1 knowledge base blueprints (accessibility, interaction, responsive, i18n)

morgan1bathe:
### Description

Add four design-system-wide Tier 1 topics — accessibility, interaction patterns, responsive design, and internationalization — establishing the cross-cutting "context engine" that component topics defer to. Targets WCAG 2.2 AA, US English, and regenerates index.root.json to include them.

### Breaking Change

None


knowledge-base/responsive-design.md
```diff
@@ -0,0 +1,77 @@
...
+
+Momentum does **not** detect screen size for you, and it does **not** prescribe breakpoints. Instead:
+
+- **You own breakpoint detection.** Decide device class from whatever signal fits your app — CSS media queries, container queries, or device detection — and choose the breakpoint values that suit your product.
+- **Components read a shared signal.** Rather than each component re-implementing breakpoint logic, they read a shared responsive context and adapt their behavior.
+- **You own page layout.** Grid, columns, and spacing are yours to build with standard CSS. Use flexible, relative units so the result reflows and zooms cleanly (see below).
+
+This keeps the system framework-agnostic and lets one set of components serve products with very different layout needs.
+
+## Shared responsive context
```
maxinteger, line 32: note: `responsivesettingsprovider` is created to enforce media (mobile, desktop, etc) and very special settings responsive design related settings on specific components.

Consumer should add it only when for example 
- they want to render "mobile" like UI on a big screen or
- render differently some smaller part of the app

Most of the app should not use it at all. The `rem` based design should handle most of the responsive use cases.

maxinteger (changes requested):
I think there is a miss-understanding around the spatial navigation provider.

Even though it can give a 4 direction navigation in a grid like layout, but it is just a side effect and not the main function of it.

First of all, you can not have more then one `mdc-spacialnavigation` provider in a document, so you can not wrap all your grids with one.

Spatial navigation is basically the third navigation option after mouse and keyboard. It is made for Remote and game controllers where you do not have tab key, instead you use 4 direction, enter and escape/back buttons.

I think this document mix up spatial navigation with another concept which is called focusgroup.

Component which use arrow keys to navigate (tabs, list, etc.) already implementing it, but the grid navigation still missing.

Some reference for focusgroups:
- https://open-ui.org/components/scoped-focusgroup.explainer/
- https://open-ui.org/components/focusgroup-v2.explainer/

PR #2017, opened 2026-08-14
fix(tools): fix hot module reload for storybook

maxinteger:
### Description

Rewrite import path in storybook to load the search instead of the last built version.

breaking change: none


packages/components/config/storybook/main.js
```diff
@@ -24,7 +75,12 @@ const config = {
     return mergeConfig(config, {
       // adding dynamic import to support dynamic icon import
       // in icon component
-      plugins: [dynamicImport({})],
+      plugins: [preferSourceComponentsInStorybook(), dynamicImport({})],
```
cscheffauer, line 78: question: will this impact the prod build? this change seems to be a dev build only change no?

cscheffauer (changes requested):

packages/components/config/storybook/main.js
maxinteger: good question, it should have effect on dev only, but I double check

maxinteger: @cscheffauer I run `yarn build` then used the index.html from `momentum-design/packages/documentation/dist` and both the site and storybook looked Ok

cscheffauer (approved):

PR #2019, opened 2026-08-18
docs(skills): add Momentum Concierge

morgan1bathe:
## Summary
- add the Momentum Concierge skill under `.github/skills`
- include its usage and capability documentation
- route approved knowledge-base feedback to `momentum-design/momentum-design`

## Test plan
- [x] Confirm the copied skill differs from the source only where feedback routing changed
- [x] Confirm no `webexdesign/momentum-skills` feedback references remain
- [x] Run `git diff --check`

Made with [Cursor](https://cursor.com)


knowledge-base/skills/momentum-concierge/README.md
```diff
@@ -0,0 +1,95 @@
+# Momentum Concierge
```
cscheffauer, line 8: suggestion: storing this skill in .github marks this as a internal skill for this repo. Instead we have to move this one into the knowledge-base i think and export it as part of the website. Lets keep this PR open until that structure is decided. Then lets get back to this PR again

cscheffauer (changes requested):

knowledge-base/skills/momentum-concierge/README.md
morgan1bathe: @cscheffauer Can you review these updates and make sure they match what you had in mind?

tricarro:
Reviewed the PR and am merging it.

Bypassing the review requirement because I still haven't been added to the Momentum Designers and given review permissions.

PR #2021, opened 2026-08-18
fix(stepper): synchronize inherited context before rendering

NotNestor:
### Description

Fix a rendering loop by synchronizing stepper item and connector context in willUpdate() instead of updated(). Add regression coverage confirming provider values are applied within a single update cycle.

### Breaking Change

No breaking change

### DIFF
--- a/packages/components/src/components/stepper/stepper.e2e-test.ts
+++ b/packages/components/src/components/stepper/stepper.e2e-test.ts
@@ -74,6 +74,55 @@ const takeScreenshot = async (componentsPage: ComponentsPage, orientation: Orien
   });
 };
 
+// AI-Assisted: Cover initial non-default context, direct child overrides, and subsequent provider updates.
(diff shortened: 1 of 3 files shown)


maxinteger:
@NotNestor the fix looks good, but I can not see what was the problem with the previous solution. Do you have a use case which I can check and verify your fix?

NotNestor:
> @NotNestor the fix looks good, but I can not see what was the problem with the previous solution. Do you have a use case which I can check and verify your fix?

I'm not sure if you have access to this. We previously had a similar issue with the popover component. Essentially in this environment updating the property in the updated function leads to an endless loop. This doesn't happen in the actual app.

Lit documentation recommends using will update for this kind of thing in general as it prevents additional re rendering

> Implement willUpdate() to compute property values that depend on other properties and are used in the rest of the update process.

maxinteger (approved):

PR #2022, opened 2026-08-19
fix(components): input label accessibility in safari

maxinteger:
### Description

Because WebKit’s accessibility engine builds its tree top-down, standard HTML <label for=""> references are handled differently than explicit ARIA tokens.

When Safari parses aria-labelledby="heading-id", it skips the native layout engine's for lookup mechanism entirely and maps the node explicitly in the Accessibility API. This is why adding both attributes ensures cross-browser coverage: the for attribute manages mouse-click focusing, while aria-labelledby ensures Safari's VoiceOver speaks the label aloud.

more details:
https://nolanlawson.com/2022/11/28/shadow-dom-and-accessibility-the-trouble-with-aria/

breaking change: none

### DIFF
--- a/packages/components/src/components/input/input.component.ts
+++ b/packages/components/src/components/input/input.component.ts
@@ -403,6 +403,7 @@ class Input
       ?readonly="${this.readonly}"
       ?required="${this.required}"
       type="${type}"
+      aria-labelledby="${DEFAULTS.HEADING_ID}"
       aria-describedby="${ifDefined(
         this.helpText ? FORMFIELD_DEFAULTS.HELPER_TEXT_ID : (this.dataAriaDescribedby ?? ''),
       )}"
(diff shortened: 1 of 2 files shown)


packages/components/src/components/input/input.component.ts
```diff
@@ -403,6 +403,7 @@ class Input
       ?readonly="${this.readonly}"
       ?required="${this.required}"
       type="${type}"
+      aria-labelledby="${DEFAULTS.HEADING_ID}"
```
cscheffauer, line 406: question: not every input has a label around it - will this break if the heading_id can't be resolved?

cscheffauer (changes requested):

packages/components/src/components/input/input.component.ts
maxinteger: it should not cause problem, but I updated the code anyway

maxinteger: Sorry I had to open another PR 
https://github.com/momentum-design/momentum-design/pull/2024

PR #2028, opened 2026-08-20
fix(popover): button tooltip does not remain visible when mouse is moved over prefix/postfix slot

dominiccarrington:
### Description

- mdc-popover: use compareDocumentPosition to check if an element is within the trigger element of a popover in order to count both the real and shadow DOM in the check

### Breaking Change

None


packages/components/src/components/button/button.stories.ts
```diff
@@ -283,3 +283,44 @@ export const PillWithPostfixSlot: StoryObj = {
...
+        aria-label="${ifDefined(args['aria-label'])}"
+        ariaStateKey="${ifDefined(args.ariaStateKey)}"
+        ?inverted="${args.inverted}"
+        style="${ifDefined(args.style)}"
+        ?auto-focus-on-mount="${args['auto-focus-on-mount']}"
+        id="button"
+      >
+        ${args.children}
+
+        <div slot="prefix">
```
cscheffauer, line 313: why slotting them? we can use the prefixIcon and postfixIcon prop instead

packages/components/src/components/button/button.stories.ts
```diff
@@ -283,3 +283,44 @@ export const PillWithPostfixSlot: StoryObj = {
...
+        @keyup="${action('onkeyup')}"
+        @focus="${action('onfocus')}"
+        ?active="${args.active}"
+        ?disabled="${args.disabled}"
+        ?soft-disabled="${args['soft-disabled']}"
+        variant="${ifDefined(args.variant)}"
+        size="${ifDefined(args.size)}"
+        color="${ifDefined(args.color)}"
+        type="${ifDefined(args.type)}"
+        role="${ifDefined(args.role)}"
```
cscheffauer, line 302: passing in all those args like role etc. doesn't really make sense imho

packages/components/src/components/button/button.e2e-test.ts
```diff
@@ -768,4 +768,50 @@ test.describe.parallel('mdc-button', () => {
...
     });
   });
+
+  test('mdc-button with tooltip stays visible when hovering over prefix/postfix icons', async ({ componentsPage }) => {
+    await componentsPage.mount({
+      html: `
+        <div id="wrapper">
+          <mdc-button id="button-with-tooltip">
+            Hover me
+            <div slot="prefix">
```
cscheffauer, line 778: is this issue only present if a icon gets slotted in? what if "prefixIcon" property is used?

cscheffauer (changes requested):

packages/components/src/components/button/button.e2e-test.ts
dominiccarrington: The problem only exists when an element is slotted in. `prefixIcon` and `postfixIcon` work correctly.

packages/components/src/components/button/button.stories.ts
dominiccarrington: Agreed we should in this case. However if we only have a brandvisual in the prefix slot, the issue appears also

cscheffauer (approved):

PR #2029, opened 2026-08-20
feat(components): add checkbox tree widget

iplesko:
### Description

Introduce `mdc-checkboxtree`, a new component for hierarchical multi-selection. It allows users to select individual options or entire branches, with parent states reflecting the selections below them.

The component includes accessible keyboard navigation and group-level messaging and validation states.

This release also improves `mdc-checkbox` interaction consistency for programmatic activation.

Screenshot 2026-08-21 at 13 28 25

### Breaking Change

None.


packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -139,16 +153,21 @@ class Checkbox
...
       this.indeterminate = false;
+      return true;
     }
+    return false;
   }
 
   override click() {
     super.click();
-    this.toggleState();
+    if (this.toggleState()) {
```
cscheffauer: suggestion: it would be better readable if the return value of this.toggleState() is saved as a const.
That way its clearer that toggleState() gets fired on click.

same for line 194

```js
const hasBeenToggled = this.toggleState();

If(hasBeenToggled) {
...
```

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -172,7 +191,9 @@ class Checkbox
     }
     if (this.getKeyboardNavMode() === NAV_MODES.SPATIAL) {
       if (!(this.readonly || this.softDisabled) && action === ACTIONS.ENTER) {
-        this.toggleState();
+        if (this.toggleState()) {
+          this.dispatchEvent(new Event('change', { bubbles: true, composed: true }));
```
cscheffauer, line 197: q: do we need change events? the dispatchModifiedEvent already covers if the checked changes.

packages/components/src/components/checkboxtree/knowledge-base/checkboxtree.component.md
```diff
@@ -0,0 +1,151 @@
...
+- Write checkbox labels as distinct noun phrases in sentence case. Do not add ending punctuation unless a label is a
+  full sentence.
+- Keep helper text brief and explain a requirement or consequence instead of repeating the group label.
+- State errors with the action needed to continue, such as “Select at least one team member.”
+
+### Property/Attribute details
+
+| Option                      | Intent                                                                                                                                                 |
+| --------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `--mdc-checkboxtree-indent` | Changes the logical indentation added at every nested level. Keep the default unless the surrounding layout requires a coordinated spacing adjustment. |
```
cscheffauer: css properties do not need to be documented here. Its enough to document this in the checkboxtree.component JSDoc, where all CSS properties are described.

packages/components/src/components/checkboxtree/knowledge-base/checkboxtree.component.md
```diff
@@ -0,0 +1,151 @@
...
+and `aria-checked` value. Parent checkboxes expose `mixed` when their descendants do not share one state.
+
+Disabled checkboxes are skipped during arrow navigation. Read-only and soft-disabled checkboxes remain focusable so
+their state can still be discovered, but they cannot be changed.
+
+#### Internal ARIA managed by the component
+
+| Element                       | Attribute          | Value                                                                             |
+| ----------------------------- | ------------------ | --------------------------------------------------------------------------------- |
+| Outermost host                | `role`             | `group`                                                                           |
```
cscheffauer: i just inspected the tree in chrome devtools, this role=group is not visible.
Also curious how this works - is the aria-label passed in on the "Parent checkbox", or on the group? maybe we need 2 aria-label props?

cscheffauer (changes requested):

packages/components/src/components/checkbox/checkbox.component.ts
iplesko: Sure, makes sense, done.

packages/components/src/components/checkboxtree/knowledge-base/checkboxtree.component.md
iplesko: Yes, sounds good, removed, thanks.

packages/components/src/components/checkbox/checkbox.component.ts
iplesko: Well, we don't necessarily need them, it works for users just fine without them, but programatic `.click()`s and spatial navigation "Enter" on a parent checkbox will not cascade the new state to its descendants in `mdc-checkboxtree`.

cscheffauer (approved):

PR #2033, opened 2026-08-21
feat(checkbox): support leading visuals

iplesko:
### Description

Allow clients to compose a checkbox with a decorative leading visual between the checkbox control and its existing label and help text.

An example of checkbox with the leading visual:

```html
<mdc-checkbox 
      label="Alex Example" 
      help-text="example.com" 
      toggletip-text="This participant is connected from a video device."
      info-icon-aria-label="About Alex Example">
  <mdc-avatar slot="leading-visual" size="32" aria-hidden="true"></mdc-avatar>
</mdc-checkbox>
```

Checkbox with the leading visual

### Breaking Change

None.


packages/components/src/components/checkbox/checkbox.stories.ts
```diff
@@ -149,6 +159,33 @@ export const Indeterminate: StoryObj = {
...
+        size="32"
+        src=${imageFixtures.avatar}
+        presence=${PRESENCE_TYPE.ON_DEVICE}
+      ></mdc-avatar>
+      <span slot="label">Alex Example</span>
+      <span slot="supporting-text">example.com</span>
+      <mdc-button
+        slot="toggletip"
+        id="lobby-participant-info-trigger"
+        prefix-icon="info-circle-regular"
```
cscheffauer: info-circle-filled is what is normally used and recommended - we should use this version here too.

packages/components/src/components/checkbox/checkbox.e2e-test.ts
```diff
@@ -312,6 +379,64 @@ test('mdc-checkbox', async ({ componentsPage }) => {
...
+    });
+
+    await test.step('toggletip trigger should not toggle the checkbox', async () => {
+      await componentsPage.mount({
+        html: `
+          <mdc-checkbox label="Alex Example" supporting-text="example.com">
+            <mdc-button
+              slot="toggletip"
+              id="participant-info-trigger"
+              prefix-icon="info-circle-regular"
```
cscheffauer: info-circle-filled is what is normally used and recommended - we should use this version here too.

packages/components/src/components/checkbox/checkbox.e2e-test.ts
```diff
@@ -210,6 +251,18 @@ test('mdc-checkbox', async ({ componentsPage }) => {
       await expect(label).toHaveText('Checkbox label');
     });
 
+    await test.step('should include supporting text in the accessible name', async () => {
+      await componentsPage.setAttributes(checkbox, {
+        label: 'Kristine Stone',
+        'supporting-text': 'cisco.com',
```
cscheffauer: lets use example.com everywhere

cscheffauer (changes requested):
first round of review

packages/components/src/components/checkbox/checkbox.stories.ts
iplesko: Instead, I completely remvoed the toggletip slot usage. Instead, the examples use the checkbox's `toggletip-text` parameter as all the other tests or stories. That seems like more standard usage.

packages/components/src/components/checkbox/checkbox.e2e-test.ts
iplesko: Same as above.

packages/components/src/components/checkbox/checkbox.e2e-test.ts
iplesko: Fixed and also improved `AGENTS.md` to improve chance this won't happen again.

packages/components/src/components/checkbox/checkbox.e2e-test.ts
```diff
@@ -186,11 +209,87 @@ test('mdc-checkbox', async ({ componentsPage }) => {
     });
   });
 
+  /**
+   * LAYOUT
+   */
+  await test.step('layout', async () => {
+    await test.step('should align a rich-label checkbox and preserve text-container box customizations', async () => {
```
iplesko: This was a real regression I introduced during development. It's quite hard to catch, imo, so I thought this test could be valuable. Let me know what you think, because on the other hand, the test is quite verbose, catches what is arguably an edge case. I would still keep it though.

packages/components/src/components/checkbox/checkbox.component.ts
```diff
@@ -75,6 +90,17 @@ class Checkbox
    */
   @property({ type: Boolean, reflect: true }) indeterminate = false;
 
+  /**
+   * Supporting text that identifies or qualifies the checkbox option. It is part of the checkbox label target.
+   */
+  @property({ type: String, reflect: true, attribute: 'supporting-text' }) supportingText?: string;
```
cscheffauer: issue: help-text already exists in formfieldwrapper and should be used. Creating another one here is causing inconsistencies

cscheffauer (changes requested):

packages/components/src/components/checkbox/checkbox.component.ts
iplesko: Removed and simplified even further.

cscheffauer (approved):

PR #2055, opened 2026-09-17
fix: updated radio component to fix a cursor/hover/active bug.

tricarro:
### Description
Observed an issue where the radio component did not have a hover or active state. It also did not have the cursor: pointer property applied to it (like the checkbox component).

Discovered that the issue what the ":part(static-radio)" reference in the CSS. There's no part with that label. The correct label is "radio-indicator". 

Updated styles in the radio.styles.ts file to fix it.

### Breaking Change

This is not a breaking change.


cscheffauer (approved):
great catch! 💯

PR #2059, opened 2026-09-22
feat(numberinput): implement numberinput component

dominiccarrington:
### Description

Implement mdc-numberinput for numerical entry. Supports min, max, step properties similar to standard HTML input type="number" implementation.

Screenshot 2026-09-22 at 11 14 03

### Breaking Change

None


packages/components/src/components/input/input.component.ts
```diff
@@ -126,6 +126,11 @@ class Input
    */
   @property({ type: String }) autocomplete: AutoCompleteType = AUTO_COMPLETE.OFF;
 
+  /**
+   * Hints at the type of data that might be entered by the user and the corresponding virtual keyboard.
+   */
+  @property({ type: String }) inputmode?: InputModeType;
```
cscheffauer, line 132: reflect true needs to be set to work appropriately

packages/components/src/components/input/input.component.ts
```diff
@@ -227,7 +232,7 @@ class Input
     }
   }
 
-  private setInputValidity() {
+  protected setInputValidity() {
```
cscheffauer, line 235: does this require a change to the password & search component?

packages/components/src/components/input/input.stories.ts
```diff
@@ -131,6 +132,10 @@ const meta: Meta = {
     autocomplete: {
       control: 'text',
     },
+    inputmode: {
```
cscheffauer, line 135: issue: the input has a story called "Slotted input element", which passes a number input. If we keep it like that, we should make it clear that consumers should use the number input vs slotting in a html input with type number (or replace it with a different example which doesn't exist)

packages/components/src/components/input/input.component.ts
```diff
@@ -126,6 +126,11 @@ class Input
    */
   @property({ type: String }) autocomplete: AutoCompleteType = AUTO_COMPLETE.OFF;
 
+  /**
+   * Hints at the type of data that might be entered by the user and the corresponding virtual keyboard.
```
cscheffauer, line 130: in the knowledge base of input, there should be information about "when to use the normal input" vs "when to use the number input"

packages/components/src/components/numberinput/knowledge-base/numberinput.component.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Use `mdc-numberinput` to collect a single numeric value that users may want to
+  fine-tune in small increments, such as a quantity, count, or price.
+- Use `mdc-numberinput` with `min` and `max` to constrain input to a valid
+  range, and `step` to set the amount each stepper click or arrow-key press
+  changes the value.
+- Use `mdc-numberinput` `step="any"` to accept decimal values without
+  step-mismatch validation. For values that must follow a fixed increment, such
+  as currency, set a numeric step like `step="0.01"` instead.
+- Use `mdc-numberinput` `hide-steppers` when the field should behave as a plain
```
cscheffauer, line 29: issue: "hide-steppers" is a bit misleading, instead spinner-buttons should be used as a word. Steppers are separate components in this library

packages/components/src/components/numberinput/numberinput.component.ts
```diff
@@ -0,0 +1,269 @@
...
+import type { InputModeType, InputType } from '../input/input.types';
+
+import { DEFAULTS } from './numberinput.constants';
+import styles from './numberinput.styles';
+
+/**
+ * Parses the `step` attribute, allowing the native `any` sentinel (no step-mismatch validation)
+ * alongside numeric values. `type: Number` can't express this, hence the custom converter.
+ */
+const stepConverter = {
```
cscheffauer, line 15: don't really see the reason for this being abstracted out of the component - instead just inline it (if its not shared with other components)

packages/components/src/components/numberinput/numberinput.component.ts
```diff
@@ -0,0 +1,269 @@
...
+        ?disabled=${this.disabled || this.readonly}
+        @click=${isIncrement ? this.handleIncrement : this.handleDecrement}
+        tabindex="-1"
+      ></mdc-button>
+    `;
+  }
+
+  /**
+   * Renders the decrement/increment steppers in place of the default trailing (clear) button,
+   * matching the Figma "Number Input" design, unless `hideSteppers` is set.
```
cscheffauer, line 248: it doesn't need to be documented that this matches a design spec

packages/components/src/components/numberinput/numberinput.component.ts
```diff
@@ -0,0 +1,269 @@
...
+      <mdc-button
+        part="stepper-button"
+        class="own-focus-ring"
+        variant=${DEFAULTS.STEPPER_BUTTON_VARIANT}
+        size="${DEFAULTS.STEPPER_BUTTON_SIZE}"
+        prefix-icon="${isIncrement ? DEFAULTS.INCREMENT_ICON : DEFAULTS.DECREMENT_ICON}"
+        aria-label="${isIncrement ? this.incrementAriaLabel : this.decrementAriaLabel}"
+        ?disabled=${this.disabled || this.readonly}
+        @click=${isIncrement ? this.handleIncrement : this.handleDecrement}
+        tabindex="-1"
```
cscheffauer, line 241: worth documenting why these stepper button don't require being focused

packages/components/src/components/numberinput/numberinput.stories.ts
```diff
@@ -0,0 +1,351 @@
...
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { action } from 'storybook/actions';
+
+import { classArgType, styleArgType } from '../../../config/storybook/commonArgTypes';
+import { disableControls, hideControls } from '../../../config/storybook/utils';
+import { VALIDATION } from '../formfieldwrapper/formfieldwrapper.constants';
+import { INPUT_MODE } from '../input/input.constants';
+
+const render = (args: Args) =>
+  html` <mdc-numberinput
```
cscheffauer, line 13: issue: if i do set min & max values and then modify the input with my keyboard, its not restricting it (i.e. if i do min 0 and max 10, i can type 11 and the input just stays as is).

packages/components/src/components/numberinput/numberinput.stories.ts
```diff
@@ -0,0 +1,351 @@
...
+    min: 0,
+    max: 10,
+    step: 1,
+    'hide-steppers': true,
+    'help-text': 'Enter a value between 0 and 10',
+    'help-text-type': 'default',
+  },
+};
+
+export const AnyStep: StoryObj = {
```
cscheffauer, line 205: issue: decimal is removed if the - / + buttons are pressed

packages/components/src/components/numberinput/numberinput.stories.ts
```diff
@@ -0,0 +1,351 @@
...
+        help-text-type="default"
+        placeholder="Placeholder"
+        value="5"
+        hide-steppers
+      ></mdc-numberinput>
+    </div>`,
+};
+
+export const FormFieldNumber: StoryObj = {
+  render: (args: any) => {
```
cscheffauer, line 315: we should build this story such that not only the form itself validates, but we also change the helptext underneath the input to error, to showcase thats also possible

packages/components/src/components/numberinput/numberinput.stories.ts
```diff
@@ -0,0 +1,351 @@
...
+    required: true,
+    'help-text': 'Helper text',
+    'help-text-type': 'default',
+    'validation-message': '',
+    'toggletip-text': 'Enter a whole number.',
+    'info-icon-aria-label': 'More information',
+  },
+};
+
+export const WithoutSteppers: StoryObj = {
```
cscheffauer, line 190: same comment as before, the word "steppers" should not be used, instead use spin-button. That follows the rest of the repo

cscheffauer (changes requested):
first round of review

packages/components/src/components/input/input.component.ts
dominiccarrington: `setInputValidity` is not overridden on password & search, therefore no change is required

packages/components/src/components/numberinput/numberinput.stories.ts
dominiccarrington: Firefox removes the decimal from the number when using the keyboard/spinner buttons, Chrome/Edge/Safari keeps them.

Will update to keep the decimal

packages/components/src/components/numberinput/numberinput.stories.ts
dominiccarrington: Updated to have an enum `clamp` prop that automatically updates the value to min or max when change event is fired.

We shouldn't use the input event here since if min=5, trying to type in 10 will reset the value to 5 as soon as the 1 is entered so the user will enter 50

PR #2060, opened 2026-09-22
feat(brandvisualprovider): add mdc-brandvisualprovider for http asset resolution

sloki009:
### Description

Adds **`mdc-brandvisualprovider`**, giving `mdc-brandvisual` the same HTTP asset resolution that
`mdc-iconprovider` already provides for icons.

Previously `mdc-brandvisual` resolved its asset exclusively through a dynamic import of the lit
template shipped in `@momentum-design/brand-visuals`. Because that specifier is not statically
analysable, a bundler has to account for the entire asset set regardless of how many visuals an
application actually renders. With a provider, hosts can serve the assets themselves and keep them
out of the bundle entirely.

```html
<mdc-brandvisualprovider
  brand-visual-set="custom-brand-visuals"
  url="/assets/brand-visuals/svg"
  file-extension="svg"
>
  <mdc-brandvisual name="cisco-logo-light-color"></mdc-brandvisual>
</mdc-brandvisualprovider>
```

| Attribute | Default | Description |
| --- | --- | --- |
| `brand-visual-set` | `momentum-brand-visuals` | `custom-brand-visuals` enables URL resolution |
| `url` | – | Base URL the visual name is appended to |
| `file-extension` | `svg` | `svg` or `png` |
| `cache-strategy` | – | `in-memory-cache` or `web-cache-api` |
| `cache-name` | – | Required when `cache-strategy` is set |

`svg` visuals are fetched and inlined, so they remain styleable through `::part(brandvisual)`.
`png` visuals are rendered as an `img` and loaded by the browser, so they are subject to the HTTP
cache rather than `cache-strategy`. Both extensions are supported because the asset set is split:
the `device-*` visuals ship only as raster images, every other visual only as a vector.

**This is fully backwards compatible.** `mdc-brandvisual` keeps using the dynamic import whenever
there is no provider, when the provider is on the `momentum-brand-visuals` set, or when a custom
set is missing a `url` — so existing usage is unaffected and the provider is entirely opt-in.

Also fixed: an unset `alt-text` previously rendered as a literal `alt="undefined"` on raster brand
visuals, which screen readers would announce.

### Breaking Change

None.


packages/components/src/utils/asset-fetch/index.ts
```diff
@@ -0,0 +1,132 @@
...
+ * @param signal - The signal to abort the fetch.
+ * It is used to cancel the fetch when the component is disconnected or updated.
+ * @param cacheStrategy - The cache strategy to use.
+ * @param cacheName - The cache name to use.
+ * @param assetType - Human readable name of the asset kind, used in error messages.
+ *
+ * @returns Response string from the fetch
+ * @throws Error if the response is not ok
+ */
+const svgFetch = async ({
```
cscheffauer: this could fetch anything, not only svg - rename appropriately

sloki009:
**Two commits, worth reviewing separately.**

`refactor(components): share svgFetch between icon and illustration` — `icon.utils.ts` and
`illustration.utils.ts` held byte-identical copies of `svgFetch`, including the abort handling, the
in-flight dedupe map and both cache strategies. Rather than add a third copy for brand visuals, it
moves to `src/utils/asset-fetch` and both components point at it; the asset kind is passed in as
`assetType` so error messages stay as specific as they were. This is the part that reaches beyond
the new component, which is why it's isolated and independently revertible.

`feat(brandvisualprovider): resolve brand visuals over http` — the provider, the `Brandvisual`
wiring, stories, knowledge base and e2e tests.

**Alt-text injection now branches.** `injectTemplateAttributes` parses `litTemplate.strings[0]`,
which only works on a `TemplateResult` from the dynamic import, while both provider branches
produce an `Element`. The previous code cast unconditionally to `Element`.

**`scripts/copyBrandvisuals.js`** now also copies the full set to the Playwright public dir, so the
URL mode can be tested against real assets rather than the single fixture the dynamic import is
rewritten to in e2e.

**Validation.** `eslint`, `stylelint`, `tsc --noEmit` and `syncpack lint` clean.
`build:components` picks the element up in `custom-elements.json` with all five attributes,
`build:react` emits `dist/react/brandvisualprovider`, `knowledge-base:validate` passes.

Playwright chrome, full suite: 275 passed / 5 failed. The failures (`list`, `menubar`,
`menupopover`, `select`, `FocusRing`) are focus/popover assertions unrelated to asset loading — two
passed on re-run and the other three fail identically on clean `main`.

I wasn't able to run `test:e2e:docker` locally. The new suite has no visual-regression step, so it
should add no baselines, and the only DOM change on the momentum path is removing the stringified
`alt="undefined"` — but CI confirming that would be good.

packages/components/src/components/brandvisualprovider/knowledge-base/brandvisualprovider.component.md
```diff
@@ -0,0 +1,121 @@
...
+
+The brandvisualprovider renders nothing of its own; it has no role, no focusable surface, and no keyboard interaction. Accessibility for brand visuals is owned entirely by the nested `mdc-brandvisual` components — see the `mdc-brandvisual` documentation for the ARIA contract.
+
+#### Internal ARIA managed by the component
+
+None. The provider contributes no element to the accessibility tree.
+
+### Implementation requirements
+
+#### General
```
cscheffauer: this implementation requirement section is meant for accessibility only!

packages/components/src/components/brandvisualprovider/brandvisualprovider.stories.ts
```diff
@@ -0,0 +1,95 @@
...
+  </mdc-brandvisualprovider>
+`;
+
+const meta: Meta = {
+  title: 'Providers/Brand Visual Provider',
+  tags: ['autodocs'],
+  component: 'mdc-brandvisualprovider',
+  render,
+  parameters: {
+    badges: ['stable'],
```
cscheffauer: badge is not required, please remove

packages/components/src/utils/asset-fetch/index.ts
```diff
@@ -0,0 +1,132 @@
...
+const fetchAsset = async (request: Request, assetType: string): Promise<Response> =>
+  fetch(request).then(response => {
+    if (!response.ok) {
+      throw new Error(`There was a problem while fetching the ${assetType}!`);
+    }
+    return response;
+  });
+
+// Map of in-flight fetch promises keyed by URL to deduplicate concurrent requests
+const pendingFetches = new Map<string, Promise<string>>();
```
cscheffauer, line 38: issue: since this is now one map for all icons, illustrations and brandvisuals, we need to make sure the key in the map is unique. For example if a icon and a illustration have the same name / same url there might be issues. To avoid that, i would recommend to combine the assetType with the assetUrl as the key for the cache map.

cscheffauer (changes requested):

packages/components/src/utils/asset-fetch/index.ts
sloki009: Renamed to assetFetch. The inner helper was fetchAsset, which would have read as a near-synonym of the exported one, so that's now requestAsset.

packages/components/src/components/brandvisualprovider/brandvisualprovider.stories.ts
sloki009: Removed

packages/components/src/components/brandvisualprovider/knowledge-base/brandvisualprovider.component.md
sloki009: Yeah, and the schema says so explicitly — body.config.json defines the section as the accessibility responsibilities a consumer must satisfy. The provider-placement bullet was already covered under Composition so I dropped it, and the caching guidance moved to a new ### Notes under Guidelines. Only #### Labeling remains.

packages/components/src/utils/asset-fetch/index.ts
sloki009: the map is now keyed by ${assetType}:${assetUrl}, so asset kinds can't share a pending request.

I also made assetType required rather than defaulting to 'asset', otherwise a future consumer that forgot to pass it would silently land in a shared bucket and reintroduce the same hazard.

